PR #5, opened 2015-11-30
Vendor specific library compile scripts for Linux

Paebbels:
So, I translated the PowerShell scripts into Bash scripts.
Feel free to test them, to compare the documentation and
find bugs :).

P.S. I don't know why GitHub lists so many commits for this PR.


Paebbels:
The problem with the `GT11` primitive is known.
I just didn't figure out into which library it should be compiled.
Maybe you have an idea.

Possibilities:
 - unisim
 - secureip

I tested both and it doesn't work.

tgingold:
Merged.
Thank you!!

PR #6, opened 2015-12-06
Add bind only option (-b) to make command (-m).

forrestv:
See https://mail.gna.org/public/ghdl-discuss/2015-08/msg00050.html for original discussion, quoted below.

>If building a project with GHDL that will be embedded within another
>program via ghdl_main, it seems that there is no equivalent of `ghdl -m`
>that automatically analyzes dependencies.
>
>Right now I'm using:
>
>    ghdl -m $(GHDLFLAGS) toplevel_tb || true
>    ghdl --bind $(GHDLFLAGS) toplevel_tb # not absolutely necessary,
>but catches errors that would have otherwise been ignored by above
>line
>
>That runs -m as far as it will get (which isn't all the way, since my
>VHDL code has some FFI calls that aren't getting linked in) and then
>runs --bind for the sole purpose of knowing when ghdl -m failed for
>any other reason.
>
>Is this functionality actually missing (automatic analysis of
>dependencies, but no linking, similar to --bind)? Am I just missing
>it?
>
>If it is actually missing, I could potentially implement it. Maybe as
>additional option provided to the -m command? `--no-link`?

I'm not sure that this is the best way to do this, but it removes the need for the workaround mentioned in that email.

### DIFF
--- a/src/ghdldrv/ghdldrv.adb
+++ b/src/ghdldrv/ghdldrv.adb
@@ -1248,6 +1248,9 @@ package body Ghdldrv is
 
    --  Command Make.
    type Command_Make is new Command_Comp with record
+      -- Bind only; don't link
+      Flag_Bind_Only : Boolean;
+
       --  Disp dependences during make.
       Flag_Depend_Unit : Boolean;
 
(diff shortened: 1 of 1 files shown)


tgingold:
Do not hesitate to also update the documentation (see doc/ )

forrestv:
Documented in #76.

PR #76, opened 2016-06-13
Document bind only (-b) option to make (-m) command from commit 876bf7fea6d.

forrestv:
See #6.


forrestv:
Note that I tried to use Sphinx (by running `make`) to regenerate `ghdl.texi`, but that resulted in a lot of other changes to `ghdl.texi`, so I ended up manually editing it. Someone should look at that. Perhaps `ghdl.texi` shouldn't be in the repository at all, I think?

tgingold:
Thanks.  Maybe the ghdl.texi generation depends on the sphinx version.

PR #100, opened 2016-06-27
Jonsba/master

### LINKED ISSUE #87: support for unbuffered output
I'm using ghdl with VHPI to setup co-simulations. In the 'external' part
stdout is either set to autoflush (e.g. in perl), or a flush is done after
each message (e.g. in C/C++).

In my test environment the output is usually send through a pipe.
Apparently ghdl uses buffered IO when vhdl textio outputs are written.

This leads to messages from ghdl and the 'external' part to be out-of-order.

Some other simulators have an option like 'UNBUFFERED' or 'AUTOFLUSH' which
will force a flush for each line for stdout writes.

If not it be great when it is added on a future release.

Jonsba:
This pull request add an --unbuffered option for simulation. It disables buffering for stdout, stderr and for files opened in write or append mode (TEXTIO).

It should fix #87, but I need confirmation from Walter (wfjm).

### DIFF
--- a/doc/Simulation_and_runtime.rst
+++ b/doc/Simulation_and_runtime.rst
@@ -113,6 +113,11 @@ all options available, including the debugging one.
   design.
 
 
+.. option:: --unbuffered
+
+  Disable buffering on stdout, stderr and files opened in write or append mode (TEXTIO).
+
+
 .. option:: --vcd=<FILENAME>
 
 .. option:: --vcdgz=<FILENAME>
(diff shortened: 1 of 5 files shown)


Paebbels:
That's strange...
* llvm backend has no errors, but
* mcode backend reports:  
  `grt-options.ads:143:80: (style) this line is too long` => [see line 336](https://travis-ci.org/tgingold/ghdl/jobs/140643815#L336)

tgingold:
Thanks!

PR #113, opened 2016-07-03
Add a description for the unbuffered option in ghdl -r <exec> --help.

Jonsba:


tgingold:
Thanks.
I might rephrase the description to have only one line.

PR #117, opened 2016-07-07
Clean up links, repository-wide

BenWiederhake:
Looking to make drive-by contributions, I noticed that there are a lot of `http://...` links that easily could be replaced by `https://...` links, which are usually deemed "safer".

As you might have objections to the reliability of certain hosts to stick with https, this path is split up by hostname.  I manually verified each unique URL to work properly.

Finally, there are a few links which are now invalid: see the last commit for that, which marks the dead links as such:
- The lz4 repository seems to now live at [https://github.com/Cyan4973/lz4]. so I'll contact upstream about getting that fixed eventually.
- I have no idea where the `vhdl.org` might have gone

This is the command I have grown while searching for non-`https` links:
```
grep -nrF http: . | grep -vF '//ghdl.free.fr' | grep -vF '//libre.adacore.com' | grep -vF '//osvvm.org' | grep -vF '//kassiopeia.juls.savba.sk' | grep -vF '//burtleburtle.net' | grep -vF '//www.opensource.org' | grep -vF '//www.xilinx.com' | grep -vF '//www.eecg.toronto.edu'
```
Fun fact: this is simultaneously the list of all webmasters that haven't switched to https, or did so incorrectly.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,7 +7,7 @@ New in 0.34 (XXXX-XX-XX)
 * Mode backend ported to x86_64
 * Support cocotb (https://github.com/potentialventures/cocotb)
 * Main repository is now on github: https://github.com/tgingold/ghdl
-* Docs available on rtd: http://ghdl.readthedocs.org/en/latest/
+* Docs available on rtd: https://ghdl.readthedocs.org/en/latest/
 * Speed improved.
 * New option --psl-report, to report status of psl assert and cover.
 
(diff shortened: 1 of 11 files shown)


Paebbels:
Nice grep and sed work, but you must not change license texts!

The website http://www.vhdl.org and other related websites are offline.

BenWiederhake:
I rebased it accordingly.  I was aware that this is changing license text, but wrongly assumed that the links are just outdated (license texts sometimes change, for example the FSF address in GPL).

Regarding the lz4 thing: [upstream changed it](https://github.com/Cyan4973/lz4/pull/219/commits/cec38d1b74e7b642b65e615517150f9587aedf7a), so I fixed the broken link in here, too.

tgingold:
Thanks!

PR #121, opened 2016-07-23
Adding support for a wave option file that selects signals to be displayed

Jonsba:
Adding support for a wave option file that selects signals to be displayed on the waveform (currently only works with the ghw wave format). Only full signal paths are supported now (no wildcards). Wave option file version set to 1.0.

### DIFF
--- a/src/grt/grt-strings.adb
+++ b/src/grt/grt-strings.adb
@@ -65,7 +65,7 @@ package body Grt.Strings is
       return -1;
    end Find;
    function Find (Str : String; Char : Character; Start : Positive)
-            return Integer is
+                 return Integer is
    begin
       return Find (Str (Start .. Str'Last), Char);
    end Find;
@@ -86,7 +86,8 @@ package body Grt.Strings is
       end loop;
    end To_Lower;
 
-   function Value (Str : String) return Integer is
+   function Value (Str : String) return Integer
+   is
       Decimal : Positive := 1;
       Value_Tmp : Natural := 0;
       Digit : Integer;
@@ -105,10 +106,10 @@ package body Grt.Strings is
    function Value (Char : Character) return Integer is
    begin
       case Char is
-      when '0' .. '9' =>
-         return Character'Pos (Char) - Character'Pos ('0');
-      when others =>
-         return -1;
+         when '0' .. '9' =>
+            return Character'Pos (Char) - Character'Pos ('0');
+         when others =>
+            return -1;
       end case;
    end Value;
 end Grt.Strings;
--- a/src/grt/grt-strings.adb
+++ b/src/grt/grt-strings.adb
@@ -32,9 +32,9 @@ package body Grt.Strings is
 
    function First_Non_Whitespace_Pos (Str : String) return Integer is
    begin
-      for Index in Str'Range loop
-         if not Is_Whitespace (Str (Index)) then
-            return Index;
+      for I in Str'Range loop
+         if not Is_Whitespace (Str (I)) then
+            return I;
          end if;
       end loop;
       return -1;
--- a/src/grt/grt-strings.adb
+++ b/src/grt/grt-strings.adb
@@ -88,10 +88,13 @@ package body Grt.Strings is
 
    function Value (Str : String) return Integer
    is
-      Decimal : Positive := 1;
-      Value_Tmp : Natural := 0;
+      Decimal : Positive;
+      Value_Tmp : Natural;
       Digit : Integer;
    begin
+      Decimal := 1;
+      Value_Tmp := 0;
+
       for Index in reverse Str'Range loop
          Digit := Value (Str (Index));
          if Digit = -1 then
--- a/src/grt/grt-strings.adb
+++ b/src/grt/grt-strings.adb
@@ -30,6 +30,46 @@ package body Grt.Strings is
       return C = ' ' or C = NBSP or C = HT;
    end Is_Whitespace;
 
+   function First_Non_Whitespace_Pos (Str : String) return Integer is
+   begin
+      for Index in Str'Range loop
+         if not Is_Whitespace (Str (Index)) then
+            return Index;
+         end if;
+      end loop;
+      return -1;
+   end First_Non_Whitespace_Pos;
+
+   function Last_Non_Whitespace_Pos (Str : String) return Integer is
+   begin
+      for Index in reverse Str'Range loop
+         if not Is_Whitespace (Str (Index)) then
+            return Index;
+         end if;
+      end loop;
+      return -1;
+   end Last_Non_Whitespace_Pos;
+
+   function New_Line_Pos (Line : String) return Integer is
+   begin
+      return Find (Line, ASCII.LF);
+   end New_Line_Pos;
+
+   function Find (Str : String; Char : Character) return Integer is
+   begin
+      for Index in Str'Range loop
+         if Str (Index) = Char then
+            return Index;
+         end if;
+      end loop;
+      return -1;
+   end Find;
+   function Find (Str : String; Char : Character; Start : Positive)
+            return Integer is
+   begin
+      return Find (Str (Start .. Str'Last), Char);
+   end Find;
+
    function To_Lower (C : Character) return Character is
    begin
       if C in 'A' .. 'Z' then
@@ -38,4 +78,37 @@ package body Grt.Strings is
          return C;
       end if;
    end To_Lower;
+
+   procedure To_Lower (S : in out String) is
+   begin
+      for I in S'Range loop
+         S (I) := To_Lower (S (I));
+      end loop;
+   end To_Lower;
+
+   function Value (Str : String) return Integer is
+      Decimal : Positive := 1;
+      Value_Tmp : Natural := 0;
+      Digit : Integer;
+   begin
+      for Index in reverse Str'Range loop
+         Digit := Value (Str (Index));
+         if Digit = -1 then
+            return -1;
+         end if;
+         Value_Tmp := Value_Tmp + Digit * Decimal;
+         Decimal := Decimal * 10;
+      end loop;
+      return Value_Tmp;
+   end Value;
+
+   function Value (Char : Character) return Integer is
+   begin
+      case Char is
+      when '0' .. '9' =>
+         return Character'Pos (Char) - Character'Pos ('0');
+      when others =>
+         return -1;
+      end case;
+   end Value;
 end Grt.Strings;
--- a/src/grt/grt-strings.ads
+++ b/src/grt/grt-strings.ads
@@ -31,18 +31,23 @@ package Grt.Strings is
    --  Return True IFF C is a whitespace character (as defined in LRM93 14.3)
    function Is_Whitespace (C : in Character) return Boolean;
 
-   -- The following functions return -1 in case there is no match in string
+   -- The following functions return -1 in case there is no match in string ---
+
    -- Return the index of the first non whitespace character in string
    function First_Non_Whitespace_Pos (Str : String) return Integer;
+
    -- Return the index of the last non whitespace character in string
    function Last_Non_Whitespace_Pos (Str : String) return Integer;
+
    -- Return the index of the new line character (ASCII.LF) in string
    function New_Line_Pos (Line : String) return Integer;
+
    -- Return the index of the first character that matches Char in string
    function Find (Str : String; Char : Character) return Integer;
    function Find (Str : String; Char : Character; Start : Positive)
-            return Integer;
-   --
+                 return Integer;
+
+   ----------------------------------------------------------------------------
 
    --  Convert C/S to lowercase.
    function To_Lower (C : Character) return Character;
--- a/src/grt/grt-strings.ads
+++ b/src/grt/grt-strings.ads
@@ -30,12 +30,19 @@ package Grt.Strings is
 
    --  Return True IFF C is a whitespace character (as defined in LRM93 14.3)
    function Is_Whitespace (C : in Character) return Boolean;
+
+   -- The following functions return -1 in case there is no match in string
+   -- Return the index of the first non whitespace character in string
    function First_Non_Whitespace_Pos (Str : String) return Integer;
+   -- Return the index of the last non whitespace character in string
    function Last_Non_Whitespace_Pos (Str : String) return Integer;
+   -- Return the index of the new line character (ASCII.LF) in string
    function New_Line_Pos (Line : String) return Integer;
+   -- Return the index of the first character that matches Char in string
    function Find (Str : String; Char : Character) return Integer;
    function Find (Str : String; Char : Character; Start : Positive)
             return Integer;
+   --
 
    --  Convert C/S to lowercase.
    function To_Lower (C : Character) return Character;
--- a/src/grt/grt-wave_opt_file-parse-debug.adb
+++ b/src/grt/grt-wave_opt_file-parse-debug.adb
@@ -1,5 +1,5 @@
---  GHDL Run Time (GRT) - mono-thread version.
---  Copyright (C) 2005 - 2014 Tristan Gingold
+--  GHDL Run Time (GRT) - Wave option file package for debugging.
+--  Copyright (C) 2016 Jonas Baggett
 --
 --  GHDL is free software; you can redistribute it and/or modify it under
 --  the terms of the GNU General Public License as published by the Free
@@ -23,9 +23,11 @@
 --  however invalidate any other reasons why the executable file might be
 --  covered by the GNU Public License.
 
+-- Description: See package specifications
+
 with Grt.Astdio; use Grt.Astdio;
 
-package body Grt.Wave_Options.Parse.Debug is
+package body Grt.Wave_Opt_File.Parse.Debug is
 
    procedure Dump_Tree is
    begin
@@ -48,8 +50,9 @@ package body Grt.Wave_Options.Parse.Debug is
 
    procedure Dump_Sub_Tree (Previous_Cursor : Elem_Acc; Level : Positive)
    is
-      Current_Cursor : Elem_Acc := Previous_Cursor;
+      Current_Cursor : Elem_Acc;
    begin
+      Current_Cursor := Previous_Cursor;
       while Current_Cursor /= null loop
          Put ((3 .. 2 * Level => ' '));
          Put ('/'); Put (Current_Cursor.Name.all); New_Line;
@@ -58,4 +61,4 @@ package body Grt.Wave_Options.Parse.Debug is
       end loop;
    end Dump_Sub_Tree;
 
-end Grt.Wave_Options.Parse.Debug;
+end Grt.Wave_Opt_File.Parse.Debug;
--- a/src/grt/grt-wave_opt_file-parse-debug.adb
+++ b/src/grt/grt-wave_opt_file-parse-debug.adb
@@ -31,33 +31,31 @@ package body Grt.Wave_Opt_File.Parse.Debug is
 
    procedure Dump_Tree is
    begin
+      New_Line;
       for Index in Tree_Index_Type'Range loop
-         New_Line;
-         New_Line;
-         Put ("----------------------------");
-         New_Line;
+         Put_Line ("----------------------------");
          if Index = Pkg then
-            Put ("Packages : ");
+            Put_Line ("Packages : ");
          else
-            Put ("Instances : ");
+            Put_Line ("Instances : ");
          end if;
-         New_Line;
          Dump_Sub_Tree (Trees (Index), 1);
       end loop;
-      Put ("----------- END -----------------");
+      Put_Line ("----------- END -----------------");
       New_Line;
    end Dump_Tree;
 
-   procedure Dump_Sub_Tree (Previous_Cursor : Elem_Acc; Level : Positive)
+   procedure Dump_Sub_Tree (Cursor : Elem_Acc; Level : Positive)
    is
-      Current_Cursor : Elem_Acc;
+      Sibling_Cursor : Elem_Acc;
    begin
-      Current_Cursor := Previous_Cursor;
-      while Current_Cursor /= null loop
+      Sibling_Cursor := Cursor;
+      while Sibling_Cursor /= null loop
          Put ((3 .. 2 * Level => ' '));
-         Put ('/'); Put (Current_Cursor.Name.all); New_Line;
-         Dump_Sub_Tree (Current_Cursor.Next_Child, Level + 1);
-         Current_Cursor := Current_Cursor.Next_Sibling;
+         Put ('/');
+         Put_Line (Sibling_Cursor.Name.all);
+         Dump_Sub_Tree (Sibling_Cursor.Next_Child, Level + 1);
+         Sibling_Cursor := Sibling_Cursor.Next_Sibling;
       end loop;
    end Dump_Sub_Tree;
 
--- a/src/grt/grt-wave_options-parse-debug.adb
+++ b/src/grt/grt-wave_options-parse-debug.adb
@@ -30,7 +30,10 @@ package body Grt.Wave_Options.Parse.Debug is
    procedure Dump_Tree is
    begin
       for Index in Tree_Index_Type'Range loop
-         New_Line; New_Line; Put ("----------------------------"); New_Line;
+         New_Line;
+         New_Line;
+         Put ("----------------------------");
+         New_Line;
          if Index = Pkg then
             Put ("Packages : ");
          else
@@ -39,16 +42,15 @@ package body Grt.Wave_Options.Parse.Debug is
          New_Line;
          Dump_Sub_Tree (Trees (Index), 1);
       end loop;
-      Put ("----------- END -----------------"); New_Line;
+      Put ("----------- END -----------------");
+      New_Line;
    end Dump_Tree;
 
    procedure Dump_Sub_Tree (Previous_Cursor : Elem_Acc; Level : Positive) is
       Current_Cursor : Elem_Acc := Previous_Cursor;
    begin
       while Current_Cursor /= null loop
-         for i in 2 .. Level loop
-            Put ("   ");
-         end loop;
+         Put ((3 .. 2 * Level => ' '));
          Put ('/'); Put (Current_Cursor.Name.all); New_Line;
          Dump_Sub_Tree (Current_Cursor.Next_Child, Level + 1);
          Current_Cursor := Current_Cursor.Next_Sibling;
--- a/src/grt/grt-wave_options-parse-debug.adb
+++ b/src/grt/grt-wave_options-parse-debug.adb
@@ -46,7 +46,8 @@ package body Grt.Wave_Options.Parse.Debug is
       New_Line;
    end Dump_Tree;
 
-   procedure Dump_Sub_Tree (Previous_Cursor : Elem_Acc; Level : Positive) is
+   procedure Dump_Sub_Tree (Previous_Cursor : Elem_Acc; Level : Positive)
+   is
       Current_Cursor : Elem_Acc := Previous_Cursor;
    begin
       while Current_Cursor /= null loop
--- a/src/grt/grt-wave_options-parse-debug.adb
+++ b/src/grt/grt-wave_options-parse-debug.adb
@@ -1,5 +1,5 @@
---  GHDL Run Time (GRT) - Misc subprograms for characters and strings
---  Copyright (C) 2016 Tristan Gingold
+--  GHDL Run Time (GRT) - mono-thread version.
+--  Copyright (C) 2005 - 2014 Tristan Gingold
 --
 --  GHDL is free software; you can redistribute it and/or modify it under
 --  the terms of the GNU General Public License as published by the Free
@@ -23,25 +23,36 @@
 --  however invalidate any other reasons why the executable file might be
 --  covered by the GNU Public License.
 
-package Grt.Strings is
-   pragma Pure;
+with Grt.Astdio; use Grt.Astdio;
 
-   NBSP : constant Character := Character'Val (160);
+package body Grt.Wave_Options.Parse.Debug is
 
-   --  Return True IFF C is a whitespace character (as defined in LRM93 14.3)
-   function Is_Whitespace (C : in Character) return Boolean;
-   function First_Non_Whitespace_Pos (Str : String) return Integer;
-   function Last_Non_Whitespace_Pos (Str : String) return Integer;
-   function New_Line_Pos (Line : String) return Integer;
-   function Find (Str : String; Char : Character) return Integer;
-   function Find (Str : String; Char : Character; Start : Positive)
-            return Integer;
+   procedure Dump_Tree is
+   begin
+      for Index in Tree_Index_Type'Range loop
+         New_Line; New_Line; Put ("----------------------------"); New_Line;
+         if Index = Pkg then
+            Put ("Packages : ");
+         else
+            Put ("Instances : ");
+         end if;
+         New_Line;
+         Dump_Sub_Tree (Trees (Index), 1);
+      end loop;
+      Put ("----------- END -----------------"); New_Line;
+   end Dump_Tree;
 
-   --  Convert C/S to lowercase.
-   function To_Lower (C : Character) return Character;
-   procedure To_Lower (S : in out String);
+   procedure Dump_Sub_Tree (Previous_Cursor : Elem_Acc; Level : Positive) is
+      Current_Cursor : Elem_Acc := Previous_Cursor;
+   begin
+      while Current_Cursor /= null loop
+         for i in 2 .. Level loop
+            Put ("   ");
+         end loop;
+         Put ('/'); Put (Current_Cursor.Name.all); New_Line;
+         Dump_Sub_Tree (Current_Cursor.Next_Child, Level + 1);
+         Current_Cursor := Current_Cursor.Next_Sibling;
+      end loop;
+   end Dump_Sub_Tree;
 
-   -- Str/Char : image of a natural number/digit
-   function Value (Str : String) return Integer;
-   function Value (Char : Character) return Integer;
-end Grt.Strings;
+end Grt.Wave_Options.Parse.Debug;
--- a/src/grt/grt-wave_options-parse.adb
+++ b/src/grt/grt-wave_options-parse.adb
@@ -192,7 +192,8 @@ package body Grt.Wave_Options.Parse is
       loop
          First := Last;
 
-         loop -- Find next identifier
+         -- Find next identifier
+         loop
             if Line (Last) = Sep (Tree_Index) then
                Last := Last - 1;
                exit;
@@ -213,8 +214,10 @@ package body Grt.Wave_Options.Parse is
             return;
          end if;
 
-         Last := Last + 2; -- Skip the separator
-         if Last > Line'Last then -- catch signal paths ending with /
+         -- Skip the separator
+         Last := Last + 2;
+         -- Catch signal paths ending with /
+         if Last > Line'Last then
             Error_Context ("invalid signal path");
          end if;
 
@@ -236,8 +239,10 @@ package body Grt.Wave_Options.Parse is
                                    Line_Context => Line_Context,
                                    Kind => Not_Found,
                                    Next_Sibling | Next_Child => null);
-            if Previous_Sibling_Cursor = null then -- First element of level ?
-               if Previous_Tree_Cursor = null then -- Is a top_level ?
+            -- First element of level ?
+            if Previous_Sibling_Cursor = null then
+               -- Is a top level ?
+               if Previous_Tree_Cursor = null then
                   Trees (Tree_Index) := Elem;
                else
                   Previous_Tree_Cursor.Next_Child := Elem;
@@ -246,7 +251,8 @@ package body Grt.Wave_Options.Parse is
                Previous_Sibling_Cursor.Next_Sibling := Elem;
             end if;
             Previous_Tree_Cursor := Elem;
-            Tree_Cursor := null; -- Point to Elem.Next_Child which is null
+            -- Point to Elem.Next_Child which is null
+            Tree_Cursor := null;
             return True;
          -- Identifier was found in the tree ? Then move to its first child
          elsif Elem_Name = Sibling_Cursor.Name.all then
--- a/src/grt/grt-wave_options-parse.adb
+++ b/src/grt/grt-wave_options-parse.adb
@@ -29,12 +29,12 @@ with Grt.Strings; use Grt.Strings;
 with Grt.Vstrings; use Grt.Vstrings;
 with Grt.Errors; use Grt.Errors;
 
-with Grt.Wave_Options.Parse.Debug;
+--~ with Grt.Wave_Options.Parse.Debug;
 
 package body Grt.Wave_Options.Parse is
 
    procedure Start (Option_File : String) is
-      Stream : FILEs := File_Open (Option_File);
+      Stream : constant FILEs := File_Open (Option_File);
       First, Last : Integer;
       Line : String (1 .. Buf_Size);
       Lineno : Natural;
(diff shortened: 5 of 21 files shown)


src/grt/grt-strings.adb
```diff
@@ -30,6 +30,46 @@ package body Grt.Strings is
       return C = ' ' or C = NBSP or C = HT;
    end Is_Whitespace;
 
+   function First_Non_Whitespace_Pos (Str : String) return Integer is
+   begin
+      for Index in Str'Range loop
```
tgingold: I usually prefer I/J/K for loop indexes, but that's a detail

src/grt/grt-strings.ads
```diff
@@ -30,7 +30,18 @@ package Grt.Strings is
 
    --  Return True IFF C is a whitespace character (as defined in LRM93 14.3)
    function Is_Whitespace (C : in Character) return Boolean;
+   function First_Non_Whitespace_Pos (Str : String) return Integer;
+   function Last_Non_Whitespace_Pos (Str : String) return Integer;
+   function New_Line_Pos (Line : String) return Integer;
+   function Find (Str : String; Char : Character) return Integer;
+   function Find (Str : String; Char : Character; Start : Positive)
+            return Integer;
```
tgingold: Please, add a comment before each specification to explain the purpose of each subprogram.

src/grt/grt-wave_options-parse-debug.adb
```diff
@@ -0,0 +1,58 @@
+--  GHDL Run Time (GRT) - mono-thread version.
+--  Copyright (C) 2005 - 2014 Tristan Gingold
```
tgingold: Don't forget to change the first two lines!

src/grt/grt-wave_options-parse-debug.adb
```diff
@@ -0,0 +1,58 @@
...
+--  covered by the GNU Public License.
+
+with Grt.Astdio; use Grt.Astdio;
+
+package body Grt.Wave_Options.Parse.Debug is
+
+   procedure Dump_Tree is
+   begin
+      for Index in Tree_Index_Type'Range loop
+         New_Line; New_Line; Put ("----------------------------"); New_Line;
```
tgingold: Please, one statement per line.

src/grt/grt-wave_options-parse-debug.adb
```diff
@@ -0,0 +1,58 @@
...
+      end loop;
+      Put ("----------- END -----------------"); New_Line;
+   end Dump_Tree;
+
+   procedure Dump_Sub_Tree (Previous_Cursor : Elem_Acc; Level : Positive) is
+      Current_Cursor : Elem_Acc := Previous_Cursor;
+   begin
+      while Current_Cursor /= null loop
+         for i in 2 .. Level loop
+            Put ("   ");
```
tgingold: You can write: Put ((2 .. 2 * Level => ' '));

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,317 @@
...
+   procedure Start (Option_File : String) is
+      Stream : FILEs := File_Open (Option_File);
+      First, Last : Integer;
+      Line : String (1 .. Buf_Size);
+      Lineno : Natural;
+   begin
+      File_Path := new String'(Option_File);
+      Lineno := 0;
+
+      loop -- Reads line after line
```
tgingold: Comments like this one should be on a separate line (and terminated by a dot).

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,317 @@
...
+         exit when fgets (Line'Address, Line'Length, Stream) = Null_Address;
+         Lineno := Lineno + 1;
+
+         Last := New_Line_Pos (Line) - 1;
```
tgingold: Grouping statements like this is nice.  And you can event add a short comment before:
 --  Determine end of line.

--  Skip empty lines.

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,317 @@
...
+         if First = -1 or else Line (First) = '#' then
+            goto Continue;
+         end if;
+
+         Last := Last_Non_Whitespace_Pos (Line (First .. Last));
+         declare
+            Line_Str : String (1 .. Last + 1 - First) := Line (First .. Last);
+         begin
+            Line_Context :=
+               new Line_Context_Type'(Str => new String'(Line_Str),
```
tgingold: You don't need to declare Line_Str here.  Simply write:
new String'(Line (First .. Last))

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,317 @@
...
+         end if;
+
+         <<Continue>> null;
+      end loop;
+
+      if Version.Major = -1 then
+         Report_C ("warning: version wasn't set at the beginning of the file;" &
+                   " currently supported version is ");
+         Print_Version (Current_Version);
+         Newline_Err;
```
tgingold: I have to improve the doc of grt.errors.  You should not use newline_err at the end, but Report_E.

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,317 @@
...
+         Error_Context ("version cannot be set after signal paths");
+      end if;
+
+      First := First_Non_Whitespace_Pos (Line (Line'First + 1 .. Line'Last));
+      if Line (First .. First + 6) /= "version" then
+         Error_Context (Msg_Invalid_Format);
+      end if;
+
+      First := First + 7;
+      if not Is_Whitespace (Line (First)) then --catch "version\n", "version1.0"
```
tgingold: Comment on a separate line.

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,318 @@
...
+   end Print_Version;
+
+   --------------------------------------------------------------------------
+
+   procedure Parse_Path (Line : String_Acc) is
+      First, Last : Positive;
+      Tree_Updated : Boolean;
+      Tree_Index : Tree_Index_Type;
+   begin
+      To_Lower (Line_Context.Str.all);
```
tgingold: I don't think you should use to_lower now, only when comparing with VHDL path components.  Suppose one day verilog is added...

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,318 @@
...
+            goto Continue;
+         end if;
+
+         Last := Last_Non_Whitespace_Pos (Line (First .. Last));
+         declare
+            Line_Str : constant String (1 .. Last + 1 - First) :=
+                       Line (First .. Last);
+         begin
+            Line_Context :=
+               new Line_Context_Type'(Str => new String'(Line_Str),
```
tgingold: You can write: new string'(Line_Str (first .. last))
or use an intermediate subtype if you need to change the bounds.
I don't think you need to use an access type: pass directly the string.
(that would avoid the need to deallocate it).

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,318 @@
...
+
+   --------------------------------------------------------------------------
+
+   procedure Parse_Path (Line : String_Acc) is
+      First, Last : Positive;
+      Tree_Updated : Boolean;
+      Tree_Index : Tree_Index_Type;
+   begin
+      To_Lower (Line_Context.Str.all);
+      Last := Line'First;
```
tgingold: I will resume review later!

src/grt/grt-wave_options-parse-debug.adb
Jonsba: Do you mean that I should write "Copyright (C) 2016 Jonas Baggett" ?
What do I need to change in the first line ?

src/grt/grt-wave_options-parse-debug.adb
Jonsba: I didn't know that Ada allows to create strings like this. That's nice to know.
If Level = 1 then no space should be printed, then it should be Put ((3 .. 2 * Level => ' ')) instead.

src/grt/grt-wave_options-parse.adb
Jonsba: I am thinking that if I add a short comment before these group statements, I should also do the same to all group statements in my code. Do you think it worths it ? Or maybe I should do it only at the more complicated parts of my code.

src/grt/grt-wave_options-parse.adb
Jonsba: I first write it that way, but there is a subtility :

If the line begin with one (or more) space, First will be 2 (or higher) and then the index of the first character of Line_Context.Str will be 2 (or higher) instead of 1.

Then I had an issue in the Check_If_Found procedure in grt-wave_options-read.adb with the function call to Find, but now I just found out that if I change the line :
   Index := 1;
to :
   Index := Cursor.Line_Context.Str'First;
then the issue is fixed.

src/grt/grt-wave_options-parse.adb
Jonsba: Fixed

src/grt/grt-wave_options-parse.adb
Jonsba: Fixed

src/grt/grt-strings.adb
Jonsba: Fixed

src/grt/grt-strings.ads
Jonsba: Fixed

src/grt/grt-wave_options-parse-debug.adb
Jonsba: Fixed

src/grt/grt-wave_options-parse.adb
Jonsba: Fixed

src/grt/grt-wave_options-parse.adb
Jonsba: The call to to_lower is actually needed, because I found out that the components of the design that I get with Vhpi_Scan have their name already lowered.

src/grt/grt-wave_options-parse.adb
Jonsba: Ok, thanks a lot for what you have already done !

src/grt/grt-wave_options-parse.adb
Jonsba: > I don't think you need to use an access type: pass directly the string.

The problem is that the Str field in the Line_Context_Type record cannot be constrained since the line size vary, that's why I made it an access type.

Should I add some deallocations after signals are put in the wave file (ghw, vcd, ...) ?

Jonsba: And Line_Context also needs to be an access because if we have : /top/sub/a
then the tree will looks like : top -> sub -> a.
Now if Line_Context isn't an access, then in the tree "top", "sub" and 
"a" will have their own copy of Line_Context.
I will use some pseudo code to explain the problem with that. Here is 
what we get after the line is parsed :

top.Line_Context.Max_Level = 1
sub.Line_Context.Max_Level = 2
a.Line_Context.Max_Level = 3

Where Max_Level is the number of tree levels of the line (3 in our 
case). So we had expected instead :

top.Line_Context.Max_Level = 3
sub.Line_Context.Max_Level = 3
a.Line_Context.Max_Level = 3

Which can be only reached when "top", "sub" and "a" share a common 
Line_Context.

Actually Line_Context could be deallocated right after Parse_Version 
(Line_Context.Str), or maybe I should rather rewrite my code so that 
Parse_Version doesn't use Line_Context anymore but passes the needed 
informations in a different way in the calls to Error_Context and 
Print_Context.

src/grt/grt-wave_options-parse-debug.adb
tgingold: On 23/07/16 08:50, Jonsba wrote:
> In src/grt/grt-wave_options-parse-debug.adb
> <https://github.com/tgingold/ghdl/pull/121#discussion_r71968540>:
>
>> @@ -0,0 +1,58 @@
>> +--  GHDL Run Time (GRT) - mono-thread version.
>> +--  Copyright (C) 2005 - 2014 Tristan Gingold
>
> Do you mean that I should write "Copyright (C) 2016 Jonas Baggett" ?

Yes, as you have written it.

> What do I need to change in the first line ?

Add a brief description after the '-', like Debug for Wave_Options parser.

src/grt/grt-wave_options-parse.adb
tgingold: On 23/07/16 09:09, Jonsba wrote:
> In src/grt/grt-wave_options-parse.adb
> <https://github.com/tgingold/ghdl/pull/121#discussion_r71968736>:
>
>> +package body Grt.Wave_Options.Parse is
>> +
>> +   procedure Start (Option_File : String) is
>> +      Stream : FILEs := File_Open (Option_File);
>> +      First, Last : Integer;
>> +      Line : String (1 .. Buf_Size);
>> +      Lineno : Natural;
>> +   begin
>> +      File_Path := new String'(Option_File);
>> +      Lineno := 0;
>> +
>> +      loop -- Reads line after line
>> +         exit when fgets (Line'Address, Line'Length, Stream) = Null_Address;
>> +         Lineno := Lineno + 1;
>> +
>> +         Last := New_Line_Pos (Line) - 1;
>
> I am thinking that if I add a short comment before these group
> statements, I should also do the same to all group statements in my
> code. Do you think it worths it ? Or maybe I should do it only at the
> more complicated parts of my code.

That's only comments so it is a matter of taste.
You don't need to have comments for all group of statements.
 From time to time, I add comments when re-reading code so that it could 
be read faster.

src/grt/grt-wave_options-parse.adb
tgingold: On 23/07/16 09:23, Jonsba wrote:
> In src/grt/grt-wave_options-parse.adb
> <https://github.com/tgingold/ghdl/pull/121#discussion_r71968896>:
>
>> +         Last := New_Line_Pos (Line) - 1;
>> +         if Last < 0 then
>> +            Last := Line'Last;
>> +         end if;
>> +
>> +         First := First_Non_Whitespace_Pos (Line (Line'First .. Last));
>> +         if First = -1 or else Line (First) = '#' then
>> +            goto Continue;
>> +         end if;
>> +
>> +         Last := Last_Non_Whitespace_Pos (Line (First .. Last));
>> +         declare
>> +            Line_Str : String (1 .. Last + 1 - First) := Line (First .. Last);
>> +         begin
>> +            Line_Context :=
>> +               new Line_Context_Type'(Str => new String'(Line_Str),
>
> I first write it that way, but there is a subtility :
>
> If the line begin with one (or more) space, First will be 2 (or higher)
> and then the index of the first character of Line_Context.Str will be 2
> (or higher) instead of 1.

Right.  But then you could also write:
   subtype line_str_type is string (1 .. last + 1 - first);
...
   new Line_Str_Type'(Line (First .. Last));

But:
> Then I had an issue in the Check_If_Found procedure in
> grt-wave_options-read.adb with the function call to Find, but now I just
> found out that if I change the line :
> Index := 1;

You should avoid to hard-code bounds!

> to :
> Index := Cursor.Line_Context.Str'First;
> then the issue is fixed.

Yes!

src/grt/grt-wave_options-parse.adb
tgingold: On 23/07/16 09:45, Jonsba wrote:
> In src/grt/grt-wave_options-parse.adb
> <https://github.com/tgingold/ghdl/pull/121#discussion_r71969045>:
>
>> +   begin
>> +      To_String (Num_Str, First, Ghdl_I32 (Version.Major));
>> +      Report_C (Num_Str (First .. Num_Str'Last));
>> +      Report_C (".");
>> +      To_String (Num_Str, First, Ghdl_I32 (Version.Minor));
>> +      Report_C (Num_Str (First .. Num_Str'Last));
>> +   end Print_Version;
>> +
>> +   --------------------------------------------------------------------------
>> +
>> +   procedure Parse_Path (Line : String_Acc) is
>> +      First, Last : Positive;
>> +      Tree_Updated : Boolean;
>> +      Tree_Index : Tree_Index_Type;
>> +   begin
>> +      To_Lower (Line_Context.Str.all);
>
> The call to to_lower is actually needed, because I found out that the
> components of the design that I get with Vhpi_Scan have their name
> already lowered.

I agree that sooner or later it is needed.  But you can also reuse 
grt.vpi.strcasecmp (once exported).

My point is: suppose we add support to verilog (or system-C). 
Identifiers are case-sensitive in these languages.  So you cannot 
convert to lower-case during parsing as you don't know the context.

I agree this could be fixed later.

src/grt/grt-strings.ads
```diff
@@ -31,6 +31,24 @@ package Grt.Strings is
    --  Return True IFF C is a whitespace character (as defined in LRM93 14.3)
    function Is_Whitespace (C : in Character) return Boolean;
 
-   --  Convert C to lowercase.
+   -- The following functions return -1 in case there is no match in string
+   -- Return the index of the first non whitespace character in string
+   function First_Non_Whitespace_Pos (Str : String) return Integer;
+   -- Return the index of the last non whitespace character in string
```
tgingold: Nit-picking: blank line before the comment for a function.

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,323 @@
...
+         if Last > Line'Last then -- catch signal paths ending with /
+            Error_Context ("invalid signal path");
+         end if;
+
+      end loop;
+
+   end Parse_Path;
+
+   function Update_Tree (Elem_Name : String; Tree_Index : Tree_Index_Type)
+            return Boolean is
```
tgingold: 'is' on a separate line.
The style rules for subprograms are not simple, but they are documented in the STYLE file.

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,323 @@
...
+      Sibling_Cursor : Elem_Acc := Tree_Cursor;
```
tgingold: As documented in the STYLE file, variables must not be initialised.  Transform the initialisation in a statement.
Maybe this rule is stupid.

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,323 @@
...
+   begin
+      loop
+         -- Already reached the last sibling and current identifier corresponds
+         -- to no existing element ? Then we will create an element
+         if Sibling_Cursor = null then
+            Elem := new Elem_Type'(Name => new String'(Elem_Name),
+                                   Line_Context => Line_Context,
+                                   Kind => Not_Found,
+                                   Next_Sibling | Next_Child => null);
+            if Previous_Sibling_Cursor = null then -- First element of level ?
```
tgingold: Position of the comment.  There are other cases below.

src/grt/grt-wave_options-parse.adb
```diff
@@ -0,0 +1,323 @@
...
+         Sibling_Cursor := Sibling_Cursor.Next_Sibling;
+      end loop;
+   end Update_Tree;
+
+   procedure Check_Validity (Elem_Name : String) is
+   begin
+      if Elem_Name'Length = 0 then
+         Error_Context ("invalid signal path");
+      end if;
+      for Index in Elem_Name'Range loop
```
tgingold: Do you need to check the validity ?
At worst, they won't match with VHDL names and thus a warning will be reported.

src/grt/grt-wave_options-parse.ads
```diff
@@ -0,0 +1,64 @@
+--  GHDL Run Time (GRT) - mono-thread version.
+--  Copyright (C) 2005 - 2014 Tristan Gingold
```
tgingold: To be changed (I think in all files)

src/grt/grt-wave_options-parse.ads
```diff
@@ -0,0 +1,64 @@
...
+
+with Grt.Stdio; use Grt.Stdio;
+
+package Grt.Wave_Options.Parse is
+   pragma Preelaborate;
+
+   procedure Start (Option_File : String);
+
+private
+
```
tgingold: Comments before each declaration.

src/grt/grt-wave_options-parse.ads
```diff
@@ -0,0 +1,64 @@
...
+      Major : Integer;
+      Minor : Integer;
+   end record;
+   Version : Version_Type := (others => -1);
+   Current_Version : constant Version_Type := (Major => 1, Minor => 0);
+
+   type Sep_Array is array (Tree_Index_Type) of Character;
+   Sep : constant Sep_Array := (Pkg => '.', Entity => '/');
+
+   procedure Parse_Version (Line : String_Acc);
```
tgingold: Unless called from a child package, you don't need to declare subprograms here.  They could be declared only in the body.
(The private part of a package is quiet different from C++ class private use).

src/grt/grt-wave_options-read.adb
```diff
@@ -0,0 +1,125 @@
...
+            Check_If_Found (Cursor.Next_Child, Sep, Level + 1);
+         end if;
+         Cursor := Cursor.Next_Sibling;
+      end loop;
+
+   end Check_If_Found;
+
+   function Display_All_Signals return Boolean is
+   begin
+      if Trees = Tree_Array'(others => null) then
```
tgingold: You can write:
  return Trees = (others => null);

src/grt/grt-wave_options-read.ads
```diff
@@ -0,0 +1,49 @@
...
+
+   function Get_Top_Cursor (Name : Ghdl_C_String; Index : Tree_Index_Type)
+            return Elem_Acc;
+   function Get_Cursor (Name : Ghdl_C_String; Parent : Elem_Acc;
+            Is_Signal : Boolean := False) return Elem_Acc;
+   function Is_Displayed (Cursor : Elem_Acc) return Boolean;
+
+   procedure Check_If_All_Found;
+
+private
```
tgingold: Comments before functions.
Here I am pretty sure you don't need the private part.

src/grt/grt-wave_options.adb
```diff
@@ -0,0 +1,65 @@
...
+with Grt.Errors; use Grt.Errors;
+
+package body Grt.Wave_Options is
+
+   procedure Print_Context (Line_Context : Line_Context_Acc;
+                            Severity : Severity_Type) is
+      Lineno_Str : String (1 .. Value_String_Size);
+      First : Natural;
+   begin
+      case Severity is
```
tgingold: Indentation of 'when' is not correct.
(Which editor do you use ?)

src/grt/grt-wave_options.ads
```diff
@@ -0,0 +1,66 @@
...
+--  unit, or you link this unit with other files to produce an executable,
+--  this unit does not by itself cause the resulting executable to be
+--  covered by the GNU General Public License. This exception does not
+--  however invalidate any other reasons why the executable file might be
+--  covered by the GNU Public License.
+
+package Grt.Wave_Options is
+   pragma Preelaborate;
+
+   type String_Acc is access String;
```
tgingold: String_Acc is already defined in grt.types.  Simply reuse that type.

src/grt/grt-wave_options-parse.adb
Jonsba: Sorry I forgot to do the changes everywhere when you first pointed the 
problem. Now it is Fixed.

src/grt/grt-wave_options-parse.adb
Jonsba: There is one problem if I don't do the conversion to lower here :
Suppose we have the following :

/top/a
/TOP/b

then without a conversion to lower, the tree will looks like :
Tree ptr --> top --> a
          |-> TOP --> b

instead of :
Tree ptr --> top --> a
                  |-> b

If in the future we add support to verilog and system-C. I see 2 solutions :

1) - Create 2 trees when parsing : one case sensitive and one case 
insensitive
    - Then latter when we have more informations, prune VHDL paths from 
the case
      sensitive tree and prune verilog / system-C paths from the case
      insensitive tree (maybe it's not really needed)
    - Use the right tree while looking for signals to be displayed in 
the design

2) - Create only 1 case sensitive tree
    - Then latter, look for VHDL paths in the tree and merge elements 
who have
      the same name after lowering their characters.
    - No change should be afterwards needed in the code that find 
signals to be
      displayed

I have the impression that the first approach is slighty easier than the
second, and currently the code match well the first approach. Anyways, I 
would
prefer not add complexity before it's time and maybe in the futur I may 
need to
do code rewrites before we had support for verilog and system-C and then 
need
to have a different approach.

Nevertheless, we still have in mind that the case sensitivity is an 
issue that
needs to be fixed some time in the future :).

Jonsba: Strangely it has removed some spaces of the tree structure put in example.
The tree structure should rather look like that :
> Tree ptr --> top --> a
>              |-> TOP --> b
>
> instead of :
> Tree ptr --> top --> a
>                          |-> b
If hopefully it gets it right this time.

tgingold: On 23/07/16 14:04, Jonsba wrote:
> In src/grt/grt-wave_options-parse.adb
> <https://github.com/tgingold/ghdl/pull/121#discussion_r71971597>:
>
>> +   begin
>> +      To_String (Num_Str, First, Ghdl_I32 (Version.Major));
>> +      Report_C (Num_Str (First .. Num_Str'Last));
>> +      Report_C (".");
>> +      To_String (Num_Str, First, Ghdl_I32 (Version.Minor));
>> +      Report_C (Num_Str (First .. Num_Str'Last));
>> +   end Print_Version;
>> +
>> +   --------------------------------------------------------------------------
>> +
>> +   procedure Parse_Path (Line : String_Acc) is
>> +      First, Last : Positive;
>> +      Tree_Updated : Boolean;
>> +      Tree_Index : Tree_Index_Type;
>> +   begin
>> +      To_Lower (Line_Context.Str.all);
>
> There is one problem if I don't do the conversion to lower here :
> Suppose we have the following : /top/a /TOP/b then without a conversion
> to lower, the tree will looks like : Tree ptr --> top --> a |-> TOP -->
> b instead of : Tree ptr --> top --> a |-> b If in the future we add
> support to verilog and system-C. I see 2 solutions : 1) - Create 2 trees
> when parsing : one case sensitive and one case insensitive - Then latter
> when we have more informations, prune VHDL paths from the case sensitive
> tree and prune verilog / system-C paths from the case insensitive tree
> (maybe it's not really needed) - Use the right tree while looking for
> signals to be displayed in the design 2) - Create only 1 case sensitive
> tree - Then latter, look for VHDL paths in the tree and merge elements
> who have the same name after lowering their characters. - No change
> should be afterwards needed in the code that find signals to be
> displayed I have the impression that the first approach is slighty
> easier than the second, and currently the code match well the first
> approach. Anyways, I would prefer not add complexity before it's time
> and maybe in the futur I may need to do code rewrites before we had
> support for verilog and system-C and then need to have a different
> approach. Nevertheless, we still have in mind that the case sensitivity
> is an issue that needs to be fixed some time in the future :).

Makes sense.

Another possibility is to parse the option file after elaboration of the 
design.

So don't add complexity now, just add a TODO comment.

tgingold: On 23/07/16 14:11, Jonsba wrote:
> In src/grt/grt-wave_options-parse.adb
> <https://github.com/tgingold/ghdl/pull/121#discussion_r71971642>:
>
>> +   begin
>> +      To_String (Num_Str, First, Ghdl_I32 (Version.Major));
>> +      Report_C (Num_Str (First .. Num_Str'Last));
>> +      Report_C (".");
>> +      To_String (Num_Str, First, Ghdl_I32 (Version.Minor));
>> +      Report_C (Num_Str (First .. Num_Str'Last));
>> +   end Print_Version;
>> +
>> +   --------------------------------------------------------------------------
>> +
>> +   procedure Parse_Path (Line : String_Acc) is
>> +      First, Last : Positive;
>> +      Tree_Updated : Boolean;
>> +      Tree_Index : Tree_Index_Type;
>> +   begin
>> +      To_Lower (Line_Context.Str.all);
>
> Strangely it has removed some spaces of the tree structure put in
> example. The tree structure should rather look like that :
> Tree ptr --> top --> a |-> TOP --> b instead of : Tree ptr --> top --> a
> |-> b
> If hopefully it gets it right this time.

Looks like github likes to wrap lines.  Anyway I understand what you mean.

src/grt/grt-wave_options-parse.adb
tgingold: On 23/07/16 10:33, Jonsba wrote:
> In src/grt/grt-wave_options-parse.adb
> <https://github.com/tgingold/ghdl/pull/121#discussion_r71969510>:
>
>> +         if Last < 0 then
>> +            Last := Line'Last;
>> +         end if;
>> +
>> +         First := First_Non_Whitespace_Pos (Line (Line'First .. Last));
>> +         if First = -1 or else Line (First) = '#' then
>> +            goto Continue;
>> +         end if;
>> +
>> +         Last := Last_Non_Whitespace_Pos (Line (First .. Last));
>> +         declare
>> +            Line_Str : constant String (1 .. Last + 1 - First) :=
>> +                       Line (First .. Last);
>> +         begin
>> +            Line_Context :=
>> +               new Line_Context_Type'(Str => new String'(Line_Str),
>
> And Line_Context also needs to be an access because if we have :
> /top/sub/a then the tree will looks like : top -> sub -> a. Now if
> Line_Context isn't an access, then in the tree "top", "sub" and "a" will
> have their own copy of Line_Context. I will use some pseudo code to
> explain the problem with that. Here is what we get after the line is
> parsed : top.Line_Context.Max_Level = 1 sub.Line_Context.Max_Level = 2
> a.Line_Context.Max_Level = 3 Where Max_Level is the number of tree
> levels of the line (3 in our case). So we had expected instead :
> top.Line_Context.Max_Level = 3 sub.Line_Context.Max_Level = 3
> a.Line_Context.Max_Level = 3 Which can be only reached when "top", "sub"
> and "a" share a common Line_Context. Actually Line_Context could be
> deallocated right after Parse_Version (Line_Context.Str), or maybe I
> should rather rewrite my code so that Parse_Version doesn't use
> Line_Context anymore but passes the needed informations in a different
> way in the calls to Error_Context and Print_Context.

I am not sure why you really need the Str in Line_Context.
You are saving all significant lines of the file, while I think the line 
number would be enough for error message.

OTOH, the option file shouldn't be very large.

So keep Str.

tgingold: On 23/07/16 10:01, Jonsba wrote:
> In src/grt/grt-wave_options-parse.adb
> <https://github.com/tgingold/ghdl/pull/121#discussion_r71969276>:
>
>> +         if Last < 0 then
>> +            Last := Line'Last;
>> +         end if;
>> +
>> +         First := First_Non_Whitespace_Pos (Line (Line'First .. Last));
>> +         if First = -1 or else Line (First) = '#' then
>> +            goto Continue;
>> +         end if;
>> +
>> +         Last := Last_Non_Whitespace_Pos (Line (First .. Last));
>> +         declare
>> +            Line_Str : constant String (1 .. Last + 1 - First) :=
>> +                       Line (First .. Last);
>> +         begin
>> +            Line_Context :=
>> +               new Line_Context_Type'(Str => new String'(Line_Str),
>
>     I don't think you need to use an access type: pass directly the string.
>
> The problem is that the Str field in the Line_Context_Type record cannot
> be constrained since the line size vary, that's why I made it an access
> type.
>
> Should I add some deallocations after signals are put in the wave file
> (ghw, vcd, ...) ?

Add a TODO marker.  The amount of memory shouldn't be that significative.

src/grt/grt-wave_options-parse.adb
Jonsba: I have never look on the STYLE file before. Now I am correcting my code 
accordingly.

src/grt/grt-strings.ads
Jonsba: Fixed

src/grt/grt-wave_options.adb
Jonsba: I am using geany, but the indentation I use for 'when' is actually a 
personnal code style.

Fixed

src/grt/grt-wave_options-parse.adb
Jonsba: Fixed

src/grt/grt-wave_options-parse.adb
Jonsba: Personally, I am ok with initialisation in the declarative part of a 
variable, especially with simple functions and procedures, but in some 
cases I believe that doing the initialisation after the declarative part 
improves readability.

Fixed anyways

src/grt/grt-wave_options-parse.ads
Jonsba: Fixed

By the way, I made some renaming to improve readability :

Grt.Wave_Options -> Grt.Wave_Opt_File
Grt.Wave_Opt_File.Read -> Grt.Wave_Opt_File.Tree_Reading

src/grt/grt-wave_options-parse.ads
Jonsba: Do you mean before each declaration of function/procedure only ?

src/grt/grt-wave_options-parse.adb
Jonsba: You are right, it's useless, so I fixed it.

I have nevertheless added some more checks for path validity that won't 
be reported otherwise.

src/grt/grt-wave_options-read.adb
Jonsba: Fixed but gnat didn't allow me to write it that way (is it a compiler 
bug ?). I had to write instead :

return Trees = Tree_Array'(others => null);

src/grt/grt-wave_options.ads
Jonsba: Fixed

src/grt/grt-wave_options-parse.ads
Jonsba: Actually since I made some modifications in other packages (like Grt.Strings and Grt.Waves), should I also add my name to the copyright notice of the corresponding files ?

src/grt/grt-wave_opt_file-parse-debug.adb
```diff
@@ -0,0 +1,64 @@
...
+
+package body Grt.Wave_Opt_File.Parse.Debug is
+
+   procedure Dump_Tree is
+   begin
+      for Index in Tree_Index_Type'Range loop
+         New_Line;
+         New_Line;
+         Put ("----------------------------");
+         New_Line;
```
tgingold: For Put + New_LIne, you can use Put_Line.
(Sorry I missed that in my first review).

src/grt/grt-wave_options-parse.ads
tgingold: On 24/07/16 10:21, Jonsba wrote:
> In src/grt/grt-wave_options-parse.ads
> <https://github.com/tgingold/ghdl/pull/121#discussion_r71984179>:
>
>> @@ -0,0 +1,64 @@
>> +--  GHDL Run Time (GRT) - mono-thread version.
>> +--  Copyright (C) 2005 - 2014 Tristan Gingold
>
> Actually since I made some modifications in other packages (like
> Grt.Strings and Grt.Waves), should I also add my name to the copyright
> notice of the corresponding files ?

No, usually only the initial author is marked.
(That's in the GNU coding style that I try to follow).

src/grt/grt-wave_options-read.adb
tgingold: On 24/07/16 09:50, Jonsba wrote:
> In src/grt/grt-wave_options-read.adb
> <https://github.com/tgingold/ghdl/pull/121#discussion_r71983765>:
>
>> +            Report_E (Cursor.Name.all);
>> +         elsif Level = Cursor.Line_Context.Max_Level
>> +           and then Cursor.Kind = Pkg_Entity
>> +         then
>> +            Error_Context ("not a signal", Cursor.Line_Context, Warning);
>> +         else
>> +            Check_If_Found (Cursor.Next_Child, Sep, Level + 1);
>> +         end if;
>> +         Cursor := Cursor.Next_Sibling;
>> +      end loop;
>> +
>> +   end Check_If_Found;
>> +
>> +   function Display_All_Signals return Boolean is
>> +   begin
>> +      if Trees = Tree_Array'(others => null) then
>
> Fixed but gnat didn't allow me to write it that way (is it a compiler
> bug ?). I had to write instead : return Trees = Tree_Array'(others =>
> null);

No, indeed you need to write Tree_Array (as it is a constrained subtype).

src/grt/grt-wave_options-parse.ads
tgingold: Yes, each public subprogram must have a comment that describe it briefly.
I think this is important when you read the code 2 years later.

tgingold:
I think it is almost ready to be merged.
I will certainly do minor changes after that because it takes less time to do some reindentation than to write a comment for asking it!

src/grt/grt-wave_options-parse.adb
Jonsba: I will have a look on how ghdl reports compilation error in order to print errors/warning in a more similar format.

Actually I need to save the line because when the user makes a typo in a path like this :
/top/sub1/sub2/sub2/sub4/a
instead of :
/top/sub1/sub2/sub2/sub4/a​​

Then he will get a warning telling him that /top/sub1/sub2/sub2 wasn't found in design. Then he will immediately now that he made a typo with subb3.
I believe that this is really helpfull because if the path is long it isn't always obvious to find the error.
Maybe, I could also just warn that sub2 wasn't found, although it could exist several times in the path (as in this example​), then it won't point exactly to the problematic one.

But I think there is a better way to do that : I can add a parent field to the elements of the tree so that it is possible to find the full path of an element because we could now go from this element to the top of the tree.

src/grt/grt-wave_options-parse.adb
Jonsba: I added the TODO marker.

I will also have a look on how ghdl reports compilation error in order to print errors/warning in a similar format.

Actually I need to save the line because when the user makes a typo in a path like this :

/top/sub1/sub2/sub2/sub4/a

instead of :

/top/sub1/sub2/sub3/sub4/a

Then he will get a warning telling him that /top/sub1/sub2/sub2 wasn't found in design. Then he will immediately now that he made a typo with sub2.
I believe that this is really helpfull because if the path is long it isn't always obvious to find the error at first glance.
Maybe, I could also just warn that sub2 wasn't found, although it could exist several times in the path (like in this example), then it won't point exactly to the problematic one.

But I think there is a better way to do that : I can add a parent field to the elements of the tree so that it is possible to find the full path of an element because we could now go from this element to the top of the tree.

src/grt/grt-wave_options-parse.adb
tgingold: Adding a parent field looks to be a good idea.  Maybe for later ?

Jonsba: I was about to add it now. I just added (but not pushed yet) a Level 
field because I have subprogramms like Dump_Sub_Tree and 
Check_Sub_Tree_If_All_Found who need to know the level of the current 
element. Until recently there is a Level variable created in these 
functions but now it looks like to me that I should create a Level field 
instead.

src/grt/grt-wave_options-parse.adb
Jonsba: After looking at how ghdl reports compilations error, I will go for a warning like this :
warning: trace.txt:<line num>:<column num>: no VHDL object in design matches sub2
But I still need to be able to find the path of the parent of an element in order to find the value for <column num>.

By the way, what I mean by object could be either a signal an entity instanciation or a package name. Do you think it is the right term ?

src/grt/grt-wave_options-parse.adb
Jonsba: Would you prefer that I add the parent and the level field latter ? That's fine for me too

Jonsba: Hum, normally I don't even need the parent field. A column position field should be enough.

Jonsba: And the planned changes for the warning/error formatting could happen only after I add the column position field.

I think I should now stop doing changes, except the changes needed to meet the coding style requirements. How do you think ?

Jonsba: Then I will make another pull request for the changes with the warning/error messages formatting, which will include also the modifications of the tree fields.

Jonsba: I have finished to change the error/warning messages format and the changes on the tree fields. So it's up to you if you prefer that I push the changes now or do a new pull request for that once the current one is accepted.

tgingold:
Thanks!

Jonsba:
Thanks for your patience :).

PR #128, opened 2016-08-02
Jonsba/signals selection pr

Jonsba:
The wave option file support is extended to VCD and FST waves
    
There are still current issues which are described in the header TODO section of grt-fst.adb and grt-vcd.adb.


tgingold:
Thanks!

PR #135, opened 2016-08-04
When the wave option file provided as argument doesn't exist, creates…

Jonsba:
… it with all the signals of the design.


tgingold:
I think it would be better to have a new option to generate the option file.
(I have no example in mind of a unix tool that provide this 'write if doesn't
exist' behaviour so this is not very intuitive to me).

I am fine with the refactoring part.  I would have merged it if it were
a separate pull request.  It is better to create different requests for
different work topic, but I agree that sometimes one starts to refactor
during the development of a feature.

Tristan.

Jonsba:
What about touch ?

Jonsba:
Ok I can make 2 PR instead of 1.

Jonsba:
But yes, if you prefer I can still add an option for that ( something 
like : --wave-opt-file -c <file to be created>).

tgingold:
About touch: the behaviour is consistent. After a touch, the file has its time and date set to now.

I don't request to split this pull in two, I just say that it is better (in the future) to separate work that can be approved easily and work that could be discussed.  And I know we sometimes start a work, then refactor in the middle so it is not always easy to have two patches.

I really prefer to have two separate options, like --wave-opt-file and --write-wave-opt.

Jonsba:
Hello Tristan,

--wave-opt-file and --write-wave-opt looks inconsistent in my eyes.
I would prefer to have --read-wave-opt and --write-wave-opt.

What do you think ?
Jonas

tgingold:
On 17/10/16 07:49, Jonas Baggett wrote:
> Hello Tristan,
>
> --wave-opt-file and --write-wave-opt looks inconsistent in my eyes.
> I would prefer to have --read-wave-opt and --write-wave-opt.
>
> What do you think ?

That's fine with me.  We can change the option names and slightly break 
compatibility.

Tristan.

Jonsba:
Ok, great !

Jonsba:
I have one question : What should happen if the user wants to create the wave opt file with --write-wave-opt, but the file already exists ?
a) Should it be silently erased 
b) or an error should be thrown ?

I am thinking that it is possible that the user first creates the wave opt file, customize it and when he launches a new simulation, he might forget to change --write-wave-opt to --read-wave-opt, which would erase the customizations he made if we choose option a). And it makes me no sense to write a wave opt file without having in mind to customize it afterwards, because otherwise you could just launch the simulation without a wave opt file, you will get the same results.

So I am in favor of option a). What do you think ?

Paebbels:
I think I would expect an error.

tgingold:
On 17/10/16 08:25, Jonas Baggett wrote:
> I have one question : What should happen if the user wants to create the
> wave opt file with --write-wave-opt, but the file already exists ?
> a) Should it be silently erased
> b) or an error should be thrown ?

In general tools do a)

Tristan.

Jonsba:
Le 17. 10. 16 à 20:45, tgingold a écrit :

> On 17/10/16 08:25, Jonas Baggett wrote:
> > I have one question : What should happen if the user wants to create the
> > wave opt file with --write-wave-opt, but the file already exists ?
> > a) Should it be silently erased
> > b) or an error should be thrown ?
>
> In general tools do a)
>
> Tristan.

Ok, but is a) really a desirable behaviour in our case ? I believe that 
the typical workflow would be :

1) Use of --write-wave-opt to generate the list of the signals
2) Customization of the wave opt file
3) Use of --read-wave-opt to read the wave opt file and update the 
waveform accordingly.
4) After some changes in the VHDL code, use of --write-wave-opt to 
generate the updated list of the signals in a new file so that 
previously made customizations aren't erased.
5) Manual update of the original wave opt file with the updates from the 
newly generated wave opt file.
6) Repeat 4) and 5) as much as needed.

So if we choose option a), then the wave opt file could be erased by 
mistake during step 3) or 4). It seems in my eyes to be very likely that 
accidental erasures will happend during step 4) as this step will be 
done a numerous of times.

Jonas


> > So I am in favor of option a). What do you think ?
I made a typo, I meant that I am in favor of option b).

tgingold:
On 18/10/16 09:44, Jonas Baggett wrote:
> Le 17. 10. 16 à 20:45, tgingold a écrit :
>
>> On 17/10/16 08:25, Jonas Baggett wrote:
>> > I have one question : What should happen if the user wants to create the
>> > wave opt file with --write-wave-opt, but the file already exists ?
>> > a) Should it be silently erased
>> > b) or an error should be thrown ?
>>
>> In general tools do a)
>>
>> Tristan.
>
> Ok, but is a) really a desirable behaviour in our case ? I believe that
> the typical workflow would be :
>
> 1) Use of --write-wave-opt to generate the list of the signals
> 2) Customization of the wave opt file
> 3) Use of --read-wave-opt to read the wave opt file and update the
> waveform accordingly.
> 4) After some changes in the VHDL code, use of --write-wave-opt to
> generate the updated list of the signals in a new file so that
> previously made customizations aren't erased.
> 5) Manual update of the original wave opt file with the updates from the
> newly generated wave opt file.
> 6) Repeat 4) and 5) as much as needed.
>
> So if we choose option a), then the wave opt file could be erased by
> mistake during step 3) or 4). It seems in my eyes to be very likely that
> accidental erasures will happend during step 4) as this step will be
> done a numerous of times.

I don't have a strong opinion, it's up to you!

Tristan.

Jonsba:
Ok, then I will choose option b).

PR #139, opened 2016-08-08
Reworked compile scripts for Windows.

Paebbels:
Hello,

here are the latest reworks for the Windows compile flow with PowerShell.
**This commit changes:**
- Removed `-InstallPath <Path>` -> `-Install [<Path>]` is enough
- Uses internal help pages like the user compile scripts
- Steps can be run as a group e.g. `-Compile` or as single tasks, e.g. `-Clean_GHDL -Compile_GHDL`
- Improved error reporting
- `filter.exe` is no longer needed and build on Windows  
  It got replaced by PowerShell commands and act like the original `sed` commands in `Makefile`.
- The zip-file is created with a version number and back-end name. (`-CreatePackage -Zip`)

**Known issues:**
- Some lines are printed in verbose mode, even if `-Verbose` is disabled
- Installing GHDL in `PATH` is broken -> disabled  
  I need to rewrite the functions for setting/removing paths
- Local `README.md` needs updates.

### DIFF
--- a/dist/mcode/winbuild.ps1
+++ b/dist/mcode/winbuild.ps1
@@ -35,33 +35,35 @@
 #	Software Foundation, 59 Temple Place - Suite 330, Boston, MA
 #	02111-1307, USA.
 # ==============================================================================
-<#
-	.SYNOPSIS 
-		GHDL for Windows - GHDL compile script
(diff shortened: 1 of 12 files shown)


tgingold:
Thanks!

PR #171, opened 2016-10-19
winbuild: Fixed add and remove from PATH.

Paebbels:
Hello,

this PR fixes the long time intended behavior to add/remove GHDL from `$env:PATH`.

While installing, the script asks if GHDL should be registered in PATH (at machine level). The default is yes.
`Install GHDL in PATH at machine level? [Y/n]:`
GHDL gets removed from all levels (machine, user, session) while uninstalling.

Is this behavior ok?


Paebbels:
I hope GitHub recognized my forced push to add a hotfix into the same PR....
Travis is running again, so it seems to notice such events.

PR #179, opened 2016-10-31
Jonsba/signals selection

Jonsba:
There is a new --write-opt-file option that will create a wave option file with all the signals of the design.

The --wave-opt-file option is renamed to --read-opt-file for consistency

Some code cleanup and bug fixes


markNZed:
It seems the parser is not dealing with the escaping in VHDL 93, in the VHDL standard Std. 1076-1993 (page 174) you can use extended identifiers using a backslash at the beginning AND at the end.

\So this is one identifier\

With a signal name like \abc/xyz\ the '/' seems to be treated as a level of hierarchy.

Jonsba:
Hello,

I have never heard before about this feature. If I understand well a 
signal path could be something like the following :

/top/my_entity/\my signal\
/top/\my entity\/my_signal
/top/\my entity\/\my signal\

Is that right ?

Paebbels:
Names in VHDL can be either a normal identifier or an extended identifier. The latter one is a backslash enclosed string. So both `\` signs act like a double quote for a string. This allows user to e.g. use reserved words as identifiers.

Reference: LRM §15.4.3 Extended identifiers

PR #186, opened 2016-11-04
Win32.c quickfix for mingw64 llvm.

Paebbels:
This commit applies a quickfix for win32.c for MinGW64 with LLVM backend.


tgingold:
I am testing a slightly different version of this patch (with less code duplication).

Paebbels:
I was not sure how many asm lines will be added for x86_64, so I chose to duplicate more code:).

tgingold:
I don't think that the x86-64 version can be written in C.

Issue #213, opened 2016-12-02
Has GCCs -fPIC behavior changed?


Paebbels:
Hello,

I did a fresh GHDL build and executed a simple testbench. I get the following output. Please see the second half (elaboration), where `/usr/bin/ld` complains about `-fPIC`.

```Bash
paebbels@debian8:~/git/PoC$ ./poc.sh -v ghdl arith.prng
================================================================================
                         The PoC-Library - Service Tool                         
================================================================================
Initializing PoC-Library Service Tool for simulations
  Reading configuration files...
Preparing simulation environment...
  Creating fresh temporary directory.
  Changing working directory to temporary directory.
  Preparing GHDL simulator.
Testbench: PoC.arith.prng
  Creating PoC project 'arith_prng_tb'
  Reading filelist '/home/paebbels/git/PoC/tb/arith/arith_prng_tb.files'
Running analysis for every vhdl file...
  command: /opt/ghdl/0.34-dev-llvm/bin/ghdl -a -fexplicit -frelaxed-rules --warn-binding --no-vital-checks --mb-comments --syn-binding -fpsl --ieee=synopsys --std=08 --work=poc /home/paebbels/git/PoC/tb/common/my_project.vhdl
  command: /opt/ghdl/0.34-dev-llvm/bin/ghdl -a -fexplicit -frelaxed-rules --warn-binding --no-vital-checks --mb-comments --syn-binding -fpsl --ieee=synopsys --std=08 --work=poc /home/paebbels/git/PoC/tb/common/my_config_GENERIC.vhdl
  command: /opt/ghdl/0.34-dev-llvm/bin/ghdl -a -fexplicit -frelaxed-rules --warn-binding --no-vital-checks --mb-comments --syn-binding -fpsl --ieee=synopsys --std=08 --work=poc /home/paebbels/git/PoC/src/common/utils.vhdl
  command: /opt/ghdl/0.34-dev-llvm/bin/ghdl -a -fexplicit -frelaxed-rules --warn-binding --no-vital-checks --mb-comments --syn-binding -fpsl --ieee=synopsys --std=08 --work=poc /home/paebbels/git/PoC/src/common/config.vhdl
  command: /opt/ghdl/0.34-dev-llvm/bin/ghdl -a -fexplicit -frelaxed-rules --warn-binding --no-vital-checks --mb-comments --syn-binding -fpsl --ieee=synopsys --std=08 --work=poc /home/paebbels/git/PoC/src/common/math.vhdl
[...]

PR #214, opened 2016-12-05
This fixes issue #213.

Paebbels:
See explanations by @mzabeltud.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -69,7 +69,7 @@ OPT_FLAGS+=-fprofile-arcs -ftest-coverage
 endif
 
 GNATFLAGS=-gnaty3befhkmr -gnatwae -aO. -gnatf -gnat05 $(OPT_FLAGS) $(ADA_FLAGS)
-GRT_FLAGS=$(OPT_FLAGS)
+GRT_FLAGS=$(OPT_FLAGS) -fPIC
 
 WARN_CFLAGS=-Wall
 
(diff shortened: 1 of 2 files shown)


tgingold:
I think that relocPic should be selected only if needed.  Configure should check if gcc was configured with --enable-default-pie

Paebbels:
Hmmm the Windows GCC platforms (MinGW32, MinGW64, any other?) won't support PIC, because it's Linux specific, right? As far as we found out, PIC is supported by GCC since 4.9 (I have no older version to check it). Only the default seams to be changed since 6.2.

Is PIC a bad feature or performance killer, so it shouldn't be enabled by default?

-----------------
Distribution list: @mzabeltud

tgingold:
On 05/12/16 20:51, Patrick Lehmann wrote:
> Hmmm the Windows GCC platforms (MinGW32, MinGW64, any other?) won't
> support PIC, because it's Linux specific, right? As far as we found out,
> PIC is supported by GCC since 4.9 (I have no older version to check it).

I think -fpic/-fPIC is available for much longer, and on almost all 
platforms.

> Only the default seams to be changed since 6.2.

Indeed, and possibly only on debian.

> Is PIC a bad feature or performance killer, so it shouldn't be enabled
> by default?

Not a performance killer, but I prefer to enable it only when needed.
I am testing a patch.  Don't worry!

tgingold:
Should now be fixed (at least with llvm backend).

Issue #270, opened 2017-02-07
Dockerizing GHDL as a step towards the libre and portable VHDL design framework


1138-4EB:
After a chat with @Paebbels on [gitter](https://gitter.im/ghdl1/Lobby) last night, I open this issue to show the proof of concept of a possible alternative on how GHDL approaches Continuous Integration and Release publication.

I've known [Docker](https://www.docker.com/) for almost three years now, and I have been following the project with interest. I've sporadically run some containers, but neved went far from the examples/tutorials. Mostly because:

- Even though the idea is to sandbox the applications run in the containers, root privileges are required.
- The Windows version was run on VirtualBox, which, at first, didn't seem very different from actually using a VM.
-Most of the references I found where about web apps, which I'm not interested in. Moreover, it was not recommended for production, so I always left it apart.

Since it's been a while now, the project has improved. When I tried it againt last week, I was really pleased to know that the Windows version can run in the PowerShell (and in MSYS2 too). It seems not to be discouraged for [production](https://docs.docker.com/compose/production/) any more. And [security](https://docs.docker.com/engine/security/security/) issues are explained.

Inmediately GHDL came to my mind. On the one hand, realease 0.33 (2015-10-23) had multiple binaries and `*.deb` packages (darwin13, win32, debian, ubuntu), but later releases are provided for linux only. On the other hand, although [travis-ci](https://travis-ci.org/tgingold/ghdl) is used, no 'nightly builds' are published available. I think that docker can be really useful since a bunch of files suffice to automate the build and test process with multiple toolchains and/or library versions. Moreover, having a set of images as a reference can help with issues/bugs/reports, since the users can test their designs in a well known environment. Last, but not least, mutiple libre VHDL design resources can be distributed in a lighweight, modular and portable bundle.

# Objective

- Provide a single Dockerfile so that anyone can build GHDL and test cleanly in any computer supporting docker. This is explained [here](https://github.com/1138-4EB/ghdl-tools/blob/master/HOWTO_dockerize.md).
- Integrate it on travis-ci, so that the same environment available to users and developers is used.
- Optimize the workflow.

NOTE: You may want to jump to 'Proof of concept' at the end, and come back later.

# Tricking Travis CI

[...]

PR #272, opened 2017-02-09
Replace travis-ci matrix with docker containers

1138-4EB:
*NOTE: I commented the `deploy` section because I don't understand well how it works and I don't want to mess anything.*

This is a practical example derived from issue #270. I added a 4th option to the travis-ci matrix, which performs nearly the same operations as the 3rd, i.e., it compiles and tests GHDL with mcode. The main difference is that the added version runs in a docker container. Therefore:

- It is not Trusty, as in the 3rd, but `ubuntu:latest` (Xenial) from [docker-hub/ubuntu](https://hub.docker.com/_/ubuntu/).
- The `script` line in the `.travis.yml` calls `./dist/travis-ci.sh`, which is previously run in the docker container only in option 4. To avoid running it twice, I put an unexpecte content in $BLD, after the container is run. However, `.dist/travis-ci.sh` exits `1` when an unknown option is received. That's the reason of modifying `dist/travis-ci.sh `.
- Instead of putting the compilation commands in the `.travis.yml` file, I put them in a separate file `dist/docker-mcode-ubu_lat.sh`.
  - The script leaves a file with the format specifiec in line 19 `dist/docker-mcode-ubu_lat.sh` on the root. The content is equivalent to that on your releases (with libraries and include).
  - On the one hand, travis-ci suggests to follow this approach when launching 'complex commands'.
  - But the main reason is that, since it is dockerized, any user can just dowload that script and execute exactly the same toolchain that you, the developers, and the automated build system are using. It would be slightly more complicated if user had to extract it from the `.travis.yml`. An equivalent option is to use this Dockerfile:
```
FROM ubuntu:latest

ENV BLD mcode

WORKDIR /work

RUN apt-get update -qq \
 && apt-get -y --force-yes install gnat gcc zlib1g-dev make curl git \
 && git clone -b master --depth=1 http://github.com/tgingold/ghdl.git ./ \
 && curl -L https://github.com/1138-4EB/ghdl/raw/master/travis-ci.sh | sh
```
*NOTE: I tried to change permissions for `dist/docker-mcode-ubu_lat.sh` so that it is executable as `./dist/travis-ci.sh`, but I couldn't. I don't know if it can be because I'm on a Win10 machine.*
- The image which is actually being used is `ghdl/ghdl-tools:ubu_cmp-mcode`. It is just the fisrt two RUN lines in the Dockerfile above, as you can see [here](https://github.com/1138-4EB/ghdl-tools/blob/img-ubu_cmp-mcode/Dockerfile) ([size](https://hub.docker.com/r/ghdl/ghdl-tools/tags/)).
  - However, you can use the commented lines in `dist/docker-mcode-ubu_lat.sh` to use `ubuntu:latest instead`. The difference is having to download and install dependencies each time.
  - Execution time of options 3 and 4 is nearly the same, as you can see.
- On Win10 I haven't managed to shared volumens between the host and the containers properly, so it'll some more effort to publish it as a general solution (still not optimal, though).


tgingold:
I have never used docker, so I am a little bit lost!

On 09/02/17 07:03, 1138-4EB wrote:
> /NOTE: I commented the |deploy| section because I don't understand well
> how it works and I don't want to mess anything./

Please, do not comment that part.  This is used to upload to github the 
binaries when a branch or a tag is added.

> This is a practical example derived from issue #270
> <https://github.com/tgingold/ghdl/issues/270>. I added a 4th option to
> the travis-ci matrix, which performs nearly the same operations as the
> 3rd, i.e., it compiles and tests GHDL with mcode. The main difference is
> that the added version runs in a docker container. Therefore:

That part is ok.

>   * It is not Trusty, as in the 3rd, but |ubuntu:latest| (Xenial) from
>     docker-hub/ubuntu <https://hub.docker.com/_/ubuntu/>.
>   * The |script| line in the |.travis.yml| calls |./dist/travis-ci.sh|,
>     which is previously run in the docker container only in option 4. To
>     avoid running it twice, I put an unexpecte content in $BLD, after
>     the container is run. However, |.dist/travis-ci.sh| exits |1| when
>     an unknown option is received. That's the reason of modifying
>     |dist/travis-ci.sh|.

That's ok too.

>   * Instead of putting the compilation commands in the |.travis.yml|
>     file, I put them in a separate file |dist/docker-mcode-ubu_lat.sh|.
>       o The script leaves a file with the format specifiec in line 19
>         |dist/docker-mcode-ubu_lat.sh| on the root. The content is
>         equivalent to that on your releases (with libraries and include).

How and where ghdl/ghdl-tools is found ?

>       o On the one hand, travis-ci suggests to follow this approach when
>         launching 'complex commands'.
>       o But the main reason is that, since it is dockerized, any user
>         can just dowload that script and execute exactly the same
>         toolchain that you, the developers, and the automated build
>         system are using. It would be slightly more complicated if user
>         had to extract it from the |.travis.yml|. An equivalent option
>         is to use this Dockerfile:
>
> |FROM ubuntu:latest ENV BLD mcode WORKDIR /work RUN apt-get update -qq \
> && apt-get -y --force-yes install gnat gcc zlib1g-dev make curl git \ &&
> git clone -b master --depth=1 http://github.com/tgingold/ghdl.git ./ \
> && curl -L https://github.com/1138-4EB/ghdl/raw/master/travis-ci.sh | sh |

I think that should be documented other than in a commit.  What about 
adding a dist/README-docker.md file ?

> /NOTE: I tried to change permissions for |dist/docker-mcode-ubu_lat.sh|
> so that it is executable as |./dist/travis-ci.sh|, but I couldn't. I
> don't know if it can be because I'm on a Win10 machine./

I can do it.

>   * The image which is actually being used is
>     |ghdl/ghdl-tools:ubu_cmp-mcode|. It is just the fisrt two RUN lines
>     in the Dockerfile above, as you can see here
>     <https://github.com/1138-4EB/ghdl-tools/blob/img-ubu_cmp-mcode/Dockerfile>
>     (size <https://hub.docker.com/r/ghdl/ghdl-tools/tags/>).

Ah ok.  This is the origin of the image.  But how was it created ?

>       o However, you can use the commented lines in
>         |dist/docker-mcode-ubu_lat.sh| to use |ubuntu:latest instead|.
>         The difference is having to download and install dependencies
>         each time.
>       o Execution time of options 3 and 4 is nearly the same, as you can
>         see.
>   * On Win10 I haven't managed to shared volumens between the host and
>     the containers properly, so it'll some more effort to publish it as
>     a general solution (still not optimal, though).

Ok.

Just uncomment the deploy section in .travis.yml.  But a readme would be 
nice too.

Thanks,
Tristan.

1138-4EB:
> I have never used docker, so I am a little bit lost!

I started with a single script (a simplified version of `travis-ci.sh`). I wrote a How-to [here](https://github.com/1138-4EB/ghdl-tools/blob/master/HOWTO_dockerize.md). In a nutshell, the most important step is changing from `./script.sh` to `chmod u+w+x script.sh &&
docker run --name ghdl -itv $(pwd):/tmpDir:Z fedora bash -c 'cd /tmpDir && sh script.sh && exit'`:

- `docker`: call the daemon
- `run`: launch a new container from an existing image
- `--name ghdl`: name of the launched container/instance
- `-it`: make the session interactive (alternatively, containers can be lauched in the background with `-b`).
- `-v $(pwd):/tmpDir:Z`: create a dir name `tmpDir` in the container, and mount the hosts `$(pwd)` in it.
- `fedora`: the base image to be used for the container.
- `bash -c 'cd /tmpDir && sh script.sh && exit'`: the commands to execute as soon as the container is initialized.

[This](https://image.slidesharecdn.com/dockerinanutshell-dockeraustinjulymeetup-140824010119-phpapp02/95/docker-in-a-nutsell-6-638.jpg?cb=1408850763) image is pretty clarifying. The Host is either Travis or your own PC. The docker registry is [docker-hub](https://hub.docker.com/).

So, we are telling: docker, please, grab image `fedora` from wherever you can get it (i.e., hub.docker.com), make an instance/container of it, and launch these commands inside. The container will be seeing a fake and temporal filesystem. The only 'real' folder is the one we shared. Well, let's just believe it until we get how [volumes](https://docs.docker.com/engine/tutorials/dockervolumes/) work.

On you machine, check the difference between:

```
$ docker run --rm -it fedora bash -c 'dnf update -y && exit'
Fedora 25 - x86_64 - Updates                    3.8 MB/s |  18 MB     00:04
[...]
Complete!
$ uname -a
```

and

```
$ docker run --rm -it fedora bash -c 'dnf update -y'
Fedora 25 - x86_64 - Updates                    3.8 MB/s |  18 MB     00:04
[...]
Complete! 
$ uname -a
$ exit
$ uname -a
```
NOTES:
- On MSYS2, prepend `winpty` to `docker run -it` calls.
- I added `--rm` in order to remove container as soon as we exit. Then, its name is not useful.

Moreover, you can open four terminals and do:

```
$ docker run --name fed -d fedora bash \
&& docker run --name ubu -d ubuntu bash \
&& docker run --name alp -d alpine sh \
&& docker run --name cen -d centos bash
```

I will return an ID for each created container:
```
749d4a766e0171129c3ed264218c19ad0887d070ae507fa61ee2c71683ca65b3
e3e3e8dc065d1236efcf5de96c46ac76826b7f0c954d215e99f5bbc45d6ae278
7d2fb91b8c05e05c56d583e675204811a88defb6a36270578ddbf50c84fe3107
9195795c5e0e585aea97dc54541029588bc1dee7670aee74c99094e6c8fc4225

Then

$ docker ps -a
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS                           PORTS               NAMES
3499262d4a36        centos              "bash"                   34 hours ago        Exited (0) 39 seconds ago                            cen
e9a629babfd9        alpine              "sh"                     34 hours ago        Exited (0) About a minute ago                        alp
850edffbb649        ubuntu              "bash"                   34 hours ago        Exited (0) About a minute ago                        ubu
bacb134cb7cc        fedora              "bash"                   34 hours ago        Exited (0) 2 minutes ago                             fed
```

1138-4EB:
> Please, do not comment that part. This is used to upload to github the binaries when a branch or a tag is added.

I think I get how it is going now. Let me confirm. Each build generates a different `.tgz`. Since they run the same script, different VMs are used. Then `deploy` finds a single file in each working directory. Therefore, I've changed

```
docker cp ghdl_cmp:/work/install-mcode deploy
docker rm ghdl_cmp

tar -zcvf $(printf "$TRAVIS_EVENT_TYPE-$TRAVIS_BRANCH-%s.tgz" $(echo $TRAVIS_COMMIT | cut -c1-10)) deploy
rm -rf deploy 
```

to

```
PKG_TAG=`date -u +%Y%m%d`
PKG_VER=`curl -L https://raw.githubusercontent.com/tgingold/ghdl/master/src/version.in | grep Ghdl_Ver | sed -e 's/.*"\(.*\)";/\1/'`

docker cp ghdl_cmp:/work/ghdl-$PKG_VER-$BLD-$PKG_TAG.tgz ./
docker rm ghdl_cmp
```

However, since those tags are generated inside the container once, and outside again, this can produce errors at midnight. Shall I modify `travis_ci.sh` so that no tags are generated there, but passed as arguments?

> How and where ghdl/ghdl-tools is found ?

As explained in the previous comment, it is at [hub.docker.com/r/ghdl/ghdl-tools](https://hub.docker.com/r/ghdl/ghdl-tools). However, I wouldn't like to impose it. We have several options:

1. Use `ubuntu:latest` as explained above. This will install dependencies every time, instead of taking a preset image.
2. Copy the Dockerfile from `github.com/ghdl-tools/img-ubu_cmp-mcode` to your repo, sign up at docker-hub and create an automated repo there which will grab your Dockerfile and build an image for you.
3. Copy the Dockerfile, but build it at travis-ci instead, just before running the container.
  3.1 The result is similar to the first case.
  3.2 However, travis-ci's cache can be used to save images (not containers) between builds, so that the result equals the second case.
4. Transfer `hub.docker.com/r/ghdl` to you, or make you admin.
  4.1 Dockerfiles still in `github.com/ghdl-tools/` branches.
  4.2 Dockerfile(s) in your repo.

- 1, 2, 3, and 4.2 will give you full control of the process. 4.1 will require coordination with me (or make you contributor of ghdl-tools).
- 1 and 3.1 will install dependencies each time a build of GHDL is triggered. Images are smaller. Containers the same size.
- 2 and 3.2 will require to trigger image builds and GHDL builds separately. A cron job can be set to build images, or hooks can be used to check for major updates. Indeed, they can be daisy chained, so that images are rebuilt prior to any GHDL build.

> I think that should be documented other than in a commit. What about adding a dist/README-docker.md file ?

NOTE: I haved changed it so that it checks for `$CONTINUOUS_INTEGRATION`. If it is not present, the latest master from `tgingold/ghdl` is downloaded through `curl`, inside the container. This way the user does not even see the sources, just gets a tarball with the binaries.

What should I incude? Executing it locally is just `sh docker-mcode-ubu_lat.sh`. We can say that GHDL's dev-build installation is (replace `https://raw.githubusercontent.com/1138-4EB/ghdl/add-docker-proof-to-travis/` with `tgingold/ghdl/master` when the merge is done:

`curl -L https://raw.githubusercontent.com/1138-4EB/ghdl/add-docker-proof-to-travis/dist/docker-mcode-ubu_lat.sh | sh`

which takes 2 minutes :D.

Shall I put the long comment above and a copy of [1138-4EB/ghdl-tools/blob/master/HOWTO_dockerize.md](https://github.com/1138-4EB/ghdl-tools/blob/master/HOWTO_dockerize.md)? Or just this command?

NOTE: *If using MSYS on Win10, `winpty` must be prepended at line 6. So, the pipe must be split in two steps. On top of that, I am having problems with `docker cp` and shared folders on Win10 too. Everything works, except taking the content out of the container.

---

I have triggered serveral builds and I think that travis-ci only counts as build time that spent in `script` commands, no any other. Is it ok? If so, the solution with prebuild images is faster than actually installing packages.

1138-4EB:
#277 is already merged here. On top of that, the latest change adds the first 10 digits of the commit SHA, so that the filename of the tgz does not collide with the one generated in the 3rd build of travis.

1138-4EB:
#277 is already merged here. On top of that, the latest changes add:

- Remove previous travis-ci builds and dockerize them. Add fedora containers. There is a matrix of five containers now:
  - `ubu1404+llvm-3.5`: equivalent to previous `trusty` build [[Dockerfile](https://github.com/1138-4EB/ghdl-tools/blob/img-ubu1404_cmp-llvm/Dockerfile)]
  - `ubu1204+llvm-3.8`: equivalent to previous `precise` build [[Dockerfile](https://github.com/1138-4EB/ghdl-tools/blob/img-ubu1204_cmp-llvm/Dockerfile)]
  - `ubu+mcode`: equivalent to previous `mcode` build, but based on `ubuntu:latest` [[Dockerfile](https://github.com/1138-4EB/ghdl-tools/blob/img-ubu_cmp-mcode/Dockerfile)]
  - `fed+mcode`: based on `fedora:latest` [[Dockerfile](https://github.com/1138-4EB/ghdl-tools/blob/img-fed_cmp-llvm-mcode/Dockerfile)]
  - `fed+llvm`: same image as `fed+mcode`.
- Three characters identifying the image and the first 10 digits of the commit SHA are appended to the filename of the tgz, so that it does not collide with the other files generated with the same compiler but in a different container.
- The image to be used to instantiate the container is set with an envvar in `.travis.yml`. This image should have dependencies installed, so that `buildtest.sh` and `docker-buildtest.sh` can be mainainted image-agnostic.
- If the host dir is mounted in the container, `mv` is enough, no `docker cp` needed in `docker-buildtest.sh`.

1138-4EB:
- Moved `dist/appveyor`, `dist/gcc` and `dist/mcode` to `dist/win`.
- Moved all the added files in this commit to `dist/lin`. Created a file in `dist/lin/envs` for each build to be run in travis.

1138-4EB:
- All the names have been modified so that no shorthands are used: lin->linux, win->windows, fed->fedora, ubu->ubuntu,...
- Added conditional builds. Now only 'ubuntu_mcode' and 'fedora_llvm' are built on a regular basis. All the other are only run when `$TRAVIS_TAG` is not empty. Actually, all of them are always executed, but most of the skip the `script`. It would be interesting to effectively remove those, so that the VM are not even launched. However, I don't think it can be done in a single `.travis.yml` and a single branch.

Anyway, it is mature enough to be merged.

1138-4EB:
Just found how to share folders and volumen on windows with MinGW/MSYS2: https://github.com/1138-4EB/ghdl-tools/issues/8

I'm starting to move the content from this thread to a doc file. Shall I put it in `dist/linux/docker-README.md` or will I add it to `doc`? Although it is a merge related to CI, anyone can replicate it with a few commands.  Therefore, I think that a `Build` section in the documentation could be a good place.

tgingold:
On 13/02/17 21:00, 1138-4EB wrote:
> Just found how to share folders and volumen on windows with MinGW/MSYS2:
> 1138-4EB/ghdl-tools#8 <https://github.com/1138-4EB/ghdl-tools/issues/8>

Ok, although I don't see how.

> I'm starting to move the content from this thread to a doc file. Shall I
> put it in |dist/linux/docker-README.md| or will I add it to |doc|?
> Although it is a merge related to CI, anyone can replicate it with a few
> commands. Therefore, I think that a |Build| section in the documentation
> could be a good place.

Indeed, that would make sense.

I hope to be able to merge the pull request tomorrow (I have a pending 
patch from me I am reg-testing).

Thanks!

1138-4EB:
> Ok, although I don't see how.
Ouch! Forgot to click the 'comment' button. It's posted now!

> Indeed, that would make sense.
I may try mo put these together in the doc:

- BUILD.txt
- dist/mcode/README
- dist/gcc/INSTALL
- dist/gcc/README

And add some notes about building on windows (with[out] MinGW), plus some additional info (gcov only with gcc, backtraces optional -patchable-).

> I hope to be able to merge the pull request tomorrow (I have a pending patch from me I am reg-testing).
There should be no confilct. However, ping me in case there is, and I'll merge it for you, so that you can pull clean.

PR #275, opened 2017-02-10
target install.libghdlsynth added

Obijuan:
The target **install.libghdlsynth** was missing in the Makefile file, so the command make install.libghdlsynth did not work when following the instructions for building ghdlsynth


tgingold:
Do we want to install both the static and the shared library ?

Obijuan:
You are right. Maybe is it better to install only the shared library

PR #276, opened 2017-02-10
Avoid problems from not using quotes in shell script.

1138-4EB:
While developing #272 I had to add quotes to line `PKG_FILE="ghdl-$PKG_VER-$BLD-$PKG_TAG.tgz"`. Without them, the tarball generated in the container was `ghdl--mcode-20170210.tgz` ([see](https://api.travis-ci.org/jobs/200292577/log.txt?deansi=true)). Just in case, I added a few more.

### DIFF
--- a/dist/travis-ci.sh
+++ b/dist/travis-ci.sh
@@ -4,9 +4,9 @@
 # Stop in case of error
 set -e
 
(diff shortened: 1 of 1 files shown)


tgingold:
Thanks!

PR #277, opened 2017-02-11
Extract tag declarations from <travis-ci.sh>. Rename it to <buildtest…

1138-4EB:
….sh> and allow to pass -b  ($BLD) and -f ($PKG_FILE) as arguments. Save tags declarations in <travis-ci.sh>, with a different meaning now. Modify <.travis.yml> accordingly.

If $PKG_FILE is to be used in different scripts, it is safer to generate it just once, since it depends on 'date'.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,9 @@ matrix:
         apt:
           packages:
           - gnat
-script: ./dist/travis-ci.sh $BLD
+script:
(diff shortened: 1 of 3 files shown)


1138-4EB:
Fixed now.

tgingold:
Thanks!

Issue #280, opened 2017-02-13
Integrate build instructions into the GHDL documentation


Paebbels:
This is a collective issue to gather notes and workitems for the integration of the build documentation into the official documentation on ReadTheDocs.

I think we need a new table of contents structure like this:

* Introduction (Project overview, quick start)
* GHDL usage (explain all options and work flows)
* Getting GHDL (from package, distribution, GH releases, ...)
* Building instructions
  * GHDL
    * GCC backend
      * Linux
    * LLVM backend
      * Linux
      * Mac OS?
      * Windows MinGW 32/64
    * Mcode backend
      * Linux
      * Windows GNAT GPL
      * Windows MinGW 32/64
  * Containers
  * Vendor Primitives
    * Alters / Intel
    * Lattice
    * Xilinx ISE
    * Xilinx Vivado
* Contributing
* License

1138-4EB:
These are the files I found, which I think could be moved to the doc:

- BUILD.txt
- dist/mcode/README
- dist/gcc/INSTALL
- gcc/README

- dist/gcc/ANNOUNCE
- NEWS
- COPYING

However,

- I think that it is still interesting to keep a copy of 'COPYING' and a minimum version of 'BUILD.txt' in the root. Maybe just extending the section in the `./README` can be enough, instead of 'BUILD.txt'. At the same time, 'COPYING' can be written in markdown and copied to `doc` as 'License' on readthedocs deploy.
- It'd be very useful to add a README/INSTALL and COPYING to each release. Some parts of the `doc` can be used to generate these on travis-ci deploy.
- I'd add a specific 'NEWS' section (maybe renamed to 'Release log'), but keep the file in the root (renamed to `*.md`) so that it is copied on deploy.
- `dist/gcc/README`, `dist/gcc/INSTALL` and `dist/mcode/README` would go directly to the doc.
- No idea where to put `dist/gcc/ANNOUNCE`. But I think that it should be kept somewhere.

tgingold:
On 13/02/17 23:58, 1138-4EB wrote:

[...]

> However,
>
>   * I think that it is still interesting to keep a copy of 'COPYING' and
>     a minimum version of 'BUILD.txt' in the root. Maybe just extending
>     the section in the |./README| can be enough, instead of 'BUILD.txt'.
>     At the same time, 'COPYING' can be written in markdown and copied to
>     |doc| as 'License' on readthedocs deploy.

[...]

PR #294, opened 2017-02-21
New documentation structure: ch. Introduction

1138-4EB:
> Related to #280, note that the PR targets branch `new-documentation-structure`, not `master`.

I believe the first chapter is ready for you to review and give feedback: http://ghdl-devfork.readthedocs.io/en/new-documentation-structure/


1138-4EB:
The latest update fixes the sidebar for small screens (mobile devices). It also hides redundant info in the 'version' tab on the bottom-left. It reduces the font size slightly, and specially in the footer, where margins are also reduced. All of it is done with CSS, on top of the default RTD theme, through a file which I named `theme_overrides.css`.

Paebbels:
**Questions:**
- [ ] Why is the documentation content centered?  
  It looks strange on big screens.
- [ ] Why are selected navigation bar items indented?  
  That suggests a sub-level which does not exist.
- [ ] Do we want to use justified text in stead of ragged right?  
  The PoC documentation contains such a style fix.
- [ ] Page "About GHDL"
  - [ ] I would't write VHDL and GHDl every time in italic. Italic is usually used to introduce a term or wording. So italic is used the first time a word or wording appears, but not repeated.
  - [ ] directive `include` is broken (visible in text), thus `|trade|` doesn't work.
  - [ ] GHW points to a local GHDL parameter, whereas VCD points to wikipedia. (you can create an external link syntax for the English wikipedia if you like.)
  - [ ] Who uses GHDL:
    - [ ] **PoC** - A Vendor-Independent, Open-Source IP Core and Utility Library.  
      [GitHub shield] [RTD shield]
    - [ ] **VUnit** -   
      [GitHub shield] [RTD shield]
    - [ ] **IEEE P1076 WG** -   
      [GitLab shield]
- [ ] Page "Contributing"
  - [ ] I wouldn't use shields in the text. That's not the purpose of shields. I would collect the related shields between headline and text, which are referenced in the text.
    image
  - [ ] Maybe this blog post (or made available as a local copy, license?) should be linked in the fork/branching section: http://nvie.com/posts/a-successful-git-branching-model/
  - [ ] In the test: `See Directory structure ` uses the title of the reference. It should be either all in upper case like titles are written in US English or in lower case to fit into the text.
  - [ ] a list of actions can be written as:
    ```
    1. foo 
    1. foo
    ```
- License
  - [ ] The number of contributors is wrong - it's a bug in the shield itself.
  - [ ] Missing Authors:
    - [ ] @Jonsba - signal selection
    - [ ] @cclassic - vendor pre-compile script for Lattice (linux)

1138-4EB:
- [ ]  > Why is the documentation content centered?

That is the default behaviour of the RTD theme for screens more than `768px` wide but less than `max-width`, where `max-width` is somewhere between 768 and 1400. Above `1400px` the main container is aligned to the left. However, with Firefox (with Chrome too, I think), you can change it online. Right click at any point and select `Inspect element`. Then go to `theme_overrides.css` and change line 17:
- Centered: `.wy-nav-content{margin: auto; max-width: 1020px}`
- Aligned to the left: `.wy-nav-content{margin: 0; max-width: 1020px}`

I think it looks worse with large screens when it is not centered, than it does with small ones when it is. However, it's just a matter of taste. On mobile devices (small screens) it is always aligned to the left. The default sidebar took most of the screen, so I modified it so that its width does not change.

- [x] > Why are selected navigation bar items indented?

Should be fixed now. At least levels 1, 2 and 3.

- [ ] > Do we want to use justified text in stead of ragged right? The PoC documentation contains such a style fix.

Can you point me to the correct file?

- [x] > Page "About GHDL"

    - [x] > I would't write VHDL and GHDl every time in italic. Italic is usually used to introduce a term or wording. So italic is used the first time a word or wording appears, but not repeated.
    - [x] > directive `include` is broken (visible in text), thus `|trade|` doesn't work.
    - [x] > GHW points to a local GHDL parameter, whereas VCD points to wikipedia. (you can create an external link syntax for the English wikipedia if you like.)
There is no GHW article anywere. So I pointed to the single reference I found.
I added que external link syntax.
    - [x] Who uses GHDL:
        - [x] PoC
        - [x] VUnit
        - [x] IEEE P1076 WG
- [x] Page "Contributing"
    - [x] I wouldn't use shields in the text. That's not the purpose of shields. I would collect the related shields between headline and text, which are referenced in the text.
    - [x] Maybe this blog post (or made available as a local copy, license?) should be linked in the fork/branching section: http://nvie.com/posts/a-successful-git-branching-model/
    - [x] In the test: `See Directory structure` uses the title of the reference. It should be either all in upper case like titles are written in US English or in lower case to fit into the text.
    - [x] a list of actions can be written as

- [x] License
    - [x] The number of contributors is wrong - it's a bug in the shield itself.
Mine is showing the same number as this GitHub page (21). Maybe `ctrl+f5`? `img.shields.io` allows to `Set the HTTP cache lifetime in secs` with `?maxAge=3600`. Should I add it to every badge?
    - [x] Missing Authors:
        - [x] @Jonsba - signal selection
        - [x] @cclassic - vendor pre-compile script for Lattice (linux)

1138-4EB:
@tgingold activated `new-documentation-structure` version in RTD but it won't build. Leaving this note here just to remember: in RTD, settings, advanced settings, put `doc/requirements.txt` and uncheck the box that generates the EPUB version. Indeed, although the build failed, HTML and PDF version where being properly generated.

---

The merge that @Paebbels did this morning had a couple of minor issues (the gitter shield link being broken, and the scroll bar being visible in small screens). They are fixed in my `new-documentation-structure` branch, which is a single commit ahead. You can either pull, or just wait until I finish rewritting the next chapter and I'll PR for all together.

Paebbels:
It's not a productive branch nor a major issue, so I'll wait :)

PR #297, opened 2017-02-27
Improve Travis-CI/Docker scripts

1138-4EB:
Use single VM in `.travis.yml` to speed-up launching builds. Containers are launched in parallel threads, instead of initializing different machines even for the builds which are not run on a regular basis. To allow so:

- A copy of the working directory is made for each container.
- A numeric identifier is given to each thread.
- During the execution, most of the information is saved to a logfile for each thread. Minimum information is shown to avoid travis-ci breaking de build because of inactivity. Since these are not shown in order, the identifier is used.
- When each thread is finished, the log is copied to the `$HOME`, along with the pkg.
- When all of them are finished, all the full logs are printed in order.
- If any of the threads finished because of an error and, therefore, the buildtest script is not fully executed: the build is broken.
- Since I rewrote all the outputs to screen, I added color:
  - Yellow `[TRAVIS]` messages are generated in `travis-ci.sh`
  - Yellow `[$TASK| BUILD]` messages are generated in `docker-buildtest.sh`
  - Blue `[1| GHDL` messages are generated in `buildtest.sh`

See a passing example [here](https://travis-ci.org/1138-4EB/ghdl/builds/205681812), and a failing one [here](https://travis-ci.org/1138-4EB/ghdl/builds/205681525) (I made it fail, everything is OK except that the token is not written). Those are equivalent to the actual solution, so two builds are executed: ubuntu-mcode and fedora-llvm.

---

The usage is also easier, since I removed the `dist/linux/envs` folder. Now, only three files are used: `travis-ci.sh`, `docker-buildtest.sh` and `buildtest.sh`.

- An array in `travis-ci.sh` is used to set the list of available images:
```
images=("ghdl/ghdl-tools:ubuntu-mcode"    \
        "ghdl/ghdl-tools:ubuntu1404-llvm" \
	"ghdl/ghdl-tools:ubuntu1204-llvm" \
        "ghdl/ghdl-tools:fedora-llvm-mcode")
```
- Then, an array is defined for each 'matrix':
```
regular=("0+ubuntu+mcode" \
         "3+fedora+llvm")

nightly=("1+ubuntu1404+llvm-3.5" \
         "2+ubuntu1204+llvm-3.8" \
         "3+fedora+mcode")
```
The syntax is the following: image + identifier + compiler. Therefore, `0+ubuntu+mcode` will use image 0, that is, `ghdl/ghdl-tools:ubuntu-mcode`, and `mcode` will be passed as `$BLD` to `buildtest.sh`.
- Last, depending on the triggering conditions, a matrix is executed. Now, regular is selected only: `thismatrix=regular`. That is because the `nightly` builds are not implemente yet.

---

Quite a tricky problem I had is that colors worked on Fedora, but not on Ubuntu. Well, they worked in some scripts, but not in others. That's because `bin/sh` is set in all of them, and the default shell when launching a script in Ubuntu is `dash`, not `bash`. However, the commands in `.travis.yml` are executed in `bash`.

As a result, the scripts which are sourced in `.travis.yml` work without any problem. But those which are called, behave different. To fix it, I used `\033[` instead of `\e[` to set the colors, and changed `echo` to `printf`.

On top of that, arrays are a bash extension, so they can not be used in `dash`. Therefore, only `travis-ci.sh` uses arrays, and it must be sourced from `.travis.yml`, not called.

---

I think that the output of the logs could still be improved by modifying the `testsuite.sh` files in `testsuite/gna/*` and `testsuite/vests**/*`. But I don't feel confident enough to do so. Specially `vests`.


dist/linux/travis-ci.sh
```diff
@@ -1,12 +1,65 @@
...
+
+  IFS='+' read -ra REFS <<< "$2"
+  DBLD=${REFS[2]}
+  wrkd="../wrk-$m"
+  cp -r ./ "$wrkd" && cd "$wrkd"
+  ./dist/linux/docker-buildtest.sh -i "${images[${REFS[0]}]}" -t "$1" -b "$DBLD" -f "ghdl-$PKG_VER-$DBLD-$PKG_TAG-${REFS[1]}-$PKG_SHORTCOMMIT.tgz"
+  cd "$cloned"
+  cp $(ls "$wrkd"/ghdl-*.tgz) ./
+  cp $(ls "$wrkd"/log.log) "./log_$m.log"
+}
```
tgingold: Minor point: do you really need 'ls' here ? Why cp "$wrkd"/ghdl-*.tgz ./ doesn't work ?

testsuite/gna/testsuite.sh
```diff
@@ -32,27 +32,29 @@ for opt; do
...
     else
       exit 1;
     fi
   fi
   cd ..
+}
+
+for i in $dirs; do
+  printf "\033[34mGNA $i \033[0m\n" 1>> ../../log.log 2>&1
+  singlerun $i $full 1>> ../../log.log 2>&1
```
tgingold: Please, don't redirect in testsuite.sh (or add an option for that).
I often manually run then.

Not sure that colours are a good idea: AFAIK they don't work on Windows.
(An option is ok too).

(I understand progress is displayed on stdout and only test output is logged).

dist/linux/travis-ci.sh
```diff
@@ -1,12 +1,65 @@
...
-if [ -z "$BUILDTHIS" ]; then BUILDTHIS=$(echo "$TRAVIS_TAG"); fi
 if [ -z "$TRAVIS_TAG" ]; then PKG_TAG=`date -u +%Y%m%d`; fi
 
-export PKG_FILE="ghdl-$PKG_VER-$DBLD-$PKG_TAG-$PKG_DTAG-$PKG_SHORTCOMMIT.tgz"
\ No newline at end of file
+cloned=$(pwd)
+
+printf "\033[33m[TRAVIS] Running matrix $thismatrix \033[0m\n"
+eval blds=( '"${'${thismatrix}'[@]}"' )
+for m in $(eval echo "{0..$((${#blds[@]}-1))}"); do
```
tgingold: Not sure I understand this expression!  A comment would be welcome for such complex code.

tgingold:
General comment for colours: you'd better to define shell variables like RED, GREEN, NORMAL...
and use them.
Make that optional (not supported on windows IIRC).

dist/linux/travis-ci.sh
1138-4EB: Well, although it seems complex, it is actually quite stupid. The 'golang pseudocode' would be:
```
thismatrix=*regular
blds=&thismatrix
for m, b := range blds { go task(m,b) }
```
or, ideally,
```
thismatrix=*regular
for m, b := range &thismatrix { go task(m,b) }
```
---
- `thismatrix` contains the name of a var which is expected to be an array (say, `regular`, `nightly` or `release`).
- In line 47, eval is used to make `blds` be a copy of the array whose name is saved in `thismatrix`.
  - If we used `blds=${thismatrix[@]}`, we would copy all the letters of say `regular`. Therefore, `eval` is required.
- Line 48 is just a for loop to take each of the elements in `blds`. In this case, I feel that `eval` might not be required, but I couldn't find another working expression.
- Indeed, lines 47-48 could be merged, as in the pseudocode, but I don't find the shell easy to do so.

tgingold: Couldn't it be simplified a little bit:
```shell
for m in {0..$((${#blds[@]}-1))}; do
```

I think you don't need $(eval echo )

dist/linux/travis-ci.sh
1138-4EB: Certainly, I copypasted it from a previous version where I used a loop to modify `ghdl-*.tgz` names. But it is not needed now. I removed `ls`.

testsuite/gna/testsuite.sh
1138-4EB: > Please, don't redirect in testsuite.sh (or add an option for that).
I often manually run then.

Working on it.

> Not sure that colours are a good idea: AFAIK they don't work on Windows. (An option is ok too).

What do you mean with 'don't work on Windows'? I have been testing them in MinGW64 (Win10), and I had no problem with them. Indeed, since I launch the same containers as in travis-ci, when I had problems with dash, colours where not shown in the Ubuntu version neither locally nor in travis-ci. But in the Fedora version, they where shown both locally and in travis-ci. Now all work.

Since I am going to define variables for colorurs, it is trivial to make them optional. However, I'd like to know if you are referring to launching those shell scripts from the PowerShell. Is it possible?

> (I understand progress is displayed on stdout and only test output is logged).

I don't get this sentence. Do you mean you understand that this is the actual behaviour, or that you expect this to be the behaviour?

If by 'progress' you mean printing a single line for each test group, that's it. If you mean printing a line for each individual test, that's not the actual behaviour. I tried it, but since everything is printed at the end, it looked redundant to me.

dist/linux/travis-ci.sh
1138-4EB: This minimum example fails both in MinGW and in travis-ci:
```
blds=("0+ubuntu+mcode" "3+fedora+llvm")
for m in {0..$((${#blds[@]}-1))}; do echo "$m" "${blds[$m]}"; done
```
That's because braces are evaluated before the content between them. The eval avoids that:
```
for m in $(eval echo "{0..$((${#blds[@]}-1))}"); do echo "$m" "${blds[$m]}"; done
```
And according to http://unix.stackexchange.com/questions/55392/in-bash-is-it-possible-to-use-an-integer-variable-in-the-loop-control-of-a-for an alternative is to use a C-style for loop. But I don't know if that will affect portability.

tgingold:
On 27/02/17 20:58, 1138-4EB wrote:
> *@1138-4EB* commented on this pull request.
>
> ------------------------------------------------------------------------

[...]

>     Please, don't redirect in testsuite.sh (or add an option for that).
>     I often manually run then.
>
> Working on it.

Great.

>     Not sure that colours are a good idea: AFAIK they don't work on
>     Windows. (An option is ok too).
>
> What do you mean with 'don't work on Windows'? I have been testing them
> in MinGW64 (Win10), and I had no problem with them.

According to @Paebbel, ANSI escape sequences were not always supported 
by Windows console (at least before Windows 10).

 > Indeed, since I
> launch the same containers as in travis-ci, when I had problems with
> dash, colours where not shown in the Ubuntu version neither locally nor
> in travis-ci. But in the Fedora version, they where shown both locally
> and in travis-ci. Now all work.
>
> Since I am going to define variables for colorurs, it is trivial to make
> them optional.

That would be fine.

 > However, I'd like to know if you are referring to
> launching those shell scripts from the PowerShell. Is it possible?

Not from PowerShell.  I was referring to run them through bash.

>     (I understand progress is displayed on stdout and only test output
>     is logged).
>
> I don't get this sentence. Do you mean you understand that this is the
> actual behaviour, or that you expect this to be the behaviour?

I mean this is the intent of your code: only one line is displayed per 
test, and the remaining is sent to the log file.

> If by 'progress' you mean printing a single line for each test group,
> that's it.

Ok.

 > If you mean printing a line for each individual test, that's
> not the actual behaviour. I tried it, but since everything is printed at
> the end, it looked redundant to me.

Ok.

Thanks!
Tristan.

tgingold:
On 27/02/17 21:09, 1138-4EB wrote:
> *@1138-4EB* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In dist/linux/travis-ci.sh
> <https://github.com/tgingold/ghdl/pull/297#discussion_r103299877>:
>
>> \ No newline at end of file
> +cloned=$(pwd)
> +
> +printf "\033[33m[TRAVIS] Running matrix $thismatrix \033[0m\n"
> +eval blds=( '"${'${thismatrix}'[@]}"' )
> +for m in $(eval echo "{0..$((${#blds[@]}-1))}"); do
>
> This minimum example fails both in MinGW and in travis-ci:
>
> |blds=("0+ubuntu+mcode" "3+fedora+llvm") for m in
> {0..$((${#blds[@]}-1))}; do echo "$m" "${blds[$m]}"; done |
>
> That's because braces are evaluated before the content between them. The
> eval avoids that:
>
> |for m in $(eval echo "{0..$((${#blds[@]}-1))}"); do echo "$m"
> "${blds[$m]}"; done |

Ah ah.  Ok.

> And according to
> http://unix.stackexchange.com/questions/55392/in-bash-is-it-possible-to-use-an-integer-variable-in-the-loop-control-of-a-for
> an alternative is to use a C-style for loop. But I don't know if that
> will affect portability.

I would prefer the use of seq:
$(seq 0 $((${#blds[@]}-1)))

Tristan.

1138-4EB:
>I would prefer the use of seq: `$(seq 0 $((${#blds[@]}-1)))`

Actually, we have four options (I edited again, sorry, bat habit):

- `eval`
- C-style
- `seq`
- POSIX `while`

Because of portability, I used the last one, instead of `seq` (which is not a built-in). I also added comments:

``` shell
# Copy array pointed by `thismatrix`
eval blds=( '"${'${thismatrix}'[@]}"' )
# Get numbers of builds in the array
bnum=${#blds[@]};
# Spawn builds
m=0; while [ $m -lt $bnum ]; do
  task "$m" "${blds[$m]}" &
  m=$(($m+1));
done
```
and
``` shell
# For each build.
m=0; while [ $m -lt $bnum ]; do
  # Read the last line of the log
  RESULT[$m]=$(awk '/./{line=$0} END{print line}' "log_$m.log")
  # If it did not end with [$m|SUCCESSFUL], break the build
  if [ "${RESULT[$m]}" != "[$m|SUCCESSFUL]" ]; then EXITCODE=$(($m+1)); fi
  # Anyway, always print the full log
  printf "\033[33m[TRAVIS] Print BUILD $m log \033[0m\n"
  cat "log_$m.log"
  m=$(($m+1));
done
```

---

Still working on:

> Please, don't redirect in testsuite.sh (or add an option for that). I often manually run then.

and defining variables for colors. @Paebbels , could you point us to the context in which you have problems with colors? Are those generated with `echo -e "\e` or with `printf "\033[`?

1138-4EB:
- Colors have been replaced with envvars, and they are optional now:
  - If var `DISCOLOR` is unset or empty, colors are shown in `travis-ci.sh`. To disable them, set `DISCOLOR="-d".
  - `docker-buildtest.sh` and `buildtest.sh` will not show colors if opt `-d` is passed.
    - If `docker-buildtest.sh` receives `-d` from `travis-ci.sh`, it is passed to `buildtest.sh`.
  - `testsuite/testsuite.sh` checks for `DISCOLOR`. If not empty, colors are disabled.
    - Since `testsuite.sh` is sourced in `buildtest`, `DISCOLOR` is set if opt `-d` is received.
- Nothing is redirected in `testsuite/gna/testsuite.sh`. But it is done in `testsuite/testsuite.sh` and `testsuite/vests/testsuite.sh`. Is that correct? Do you also run `vests` manually?

Paebbels:
Hello,

here are a few general comments from my side:
* `-d` is a bad option name to disable coloring, because `-d` stands generally for debug.
* I'm against short identifiers and variable names like `DISCOLOR`, `bnum` ...
  Variables and identifiers are self describing items of a program or script. If you use good names you won't need comments for what you are doing. You shouldn't need to comment your code at all if you use good structure and naming. So what is for example `DISCOLOR`? disable or display color?
* You shouldn't use this crippled shell called `dash`. Just use `bash` on all machines and we are fine.
* If you look into my scripts (e.g. `compile-xxx.sh`) you'll find the color codes in variables.
* Why do you need `printf` instead of `echo`? My usage of echo works as expected.
* Windows shells don't understand ANSI escape codes. So coloring in them doesn't work and results in displaying the escape sequences.

1138-4EB:
@tgingold, if you need to see the output of `testsuite/testsuite.sh` when you manually run it, may this be feasible?
``` shel
$ ./testsuite.sh &
$ cd .. && tail -f log.log
...
...
...
Ctrl+C
$ rm log.log
```

1138-4EB:
> `-d` is a bad option name to disable coloring, because `-d` stands generally for debug.

How about changing the switch from 'disable' to 'enable' and use `-c` ?

> I'm against short identifiers and variable names like DISCOLOR, bnum ...

Changed:
- `DISCOLOR` to `ENABLECOLOR`.
- `bnum` to `numberofbuilds`.
- `wrkd` to `thisworkdir`.

Any other?

> You shouldn't use this crippled shell called dash. Just use bash on all machines and we are fine.
> Why do you need printf instead of echo? My usage of echo works as expected.

It is not Ubuntu's `dash` only. Alpine Linux does not have `bash` neither. `printf` is more portable than `echo`: http://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo

> If you look into my scripts (e.g. compile-xxx.sh) you'll find the color codes in variables.

You mean those in 'Poc/tools/precompile' not the `*.ps1` in 'ghdl/dist/windows', isn't it?

> Windows shells don't understand ANSI escape codes. So coloring in them doesn't work and results in displaying the escape sequences.

Sorry, but I don't get it. I just tried these:
```
echo -e "\e[33mHi\e[0m"
echo -e "\033[33mHi\033[0m"
printf "\e[33mHi\e[0m"
printf "\033[33mHi\033[0m"
```
in `MINGW64_NT-10.0 2.6.1(0.305/5/3) 2017-01-25 00:01 x86_64 Msys`. All of them are correctly shown.

Paebbels:
I meant e.g. `/library/vendor/compile-altera.sh` in GHDL.

I assume you started a MinGW console (with is a bash) and executed your command in that bash, right? What I mean is, if you run PowerShell.exe and start a Linux script in PowerShell. It executes, because it's launched with the Linux interpreter, but the colors are not converted.

1138-4EB:
> I meant e.g. /library/vendor/compile-altera.sh in GHDL.

Followed a similiar approach now.

>I assume you started a MinGW console (with is a bash) and executed your command in that bash, right? What I mean is, if you run PowerShell.exe and start a Linux script in PowerShell. It executes, because it's launched with the Linux interpreter, but the colors are not converted.

That's what I did, yes. I didn't have the linux subsystem for windows installed. I installed it now. Then, I put the following script in a file and executed `bash file.sh`:
```
echo -e "\e[33mHi\e[0m"
echo -e "\033[33mHi\033[0m"
printf "\e[33mHi\e[0m\n"
printf "\033[33mHi\033[0m"
```
No color is shown, but escape sequences are not shown neither. So, I'm afraid that this is not the same context you are describing.

1138-4EB:
Examples: [without color](https://travis-ci.org/1138-4EB/ghdl/builds/206021972) and [with color](https://travis-ci.org/1138-4EB/ghdl/builds/206023414).

tgingold:
On 27/02/17 21:44, 1138-4EB wrote:
>     I would prefer the use of seq: |$(seq 0 $((${#blds[@]}-1)))|
>
> Actually, we have four options (I edited again, sorry, bat habit):
>
>   * |eval|
>   * C-style
>   * |seq|
>   * POSIX |while|
>
> Because of portability, I used the last one, instead of |seq| (which is
> not a built-in). I also added comments:
>
> # Copy array pointed by `thismatrix`
> eval blds=( '"${'${thismatrix}'[@]}"' )
> # Get numbers of builds in the array
> bnum=${#blds[@]};
> # Spawn builds
> m=0; while [ $m -lt $bnum ]; do
>   task "$m" "${blds[$m]}" &
>   m=$(($m+1));
> done
>
> and
>
> # For each build.
> m=0; while [ $m -lt $bnum ]; do
>   # Read the last line of the log
>   RESULT[$m]=$(awk '/./{line=$0} END{print line}' "log_$m.log")
>   # If it did not end with [$m|SUCCESSFUL], break the build
>   if [ "${RESULT[$m]}" != "[$m|SUCCESSFUL]" ]; then EXITCODE=$(($m+1)); fi
>   # Anyway, always print the full log
>   printf "\033[33m[TRAVIS] Print BUILD $m log \033[0m\n"
>   cat "log_$m.log"
>   m=$(($m+1));
> done

OK for me.

Tristan.

tgingold:
On 27/02/17 22:35, 1138-4EB wrote:
>   * Colors have been replaced with envvars, and they are optional now:
>       o If var |DISCOLOR| is unset or empty, colors are shown in
>         |travis-ci.sh|. To disable them, set `DISCOLOR="-d".
>       o |docker-buildtest.sh| and |buildtest.sh| will not show colors if
>         opt |-d| is passed.
>           + If |docker-buildtest.sh| receives |-d| from |travis-ci.sh|,
>             it is passed to |buildtest.sh|.
>       o |testsuite/testsuite.sh| checks for |DISCOLOR|. If not empty,
>         colors are disabled.
>           + Since |testsuite.sh| is sourced in |buildtest|, |DISCOLOR|
>             is set if opt |-d| is received.

Ok.

>   * Nothing is redirected in |testsuite/gna/testsuite.sh|. But it is
>     done in |testsuite/testsuite.sh| and |testsuite/vests/testsuite.sh|.
>     Is that correct? Do you also run |vests| manually?

Works for me.
I sometimes run vests manually, but the output is not that interesting.

tgingold:
On 27/02/17 23:01, Patrick Lehmann wrote:
> Hello,
>
> here are a few general comments from my side:
>
>   * |-d| is a bad option name to disable coloring, because |-d| stands
>     generally for debug.

May I suggest --nocolor

>   * I'm against short identifiers and variable names like |DISCOLOR|,
>     |bnum| ...
>     Variables and identifiers are self describing items of a program or
>     script. If you use good names you won't need comments for what you
>     are doing. You shouldn't need to comment your code at all if you use
>     good structure and naming. So what is for example |DISCOLOR|?
>     disable or display color?

NOCOLOR or COLORS=no is clear.

>   * You shouldn't use this crippled shell called |dash|. Just use |bash|
>     on all machines and we are fine.
>   * If you look into my scripts (e.g. |compile-xxx.sh|) you'll find the
>     color codes in variables.
>   * Why do you need |printf| instead of |echo|? My usage of echo works
>     as expected.
>   * Windows shells don't understand ANSI escape codes. So coloring in
>     them doesn't work and results in displaying the escape sequences.

1138-4EB:
About DISCOLOR. The functionality is the same, but note that I changed it to ENABLECOLOR, and `-d` to `-c`, after the comment by @Paebbels.

tgingold:
On 27/02/17 23:09, 1138-4EB wrote:
> @tgingold <https://github.com/tgingold>, if you need to see the output
> of |testsuite/testsuite.sh| when you manually run it, may this be feasible?
>
> |$ ./testsuite.sh & $ cd .. && tail -f log.log ... ... ... Ctrl+C $ rm
> log.log |

I almost never run testsuite/testsuite.sh, so that's ok.

1138-4EB:
> May I suggest --nocolor

I used `-c`, because I don't know how to add 'long' options with `optargs`. However, if you think that it is required, I'll check how to do it. If so, I'll add long options to the others (`-i`, `-b`, `-f``, `-g` and `-t`).

tgingold:
On 28/02/17 07:47, 1138-4EB wrote:
>     May I suggest --nocolor
>
> I used |-c|, because I don't know how to add 'long' options with
> |optargs|. However, if you think that it is required, I'll check how to
> do it. If so, I'll add long options to the others (|-i|, |-b|,
> |-f``,|-g|and|-t`).

Fine with me.

Paebbels:
Handling of short and long options can also be found in that `compile-altera.sh` scripts :).

1138-4EB:
Long options are also accepted now. I added a conversion before `getopts`:
```
for arg in "$@"; do
  shift
  case "$arg" in
    "--color"|"-color")   set -- "$@" "-c";;
    "--grab"|"-grab")     set -- "$@" "-g";;
    "--image"|"-image")   set -- "$@" "-i";;
    "--build"|"-build")   set -- "$@" "-b";;
    "--file"|"-file")     set -- "$@" "-f";;
    "--taskid"|"-taskid") set -- "$@" "-t";;
    *) set -- "$@" "$arg"
  esac
done
```
@Paebbels, I checked `compile-altera.sh`. You don't use `getopts`, as [here](http://stackoverflow.com/questions/402377/using-getopts-in-bash-shell-script-to-get-long-and-short-command-line-options). Instead, I followed [this](http://stackoverflow.com/questions/12022592/how-can-i-use-long-options-with-the-bash-getopts-builtin) approach. I believe the functionality is the same.
---
@Paebbels, I adapted `compile-*.sh` so that the same `ANSI_COLOR` envvars are used both in `libraries/vendors` and in `dist/linux`. I kept the names you used, in order to reduce modifications. The only difference is that they now source an additional file, instead of taking the colors from `shared.sh`.

tgingold:
Is it ok for libraries/vendors script to source dist/ansi-color.sh ?
I think they are installed, but ansi-color.sh isn't.

@Paebbels ?

tgingold:
Vendors scripts are installed by target install.vhdllib of Makefile
I would either let vendors/ untouched or install ansi-color.sh; but I'd like to know @Paebbels opinion.

tgingold:
Copying Paebbels answer:
Please install them into the lib/GHDL folder and adjust the paths. The scripts are only executed within the installation directory, never in the source directory.

Paebbels:
Please install that file too and maybe adjust the paths.

1138-4EB:
> [Paebbels] Please install them into the lib/GHDL folder and adjust the paths. The scripts are only executed within the installation directory, never in the source directory.
> [1138-4EB] then, since I took that piece from shared.sh, I have to modify the Makefile so that ansi_color.sh is installed. Is it correct?
> [Paebbels] Yes.

I'm doing it.

PS: please remove the last five 'duplicated' comments in the other thread.

tgingold:
On 01/03/17 06:18, 1138-4EB wrote:
>     [Paebbels] Please install them into the lib/GHDL folder and adjust
>     the paths. The scripts are only executed within the installation
>     directory, never in the source directory.
>     [1138-4EB] then, since I took that piece from shared.sh, I have to
>     modify the Makefile so that ansi_color.sh is installed. Is it correct?
>     [Paebbels] Yes.
>
> I'm doing it.
>
> PS: please remove the last five 'duplicated' comments in the other thread.

Done.

1138-4EB:
I think it's ready now. @Paebbels?

1138-4EB:
@tgingold, I saw that the one of the builds of the release you did earlier failed. I think that it can have some relation to the fact that multiple instances where being run: more than one could be trying to push the release at the same time.

Anyway, that won't happen after this merge, since now a single Travis VM is run.

---

A side question: I realized that the LLVM `tgz` built in Fedora is half the size of those built in Ubuntu. Does that make any sense to you?

tgingold:
On 01/03/17 08:12, 1138-4EB wrote:
> @tgingold <https://github.com/tgingold>, I saw that the one of the
> builds of the release you did earlier failed. I think that it can have
> some relation to the fact that multiple instances where being run: more
> than one could be trying to push the release at the same time.

I think this is related to S3 failures.

> Anyway, that won't happen after this merge, since now a single Travis VM
> is run.

PR #334, opened 2017-04-15
Add mechanism to restrict signals dumped by ghwdump

emogenet:
ghwdump is sometimes useful beyond debugging .ghw files

For example, it can be used to implement external unit testing of a vhdl module by dumping a set of signals post-sim and checking with some other utility that some global properties of the set of dumped signals do hold true.

The current version of ghwdump -s dumps *all* signals, which, for a complex circuit can be very slow.

This patch adds a "-f" flag which lets the user choose which signals are dumped.

Example:
    ghwdump -s -f 10 wave.ghw    # Dump only signal number 10
    ghwdump -s -f 1-32,15,23-40  # Dump signal 1 to 32, signal 15, signals 23 to 40 (for ranges, bounds are included)

### DIFF
--- a/src/grt/ghwdump.c
+++ b/src/grt/ghwdump.c
@@ -18,6 +18,7 @@
 */
 
 #include <stdio.h>
+#include <assert.h>
 #include <stdint.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


tgingold:
Hello,

interesting feature.  I am merging it, but I'd like two changes:
* please use the same coding style (the GNU one), in particular the layout of function declarations,
* instead of adding a field in struct ghw_sig (which wastes space for other users of the lib), create a separate vector of flags indexed by the signal number.

It is possible to select which signals are dumped, maybe you can use that too.

emogenet:
I didn't know that you could select which signals are dumped. Is that during simulation? Which option would that be? I didn't manage to find this in the docs. If that does indeed exist, then it makes this patch completely moot.

tgingold:
On 16/04/17 12:34, Emmanuel Mogenet wrote:
> I didn't know that you could select which signals are dumped. Is that
> during simulation? Which option would that be? I didn't manage to find
> this in the docs. If that does indeed exist, then it makes this patch
> completely moot.

The option is --read-wave-opt=FILENAME
You can also generate one using --write-wave-opt=FILENAME

You can select signals by naming them, using this syntax:
/top/blk1/sig

Tell me if that works for you.

Tristan.

emogenet:
It does, thanks for the pointer.
I opened pull request #336 that implements your suggestions.

PR #337, opened 2017-04-19
Fix spelling error: declarared => declared

silkeh:
I found a (very) minor spelling error while debugging. Here is a fix!


tgingold:
Thanks!

Issue #26, opened 2016-01-24
Missing source code line number.


Paebbels:
I managed to calculate negative delay times, which causes GHDL to stop with an error.

> ghdl.exe:error: negative timeout clause
> ghdl.exe:error: simulation failed

Is it possible to report the VHDL source code line? I suspect it's in one of my procedures....

tgingold:
That should be doable...

Brian-Drummond:
I've had a brief look at this, and I can report the name of the erroneous process, but I can't offhand see how to get closer than that ( I don't know how source code locations are handled at runtime).

Here's a brief reproducer


```vhdl
entity Time_travel is
end Time_travel;

architecture Behavioral of Time_travel is

constant start_time : time := 10 ms; 
constant stop_time : time := 5 ms; 

procedure delay(t : time) is
begin
	wait for t;
end delay;

begin

McFly : process is
begin
        wait for start_time;
        report " Started " severity NOTE;        
	delay(stop_time - start_time);
--        wait for stop_time - start_time;
        report " Stopped " severity NOTE;   
        wait;
end process;
end Behavioral;
```
And two patches 27ea531 and 1f4105e at
https://github.com/Brian-Drummond/ghdl/commits/master
as a tentative fix.

tgingold:
The location of the wait statement can be passed to ghdl_process_wait_timeout and ghdl_process_wait_set_timeout.
(See how it is done in Gen_Program_Error: filename and line number are passed to the subprogram).

Brian-Drummond:
Thanks for the hints, I'll have another look.

PR #429, opened 2017-10-05
#26 revisited

Brian-Drummond:
Filename and line number added to error report, following Tristan's suggestion.


tgingold:
I think you should rebase your branch, otherwise the continuous builders fail.

src/grt/grt-errors.adb
```diff
@@ -254,6 +254,20 @@ package body Grt.Errors is
...
+                    Filename : Ghdl_C_String;
+                    Line : Ghdl_I32) is
+   begin
+      Error_H;
+      Put_Err (Str);
+      Put_Err (" at ");
+      Put_Err (Filename);
+      Put_Err (" line ");
+      Put_I32 (Error_Stream, Line);
+      Newline_Err;
```
tgingold: grt.lib.ghdl_program_error only uses Error_* subprograms, and I think this is correct.
Might be worth creating a subprogram for filename + line.

Brian-Drummond: OK thanks, please ignore for now (I need to figure out rebase!)

Brian-Drummond: Tried rebasing, we'll see...
I thought exporting an Error subprogram that understood location could be useful for other runtime errors ... (note the `Put_Err` calls in the snippet above are only used in grt_errors). And some other subprograms in grt.lib use Error not Error_*, ... or I may be misunderstanding your point?

PR #433, opened 2017-10-06
Support LLVM 5.0

Bobo1239:
It passes the testsuite but as I'm not actively using ghdl I don't know if there are other possible problems not covered by the tests. Also not sure what `llvm_be` does...


tgingold:
Thanks!
llvm_be just select the interface with llvm.

PR #436, opened 2017-10-09
Add dummy UPF package

tsailer:
The UPF package allow power aware testbenches to control the supply voltages of the power domains of a design. Now ghdl isn't power aware, but it would still be useful to have a dummy UPF package that allows testbenches to be shared without modification across different simulators.

Unfortunately, there doesn't seem to be a formal specification of this VHDL UPF package. There is a system verilog specification in IEEE Standard 1801-2015, Chapter 11.2. This SV definition has been translated to VHDL by the commercial hdl simulation vendors, unfortunately not in a very compatible way.

This package provides a dummy implementation of the portable subset of UPF.


Paebbels:
UPF was/is an Accellera standard. So what's the copyright of these files and what license is used?

1138-4EB:
What's the status of this PR? Is it enough with moving the proposed sources to openieee and openieee/v08?

PR #481, opened 2017-12-09
fix version path when .git exists but git is not installed

### LINKED ISSUE #480: [build-system] do not require git in the Makefile
Right now, if the repo is cloned through git, it becomes a 'soft' build dependecency because https://github.com/tgingold/ghdl/blob/master/Makefile.in#L143-L150

`make` finds a folder named `.git` and it assumes that git exists. If it does not, no error is raised, but var `$desc` is empty. This produces the following result when `GHDL --version` is executed:

```
GHDL 0.35-dev () [Dunoon edition]
```

Steps to reproduce:

- (git) Clone tgindold/ghdl.
- Copy the folder to a machine without git installed (say mount the folder in a docker container).
- Build GHDL.

I realized about this in #477, as you can see in this animated gif: https://user-images.githubusercontent.com/6628437/33694969-2e7b7030-dafb-11e7-9eba-fb3abae1a161.gif

The workaround I implemented is:

``` bash
version.tmp: $(srcdir)/src/version.in force
#	Create version.tmp from version.in, using git date/hash
	if [ "`command -v git`" != "" ]; then \
		desc=`cd $(srcdir); git describe --dirty`; \
		$(SED) -e "s/[(].*[)]/($$desc)/" < $< > $@; \
	else \
		if [ "${GIT_DESCRIBE_DIRTY}" != "" ]; then \
			$(SED) -e "s/[(].*[)]/($$GIT_DESCRIBE_DIRTY)/" < $< > $@; \
		else \
			$(CP) $< $@; \
		fi \
	fi
```

Along with changing https://github.com/tgingold/ghdl/blob/master/dist/linux/travis-ci.sh#L78 to:

```
git fetch --unshallow
docker run --rm --tty --volume $(pwd):/work -w "/work" $DOCKER_NAME bash -c "export GIT_DESCRIBE_DIRTY=$(git describe --dirty --always); $BUILD_CMD"
```

Note the need of `git fetch --unshallow`: travis-ci/travis-ci#3412

---

However, I just solved one problem to introduce a different one. Now, any repo which does not contain a `.git` folder will fail if git command is found in the machine. I though about:

- Making the envvar take precedence over checking the command. This would make it necessary to define the envvar when no `.git` folder is present. At the same time, it would provide a mechanism to overwrite the value even when git and .git are found.
- Add two conditions to the first if: `$(command -v git) != "" && -d "$(srcdir)/.git"`. This will make the envvar a fallback.

Yet, because I am not confident at all with the makefile. I prefer to discuss it first.

1138-4EB:
Close #480

This is not ready to merge: https://travis-ci.org/1138-4EB/ghdl/builds/313989310

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -142,11 +142,15 @@ include $(GRTSRCDIR)/Makefile.inc
 
 version.tmp: $(srcdir)/src/version.in force
 #	Create version.tmp from version.in, using git date/hash
-	if [ -d $(srcdir)/.git ]; then \
+	if [ "`command -v git`" != "" ] && [ -d $(srcdir)/.git ]; then \
 	  desc=`cd $(srcdir); git describe --dirty`; \
 	  $(SED) -e "s/[(].*[)]/($$desc)/" < $< > $@; \
 	else \
-	  $(CP) $< $@; \
+	  if [ "${GIT_DESCRIBE_DIRTY}" != "" ]; then \
+	    $(SED) -e "s/[(].*[)]/($$GIT_DESCRIBE_DIRTY)/" < $< > $@; \
+	  else \
+	    $(CP) $< $@; \
+	  fi \
 	fi
 
 version.ads: version.tmp
--- a/dist/linux/travis-ci.sh
+++ b/dist/linux/travis-ci.sh
@@ -74,7 +74,8 @@ else
 
 
     # Run build+test in docker
-
+    git fetch --unshallow
+    make -f Makefile.in version.tmp
     docker run --rm --tty --volume $(pwd):/work -w "/work" $DOCKER_NAME bash -c "$BUILD_CMD"
 fi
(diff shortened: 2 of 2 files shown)


tgingold:
I have just committed a similar patch.

1138-4EB:
Just checked it. Is there any reason not to include the condition `"$(command -v git) != ""`, along with checking `.git`?

#480 will still produce `()` without showing any error if this is run:

```
docker run --rm --tty --volume $(pwd):/work -w "/work" $DOCKER_NAME bash -c "./dist/linux/buildtest.sh -b mcode"
```

which is a reasonable command for someone willing to build GHDL locally inside a container.
```

tgingold:
Are you sure about that ?  If git is not available (or fail to describe), then nothing is changed in the version file.

PR #488, opened 2017-12-12
use travis matrix expansion

1138-4EB:
This PR just uses travis' built-in matrix expansion to slightly reduce the verbosity of `.travis.yml`. The produced jobs are the same:

- Before: https://travis-ci.org/1138-4EB/ghdl/builds/315555203
- After: https://travis-ci.org/1138-4EB/ghdl/builds/315562326

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,49 +2,13 @@ branches:
   except:
   - appveyor
 
-language: c
-
-matrix:
-  include:
-    - os: linux
-      dist: trusty
-      sudo: required
-      services: docker
-      env: IMAGE=stretch+mcode
-    - os: linux
-      dist: trusty
-      sudo: required
-      services: docker
-      env: IMAGE=stretch+mcode+gpl
-    - os: linux
-      dist: trusty
-      sudo: required
-      services: docker
-      env: IMAGE=ubuntu14+mcode
-    - os: linux
-      dist: trusty
-      sudo: required
-      services: docker
-      env: IMAGE=ubuntu14+llvm-3.8
-    - os: linux
-      dist: trusty
-      sudo: required
-      services: docker
-      env: IMAGE=fedora26+mcode
-    - os: osx
-      osx_image: xcode7.3
-      env: IMAGE=macosx+mcode
-      cache:
-        directories:
-          - gnat
-
-install: true
-
-script:
-- ./dist/linux/travis-ci.sh
-
-before_deploy:
-- "echo ready to deploy"
+os: linux
+dist: trusty
+sudo: required
+services: docker
+language: minimal
+install: skip
+script: "./dist/linux/travis-ci.sh"
 
 deploy:
   - provider: releases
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,49 +2,40 @@ branches:
...
+sudo: required
+services: docker
 install: true
 
-script:
-- ./dist/linux/travis-ci.sh
+addons:
+  apt:
+    packages:
+      - docker-ce
```
tgingold: Do you need docker-ce for this patch ?

.travis.yml
```diff
@@ -2,49 +2,40 @@ branches:
...
+    env: IMAGE=macosx+mcode
+    cache:
+      directories:
+        - gnat
+
+script: "./dist/linux/travis-ci.sh"
 
-before_deploy:
-- "echo ready to deploy"
+before_deploy: "echo ready to deploy"
```
tgingold: Let remove it

.travis.yml
```diff
@@ -58,5 +49,4 @@ deploy:
       tags: true
       all_branches: true
 
-after_deploy:
-- "echo deployed"
+after_deploy: "echo deployed"
```
tgingold: Let remove it too.

tgingold: I wonder if it is worth.
Yes it is shorter but less readable.  In the expanded form, I can immediately see which platforms are built.  In the shorter view, I have to know more about travis-ci file.

.travis.yml
1138-4EB: Not really, sorry. It is required for multistage builds (#489). I'll remove it from here.

.travis.yml
1138-4EB: Since all of the builds except macos are executed in the exact same environment from tavis' point of view, I think it makes sense to have them defined as default, and use the `jobs include` to underline differences.

However, the most useful feature is that adding a different build is just one line:

```
  - IMAGE=fedora26+llvm
```

See [.travis.yml#L20-L31](https://github.com/1138-4EB/ghdl/blob/08494d68ecd2c0c2df60c901ec545f6c2a296f72/.travis.yml#L20-L31) in #489.

Since travis does not allow to dinamically change the jobs matrix (that was the original motivation to spin all the runs in a single travis job), being able to add/remove builds (un)commenting a single line makes it easier to have 4-5 platforms for everyday runs, but do complete tests less frequently (say, before a release).

tgingold: OK, but worth adding a comment to explain how that works.

1138-4EB: How about this?

> Defining the same environment variable in the top-level `env` field is interpreted by travis as a 'matrix expansion'. I.e., a separate job is run for each of the given values. Then, `jobs:include` is used to include additional jobs, besides the expanded matrix. See:
- tgingold/ghdl#488
- https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix
- https://docs.travis-ci.com/user/build-stages/#Build-Stages-and-Build-Matrix-Expansion

tgingold: I think you simply need to describe what you are doing; no need to refer to the documentation.
Something like:
The linux builds share the same environment and are factored out; however the mac build is described as an additional job.

1138-4EB: Done. I reordered the content to leave all the common constraints above the matrix (i.e. I moved the deploy section).
Note that linux builds do not exactly 'share the same environment', an independent machine/box is used for each job/build, but all of them are based on the same constraints/environment.

tgingold:
Thanks!

Issue #477, opened 2017-12-07
Continuous Integration, releases and ready-to-use rolling images


eine:
- The first part of this issue was implemented in #489.
- Some of the content was moved to [ghdl/docker](https://github.com/ghdl/docker), sections `Use cases for the actually available images`, `Let's have some fun! [use GHDL live/local/ci demo]` and `Extending base images` are now available at [USE_CASES.md](https://github.com/ghdl/docker/blob/master/USE_CASES.md). `Packing` is available at ghdl/docker#2.
- `Documentation overhaul` and `Project organization` were moved to 1138-4EB/ghdl-io#1. These are being discussed in #506.

---

# Table of contents
- [Maintainability](#maintain)
- [More to come](#tocome)
- [Tasks](#pending)

---

<a name="maintain"></a>
# Maintainability

- Use an encrypted travis access token ([2013-01-28-token-token-token](https://blog.travis-ci.com/2013-01-28-token-token-token), [encryption-keys](https://docs.travis-ci.com/user/encryption-keys/#Notifications-Example), [triggering-builds](https://docs.travis-ci.com/user/triggering-builds)) to cross-trigger builds:
  - Move files corresponding to stage X (let it be 2, 3 or 4) to a separate branch (say, `extended`). Have travis automatically trigger a build of branch `extended` every time a build of branch `master` is successful.
  - Same as above, but host the branch in a different repo.

See 1138-4EB/ghdl-io#1 and #506.

<a name="tocome"></a>
# More to come

I believe that having rolling docker images is a major feature for the project. Yet, it is not good enough for my target audience, which are new/young users with little to no contact with digital design. Forcing them to use the shell (say vim, nano, emacs...) is not the best starting point.

`ghdl/ext:vunit-gtkwave` is the first step to make it easier. The launch script automatically mounts the current path in the container, and the X socket is shared, so that GUI applications can be run (i.e. gtkwave). The user can move and edit files with the tools in the host (file explorer, dolphin, nautilus, gedit, notepad++...). Still, the console is required to run the scripts, but VUnit makes a pretty good job simplifying it.

However, an X server is required in order to run `ghdl/ext:vunit-gtkwave`. This makes it slightly more difficult to use it on a remote host (say play-with-docker or a machine in the lab). Sure, SSH can be used to forward the required ports, but it requires to set up users and keys in the remote machine.

There are some interesting projects that can be integrated to provide a functional web-based IDE. Indeed, there is a very interesting discussion going on in #111.

[...]

Issue #377, opened 2017-06-30
Debian Stretch package 


brungo:
Please add the .deb package for Debian Stretch.

Thanks!

1138-4EB:
Hi @brungo, have you tried to execute any of the available binary packages (not deb)?

https://github.com/tgingold/ghdl/releases/tag/2017-03-01

The one in Ubuntu 12.04 is built with these dependencies and versions:

https://github.com/1138-4EB/ghdl-tools/blob/img-ubuntu1204-llvm/Dockerfile

Then, it should work in your Debian Stretch as long as you install these three packages (for LLVM version): `libgnat-6 g++-6 zlib1g-dev`. Note that these are runtime dependencies and they are debian strectch package names. On top of that, you may have to add `ln -s /usr/bin/gcc-6 /bin/gcc` or set some specific envvar so that ghdl finds a compiler.

Moreover, I am adding stretch and sid to the builds matrix: https://github.com/1138-4EB/ghdl/tree/updatebuildimgs/dist/linux/docker Still, this will only ensure that another binary release can be added. You'd still have to download it and install the runtime dependencies mentioned above.

According to the tracker at debian.org, ghdl is no longer an oficial package: https://tracker.debian.org/pkg/ghdl This means that there is no maintainer in charge of making the deb package. Furthermore, there are non-official ghdl-0.31 and ghdl-0.33 packages: https://sourceforge.net/projects/ghdl-updates/files/Builds/

I assume that it is quite straighforward to adapt those to new binary releases. Indeed, since releases are created in a CI environment, it is 'cheap' to add packaging to the artifact deployment pipeline. However, I have never done it. Are you willing to help with this specific task? I'll do the scripting and integration in the CI. Just need to know what's the difference between those binary releases and a deb package.

Also note that, at now, all the builds are for x86_64 (we do not cross-compile at all). I don't know if Debian accepts packages which are available in a single platform. Did you mean to download it from here, or you'd like to have it available in official repos? Nevertheless, this is not a technical limitation, since the builds in the CI can be extended to support cross-compilation. No one has focused on that, though.

Paebbels:
Debian dropped GHDL support, because GHDL contains non-free content: the IEEE libraries. The IEEE package license is not considered free by Debian. But we know a person helping out in building *.deb packages if a new version is ready for assembly.

[...]

PR #489, opened 2017-12-13
Use docker multistage builds, travis stages, yaml aliases and deploy to dockerhub

1138-4EB:
NOTE: there are two deploy providers. The one corresponding to GitHub releases is untouched, so it will continue working. However, deploying to dockerhub requires encrypting a user and a password with travis cli. [.travis.yml#L44](https://github.com/1138-4EB/ghdl/blob/4a32c357b6376e5ab963f1f1efa206d7f0c42ea8/.travis.yml#L44) is to remember that this PR should not be merged until [.travis.yml#L33-L34](https://github.com/1138-4EB/ghdl/blob/4a32c357b6376e5ab963f1f1efa206d7f0c42ea8/.travis.yml#L33-L34) are updated. This can only be done by maintainers of tgingold/ghdl. See [Credentials and hosting](#credentials) below.

# Table of contents

For a quick reference about the objective of this issue, see #477.

- [Context](#context)
- [Multistage builds of docker images](#multistage)
- [Separate build and run images](#separate)
- [Ready-to-use images](#ready)
- [Update frequency](#update-frequency)
- [Travis stages](#stages)
- [Putting all together](#alltogether)
  - **Travis stage 0**
  - **Travis stage 1**
- [Credentials and hosting](#credentials)
  - GitHub
  - Travis
  - DockerHub
- [Maintainability](#maintain)

---

<a name="context"></a>
# Context

Some months ago I contributed to this repo introducing the usage of 
[Docker](https://www.docker.com/) in [travis-ci](https://travis-ci.org/tgingold/ghdl) (#272, #270). At that time, docker images were defined in a external repository ([1138-4EB/ghdl-tools](https://github.com/1138-4EB/ghdl-tools)). Therefore, the generation/updating and distribution was not in sync with this repo.

This summer @tgingold refactored the build system, introducing the following changes:

- For each docker image a file was created in `dist/linux/docker/build-<distribution>-<compiler>`.
- Some images where dropped and others added, resulting in the following list: `stretch-mcode`, `ubuntu14-mcode`, `ubuntu14-llvm-3.8` and `fedora26-mcode` (4).
- A variant to `stretch-mcode` was added: `stretch-mcode-gpl` (4+1).
- Instead of using a single job in travis-ci with multiple parallel containers, a different job was used for each image.
- An additional job was added for `macosx-mcode` (4+1+1).
- Logs were significantly improved, using travis folds and timing them.

This issue follows the path. On the one hand, the process is optimized. On the other hand, intermediate artifacts are published. Last, release distribution is simplified.

<a name="multistage"></a>
# Multistage builds of docker images

Thanks to [multistage builds](https://docs.docker.com/engine/userguide/eng-image/multistage-build/) introduced in latest docker releases, there is no need to have separate [`build-ubuntu14+mcode`](https://github.com/tgingold/ghdl/blob/master/dist/linux/docker/build-ubuntu14%2Bmcode) and [`build-ubuntu14+llvm-3.8`](https://github.com/tgingold/ghdl/blob/master/dist/linux/docker/build-ubuntu14%2Bllvm-3.8) files. Especially, because the latter must include all the dependencies already included in the first. A single file can be used instead:

``` Dockerfile
# [build] Ubuntu 14

FROM ubuntu:trusty AS mcode

RUN apt-get update -qq \
  && apt-get -y install --no-install-recommends \
            gnat gcc make zlib1g-dev \
  && apt-get autoclean && apt-get clean && apt-get autoremove


FROM mcode AS llvm-3.8

RUN apt-get update -qq \
  && apt-get -y install --no-install-recommends \
            llvm-3.8-dev clang-3.8 \
  && apt-get autoclean && apt-get clean && apt-get autoremove
```

Then, both images are built one after the other:

``` bash
docker build -t ghdl/build:ubuntu-mcode --target mcode - < Dockerfile
docker build -t ghdl/build:ubuntu-llvm-3.8 --target llvm-3.8 - < Dockerfile
```

This ensures that the second build uses the cache from the first, saving the time required to fetch the sources again.

<a name="separate"></a>
# Separate build and run images

Build/dev dependencies are not the same as runtime dependencies. Right now a single image is built for each platform and compiler, and it is used to build and test. However, I think it would be "cleaner" to use separate images for the build and test procedures. Following with the example above, these would be the images with runtime dependencies only:

``` Dockerfile
# [run] Ubuntu 14

FROM ubuntu:trusty AS mcode

RUN apt-get update -qq \
  && apt-get -y install --no-install-recommends \
            libgnat-4.6 gcc libc6-dev \
  && apt-get autoclean && apt-get clean && apt-get autoremove


FROM mcode AS llvm-3.8

RUN apt-get update -qq \
  && apt-get -y install --no-install-recommends \
            libllvm-3.8 zlib1g-dev \
  && apt-get autoclean && apt-get clean && apt-get autoremove
```

``` bash
docker build -t ghdl/run:ubuntu-mcode --target mcode - < Dockerfile
docker build -t ghdl/run:ubuntu-llvm-3.8 --target llvm-3.8 - < Dockerfile
```

<a name="ready"></a>
# Ready-to-use images

- [`ghdl/build`](https://hub.docker.com/r/ghdl/build/tags) is used to build GHDL and generate a tarball.
- The tarball is added to [`ghdl/run`](https://hub.docker.com/r/ghdl/run/tags) in order to generate [`ghdl/ghdl`](https://hub.docker.com/r/ghdl/ghdl/tags).
- Test are executed in a container based on the latter.

As a result, if the execution is successful, [`ghdl/ghdl`](https://hub.docker.com/r/ghdl/ghdl/tags) images are verified ready-to-use images, i.e., they include GHDL with minimum runtime dependencies which could successfully complete the testsuites.

<a name="update-frequency"></a>
# Update frequency

[`ghdl/build`](https://hub.docker.com/r/ghdl/build/tags) and [`ghdl/run`](https://hub.docker.com/r/ghdl/run/tags) images are expected to change slowlier than GHDL itself. Therefore, it is suboptimal to build them every time a commit is pushed, which is the current behaviour. This PR parses the commit message to look for `[ci images]`, in order to trigger updating these. If the key is not found, build and test procedures are executed with existing images.

Nevertheless, [`ghdl/ghdl`](https://hub.docker.com/r/ghdl/ghdl/tags) images include the unpacked tarball. Therefore, it makes sense to keep/distribute these after every successful build and test.

<a name="stages"></a>
# Travis stages

Not to be mixed with docker multi-stage builds, [this](https://docs.travis-ci.com/user/build-stages/) is a beta feature that allows to make sequential groups of jobs running in parallel. The default matrix is expanded to a stage named `test`. Then, any other can be added to be executed either before or after 'test' jobs. 

<a name="alltogether"></a>
# Putting all together

- Multistage Dockerfiles for [`ghdl/build`](https://hub.docker.com/r/ghdl/build/tags) images are saved in [`dist/linux/docker/build/<platform>`](https://github.com/1138-4EB/ghdl/tree/builders/dist/linux/docker/build).
- Multistage Dockerfiles for [`ghdl/run`](https://hub.docker.com/r/ghdl/run/tags) images are saved in [`dist/linux/docker/run/<platform>`](https://github.com/1138-4EB/ghdl/tree/builders/dist/linux/docker/run).

NOTE: for tha sake of simplicity, let's assume that there is a single file in each directory, both named `ubuntu14`, and each one with the content shown above.

## Travis stage 0, named `Create build|run images`
If the commit message contains `[ci images]`, sequentially do (else, skip this stage):

- For each file in `dist/linux/docker/**/*`
  - For each `FROM.*AS .*"` line
    - Build an image.
-  That is, build:
   - [`ghdl/build`](https://hub.docker.com/r/ghdl/build/tags)
    - [`ghdl/build`](https://hub.docker.com/r/ghdl/build/tags)
    - [`ghdl/run`](https://hub.docker.com/r/ghdl/run/tags)
    - [`ghdl/run`](https://hub.docker.com/r/ghdl/run/tags)
- If successful, execute deploy: push all of them to DockerHub registry (see [Credentials and hosting](#credentials)).

## Travis stage 1, named `test` (default)

- Do matrix expansion and add macosx. Execute each in a separate job.

Note that this is equivalent to the current setup. However, [`buildtest.sh`](https://github.com/tgingold/ghdl/blob/master/dist/linux/buildtest.sh) is split to [`build.sh`](https://github.com/1138-4EB/ghdl/blob/builders/dist/linux/build.sh) and [`test.sh`](https://github.com/1138-4EB/ghdl/blob/builders/dist/linux/test.sh):

- Execute [`build.sh`](https://github.com/1138-4EB/ghdl/blob/builders/dist/linux/build.sh) inside [`ghdl/build`](https://hub.docker.com/r/ghdl/build/tags) and generate tgz.
- If build failed, exit with error.
- Build [`ghdl/ghdl`](https://hub.docker.com/r/ghdl/ghdl/tags) from [`ghdl/run`](https://hub.docker.com/r/ghdl/run/tags) by adding the tgz.
- Execute [`test.sh`](https://github.com/1138-4EB/ghdl/blob/builders/dist/linux/test.sh) inside [`ghdl/ghdl`](https://hub.docker.com/r/ghdl/ghdl/tags).
- If test failed, exit with error.
- If successful, execute deploy:
  - Push [`ghdl/ghdl`](https://hub.docker.com/r/ghdl/ghdl/tags) to DockerHub registry.
  - If tagged commit, release tgz to GitHub.

The procedure above is true for all te images, except macosx, which does not use docker containers and, therefore, there is no image to push/dpeloy.

<a name="credentials"></a>
# Credentials and hosting

There are three services involved in this build, test and deployment process:

- GitHub: tgingold/ghdl repository. @tgingold is currently the owner. No change required.
- Travis: it is already linked to the github repo and to @tgingold's account. No change required.
- DockerHub: I own the organization named [ghdl](https://hub.docker.com/r/ghdl/). A Docker ID is required which has write access to it. This user and the password must be encrypted for tgingold/ghdl. This is done with the travis cli (a script is provided in this PR to automate it: `travis-enc-dockerhub.sh`), but write access to the repo is required to configure it.

I created a dummy user related to an account at mail.com, which is the one I use in my fork. @tgingold, I don't mind to transfer the email account to you, because I won't need it if/when the registry is updated from your travis.

Note that these credentials are required in order to push docker images to the dockerhub/registry. No write access is given to the github repo at all. The requirement to encrypt them with a clone of the repo is for travis to make sure that the keys are only decrypted in tests related to this repo. Indeed, PRs cannot trigger pushes to the registry. Only maintainers can do it.

<a name="maintain"></a>
# Maintainability

docker user and password need to be encrypted once only for each repo. Then all the image pulling/pushing is handled automatically.

Once in a while, `dist/linux/docker/**/*` files should be checked, to remove old platforms and add new. Images might be removed from dockerhub too.

---

The test matrix has been expanded to debian buster, ubuntu 16 and ubuntu 18 (see [.travis.yml#L20-L31](https://github.com/1138-4EB/ghdl/blob/4a32c357b6376e5ab963f1f1efa206d7f0c42ea8/.travis.yml#L20-L31)). But all of them are commented, in order to obtain the same result as the current implementation.


.gitignore
```diff
@@ -12,9 +12,9 @@ b~*.ad?
 testsuite/get_entities
 
 # Generated files on windows.
-build/
-dist/mcode/build/
-dist/mcode/lib/
+./build/
+./dist/mcode/build/
+./dist/mcode/lib/
```
tgingold: General note: you should split large patches like this in several smaller one (or multiple commit).  That makes reviewing much easier.

This change to .gitignore is ok, but somewhat unrelated to the other changes.

.travis.yml
```diff
@@ -10,35 +10,72 @@ language: minimal
...
-  - provider: releases
+  - &deploy-docker
+    provider: script
+    skip_cleanup: true
+    script: ./dist/linux/deploy2dockerhub.sh "$IMAGE" "$TRAVIS_COMMIT_MESSAGE"
+    on:
+      repo: 1138-4EB/ghdl
+      branch: master
+  - &deploy-releases
+    provider: releases
```
tgingold: Do we really want to also deploy to docker ?  I am not sure about that.

dist/linux/build.sh
```diff
@@ -0,0 +1,125 @@
...
+    llvm)
+	      config_opts="--with-llvm-config"
+        DEP_CLANG="clang"
+    ;;
+    llvm-*)
+	      arg_llvm=`find /usr/bin -name llvm-config* | grep -oP "llvm-config.*"`
+	      arg_cxx=clang++`echo $arg_llvm | grep -oP "llvm-config\K.*"`
+	      config_opts="--with-llvm-config=$arg_llvm CXX=$arg_cxx"
+        DEP_CLANG="`$(basename $(find /usr/bin -name clang++*)) --version | grep 'clang'`"
+    ;;
```
tgingold: Honestly I am not sure that llvm-* is a good idea: it makes the code unreadable.  We cannot anymore tell to a user: 'look at the script to see the options we used to configure'.

dist/linux/create.sh
```diff
@@ -0,0 +1,24 @@
...
+set -e
+
+. dist/linux/travis-utils.sh
+. dist/ansi_color.sh
+#disable_color
+
+scriptdir=$(dirname $0)
+
+for d in `echo build run`; do
+    currentdir="${scriptdir}/docker/$d"
```
tgingold: why not:
for d in build run; do

dist/linux/deploy2dockerhub.sh
```diff
@@ -0,0 +1,31 @@
+#! /bin/sh
+
+set -e
+
+. dist/linux/travis-utils.sh
+. dist/ansi_color.sh
+#disable_color
+
+if [ "$1" = "skip" ] || [ "${1}$(echo $2 | grep -o '\[ci images\]')" = "" ]; then
```
tgingold: Worth a comment.

dist/linux/travis-utils.sh
```diff
@@ -1,7 +1,7 @@
 # This is a trimmed down copy of
 # https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/templates/header.sh
 travis_time_start() {
-  travis_timer_id=$RANDOM
+  travis_timer_id=`date +%N`
```
tgingold: Unrelated change, and I fear it is wrong: the timer_id won't be uniq.

.gitignore
1138-4EB: I tried to do it. Indeed, this PR includes minimum changes to use multistage dockerfiles. I kept every unrelated enhancement in #477 as WIP.

This change precisely is motivated because some of the files in `dist/docker/build` or `dist/docker/run` were filtered by `.gitignore` if `./` is not added. I know, it makes little sense to me too.

.travis.yml
1138-4EB: Travis stages are absolutely independent. That means that the images we build in stage 0 (create ghdl/build and ghdl/run images) are not available in the jobs in stage 1 (the matrix expansion and macos). Therefore ghdl/build and ghdl/run images must be saved somewhere. We need some intermediate storage service. This can be S3, dockerhub or any other.

It also acts as an external cache, allowing us to ignore stage 0 for everyday work, and execute it only when platforms change.

However, we can avoid pushing ghdl/ghdl images, which are the ready-to-use images which passed the test successfully. Yet, I think these are the most valuable for end-users. Since we 'need' to set up credentials because of other images in stage 0, I think it is worth reusing them and push these images too.

dist/linux/build.sh
1138-4EB: The main point is that we accept any version of LLVM after 3.5 (i.e., 3.5, 3.8, 3.0, 4.0 and 5.0), not just some of them. This is stated in the docs. Having 6 different cases might make the user think that there is some difference between them. Indeed, it requires users with newer versions to edit the file. Conversely, having all the variants with only name differences handled in the same case makes it explicit that we make exactly the same steps, independently of the names of the libraries.

I am ok with reverting these changes, and adding a new case for 5.0 (which is what I did at first for ubuntu18). However, I'd like you to consider if it is worth just adding a comment here:

```
llvm-*)
    # If the version is appended, we just filter it to produce, e.g. for version 3.5:
    # config_opts="--with-llvm-config=llvm-config-3.5 CXX=clang++" ;;

    arg_llvm=`find /usr/bin -name llvm-config* | grep -oP "llvm-config.*"`
```

dist/linux/create.sh
1138-4EB: Sure.

dist/linux/deploy2dockerhub.sh
1138-4EB: Sure.

dist/linux/travis-utils.sh
1138-4EB: $RANDOM exists in travis, but not inside the containers. Since I use travis-utils in build.sh, test.sh and create.sh, which are executed inside the containers, I need a replacement. Note that it is the time in nanoseconds, which is quite random, although not strictly unique. Furthermore, there not more than 10 folds to track in the same run. It'd be very bad luck if they where repeated. Last, this is only used for timing. I don't think it is a critical issue, because the run can be restarted in the worst case.

tgingold: Ok, that makes sense.  I suppose the default shell in the containers is not bash.
Worth a comment too.

dist/linux/build.sh
tgingold: So, let's create a special case for llvm-3.5, and for later version extract the version from the name.  Find won't work if you have multiple llvm installation.

.travis.yml
tgingold: I am a little bit worried about all the network traffic!

I think docker is nice to handle multiple distributions (debian, ubuntu...).
However, I am not convinced about the need of docker images for users.  ghdl is very simple to install, and need other components (like editor).

1138-4EB: Oh, well, I have not measured it, but I think it is somehow equivalent to the requirements we have now. This is because docker images are [composed of layers](https://docs.docker.com/engine/userguide/storagedriver/imagesandcontainers/#images-and-layers). Therefore, if multiple images have common layers, these need to be downloaded once only.

At now, we build every image for every job in each run. Building means first downloading the base image, then downloading all the dependencies and installing them. E.g., for `stretch-mcode`, the base image has a single 22.49MB layer (e7bb5...) and we then download 40.7MB of archives that we need to install: https://travis-ci.org/tgingold/ghdl/jobs/316767025 Total: 63.19MB + processing time.

With the new approach, we will be downloading the base image and the dependencies all together, already preinstalled in a layer (ghdl/build). Then, we will get a second image (ghdl/run), which is based on it, so most of the content is actually shared and must not be transferred. Last, we build a third image (ghdl/ghdl) based on the latter. This last image also shares all the content with the previous one, except the GHDL tarball. Therefore, when we push the last image to the registry only the last layer needs to be pushed.

Let me show you a practical example with `stretch-mcode`:

ghdl_layers

You can see that:

- We need to download two layers, e7bb5... (22.49MB) + 8748e... (70.23MB) to get ghdl/build.
- In order to get ghdl/run, we only need to download layer 3b589... (49.22MB), because layer e7bb5... is already local.
- Last, we only need to download layer f2624... (3.1MB) to get ghdl/ghdl, because the previous two layers, e7bb5... and 3b589... are already local.

In the end, we transferred 22.49 + 70.23 + 49.22 + 3.1 = 145.04MB. In practice, in a normal run we would have downloaded 141.94MB and then uploaded 3.1MB.

Furthermore, I optimized this by using travis' cache system in order to avoid downloading images every time. Long story short, it is stupid because it seems that travis caches are stored exactly in the same place as dockerhub (S3): https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-169856919 . Well, it would make sense if building our images took (tens of) minutes. But we have it done in a minute at most. Anyway, it is straighforward to add it if you want: https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-248915326 , [faster-travis-ci-tests-with-docker-cache](http://atodorov.org/blog/2017/08/07/faster-travis-ci-tests-with-docker-cache/)

---

Yet, we can stil try to optimize it a little: [ghdl/run:stretch-mcode](https://github.com/1138-4EB/ghdl/blob/stages/dist/linux/docker/run/stretch) and [ghdl/build:stretch-mcode](https://github.com/1138-4EB/ghdl/blob/stages/dist/linux/docker/build/stretch), both install gcc in separate layers (3b589... and 8748e..., respectively). We could optimize it installing gcc in its own layer, not with the rest of the dependencies. However, I prefer not to introduce that in this PR, in order to make the transition easier to follow/understand.

---

> However, I am not convinced about the need of docker images for users. ghdl is very simple to install, and need other components (like editor).

Well, we definitely do not need to push ghdl/ghdl images to the registry. But:

- ghdl/run images are already in the registry and we have already built ghdl/ghdl images in order to run tests in them. Thus, pushing ghdl/ghdl images is uploading 3-10MB, the 'nightly' GHDL tarball.
- Any user can have ghdl/ghdl variants built just doing [travis-ci.sh#L10-L16](https://github.com/1138-4EB/ghdl/blob/stages/dist/linux/travis-ci.sh#L10-L16), i.e., using the latest ghdl/run and the lastest tarball available on GitHub releases. We are just avoiding multiple users creating images with the same content but different layer hashes (because they are built independently).
- Use cases in #477.
- There is no conflict at all to use other components (like editor). You can use your usual toolset in the host, by just sharing the folder of the repo with the container. Indeed, you can make the ghdl/ghdl image be an alias of GHDL, so that it is used exactly the same way as if it was installed natively. Of course, that's only one approach. From [#377](https://github.com/tgingold/ghdl/issues/377#issuecomment-352104230):

> Indeed, part of the 'complexity' of docker is due to it's versatility:
> - Daemon mode, how you use it: `docker run -d <image> bash`
> - Batch mode, the example I provided as an alternative to running buildtest.sh locally: `docker run -t -v /$(pwd):/src -w //src <image> bash -c "./dist/linux/buildtest.sh..."`
> - Interactive mode, the demo shown in #477: `docker run -it <image> bash`
> - Any of the previous with GUI support, the second demo shown in #477: `./dist/docker_guiapp.sh -it <image> bash`

- Building GHDL is not difficult, but it is neither straightfordward. It requires several dependencies that a user might not be allowed to install, and we are requiring the user to actually know how programs are compiled and installed (and how to clean temporal files). Making ghdl/ghdl images available makes it equally easy for the user to get mcode, llvm and/or gcc variants. Furthermore, I can have code coverage on windows! Because the stretch-based image with gcc backend will work seamlessly.

- Providing ready-to-use images sets the path for ghdl/ext images, commented in [#477](https://github.com/tgingold/ghdl/issues/477#extend). Having GHDL built with latest vunit and gtkwave versions let's us detect conflicts fast. It also allows us to redirect users to a known conflict-free environment when issues as *VCD works but GHW fails* arise. Note that in this case, a single image is built for each of the tools, all of them based on `ghdl/ghdl:stretch-mcode`, which is my 'official' distribution, as commented below:

Yet, I agree that it might be overkill and not useful to push absolutely all the ghdl/ghdl images to the registry. I think we should pick an "official" distribution, and provide only the variants of it. I'd vote for stretch or buster, because debian slim images are the [smallest](https://hub.docker.com/r/ghdl/ghdl/tags/) (until I can make GHDL run on alpine). Therefore there would be only 3 images available: `ghdl/ghdl:mcode`, `ghdl/ghdl:llvm` and `ghdl/ghdl:gcc`. Optionally, we can add 3 GPL variants: `ghdl/ghdl:mcode-gpl`, `ghdl/ghdl:llvm-gpl` and `ghdl/ghdl:gcc-gpl`.

dist/linux/build.sh
1138-4EB: Then, lets just split this to a separate PR and have this solved in the 'verbose' way here.

.travis.yml
tgingold: Ok, I was not aware of the optimizing docker image transfer.
Then it would make even more sense not to cache them in travis.

dist/linux/build.sh
tgingold: Ok.

tgingold:
THanks!

1138-4EB:
No @tgingold! It won't work! You need to set up the docker user and pass for this repo. See first comment above.

I must leave now. I'll be back in 1h.

tgingold:
> On 16 Dec 2017, at 20:31, 1138-4EB <notifications@github.com> wrote:
> 
> No @tgingold! It won't work! You need to set up the docker user and pass for this repo. See first comment above.
> 
> I must leave now. I'll be back in 1h.

We will do it later.

1138-4EB:
Back now. Available at gitter.

1138-4EB:
Well, I didn't remember that I filtered the repo, so there is no failure. However, nothing is pushed to dockerhub, it is skipped. This means that, until the encryption is done in this repo, we rely on keeping a branch in my fork up to date with the master branch here.

PR #495, opened 2017-12-17
Freebsd configure

endofexclusive:
User convenience when building GHDL on FreeBSD.

With these patches, GHDL/LLVM can be built on FreeBSD 11.1 with:
```
$ mkdir build && cd build
$ ../configure --prefix=/opt/ghdl-0.35-llvm --with-llvm-config=/usr/local/llvm40/bin/llvm-config --with-gnatmake=/usr/local/gcc6-aux/bin/gnatmake
$ gmake
$ gmake install
```

### DIFF
--- a/configure
+++ b/configure
@@ -8,6 +8,7 @@ CC=${CC:-gcc}
 CXX=${CXX:-clang++}
 CFLAGS=${CFLAGS:--g}
 GNATMAKE=${GNATMAKE:-gnatmake}
+MAKE=${MAKE:-make}
 LDFLAGS=
 prefix=/usr/local
 libdirsuffix=lib/ghdl
@@ -299,7 +300,7 @@ if ! sh ./config.status; then
 fi
 
 # Create dirs
-make create-dirs
+$MAKE create-dirs
 
 # Generate ortho_code-x86-flags
 if test $backend = mcode; then
--- a/Makefile.in
+++ b/Makefile.in
@@ -187,10 +187,10 @@ gcc_vhdl_dir=$(gcc_src_dir)/gcc/vhdl
 
 all.gcc:
 	@echo "For installing sources in gcc, do:"
-	@echo "1) make copy-sources"
+	@echo "1) $(MAKE) copy-sources"
 	@echo "2) configure, build and install gcc"
-	@echo "3) make ghdllib"
-	@echo "4) make install"
+	@echo "3) $(MAKE) ghdllib"
+	@echo "4) $(MAKE) install"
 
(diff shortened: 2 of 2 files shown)


configure
```diff
@@ -97,6 +98,7 @@ Options [defaults in brackets]:
   --srcdir=SRCDIR             source code path [$srcdir]
   --with-gcc=DIR              use gcc backend from DIR
   --with-llvm-config=PATH     use llvm from PATH (needs llvm 3.5 - 3.9)
+  --with-gnatmake=PATH        use gnatmake from PATH
```
tgingold: You can already override gnatmake with: xx/configure ... GNATMAKE=my-gnatmake
So, no need for this new option. (But doc improvement is fine!)

Contrary to llvm or gcc, gnatmake is only used to build ghdl

configure
```diff
@@ -298,8 +300,14 @@ if ! sh ./config.status; then
...
 
+if type gmake > /dev/null 2>&1 ; then
+    gmake=gmake
+else
+    gmake=make
+fi
+
 # Create dirs
-make create-dirs
+${gmake} create-dirs
```
tgingold: Let's do the same for make: allowing to use MAKE=gmake

configure
endofexclusive: Thank you for the comment. The idea with the -- option was was to have a similar user interface for specifying external tools. The option has now been removed from the pull request.

configure
endofexclusive: Yes that is indeed a more general solution.

endofexclusive:
The pull request has been updated. I changed history on the pull request branch since the current solution does not have much in common with the first proposal.

GHDL/LLVM can now be built in FreeBSD 11.1 with the following commands:
```
GNATMAKE=`find /usr/local/ -name gnatmake 2> /dev/null`
LLVMCONFIG=`find /usr/local/ -name llvm-config 2> /dev/null`
MAKE=gmake GNATMAKE=$GNATMAKE ../configure --prefix=/opt/ghdl-dev-llvm --with-llvm-config=$LLVMCONFIG
gmake
gmake install
```
I am looking into adding install/build instructions for FreeBSD to the GHDL documentation. Or possibly just the README.

Paebbels:
I would prefer detailed build instructions in the documentation, but a short instruction with a link to the official documentation in the README.

1138-4EB:
Might `find` fail with multiple versions of llvm installed, as commented [here](https://github.com/tgingold/ghdl/pull/489#discussion_r157337320)?

tgingold:
Thanks!

PR #527, opened 2018-02-11
Fixed library compilation on Windows

Paebbels:
Fixed library compilation on Windows due to changes in 7ef79a53.

### DIFF
--- a/dist/windows/compile-ghdl.ps1
+++ b/dist/windows/compile-ghdl.ps1
@@ -90,11 +90,8 @@ $Script_WorkingDir =	Get-Location
 $GHDLRootDir =				Convert-Path (Resolve-Path ($PSScriptRoot + "\" + $RelPathToRoot))
 
 # set default values
-$EnableVerbose =			$PSCmdlet.MyInvocation.BoundParameters["Verbose"]
-$EnableDebug =				$PSCmdlet.MyInvocation.BoundParameters["Debug"]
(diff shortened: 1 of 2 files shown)


Paebbels:
Oh I remember ...
AppVeyor uses the makefiles in MinGW to build GHDL for Windows. Te PowerShell flow is not used on AppVeyor due to problems in executing the Ada Core compiler on Windows Server 2016.

tgingold:
Thanks.

Do you think it is possible to execute the msys2 Ada compiler from PowerShell ?

Paebbels:
I could run the Ada compiler from PowerShell in MinGW32 and/or MinGW64. But the results of the compilation are not running standalone.

We had a discussion some time ago, that we would need to ship the resulting binaries and 5-7 DLLs with ghdl.exe to have a standalone system.

As I now understand how ArchLinux / pacman packages work, I think we can package GHDL for the MinGW pacman system to get it installed diretly in MinGW.

tgingold:
> On 13 Feb 2018, at 21:18, Patrick Lehmann <notifications@github.com> wrote:
> 
> I could run the Ada compiler from PowerShell in MinGW32 and/or MinGW64. But the results of the compilation are not running standalone.
> 
> We had a discussion some time ago, that we would need to ship the resulting binaries and 5-7 DLLs with ghdl.exe to have a standalone system.

Did you try to link using -static so that DLL aren't used ?

PR #555, opened 2018-04-14
fix version.tmp in travis-ci

### LINKED ISSUE #554: Version detection broken in travis-ci
Almost a month ago the version detection was broken in travis-ci. The last successful and correct build: [834](https://travis-ci.org/ghdl/ghdl/jobs/354693488) (see line 1518) and the first successful but incorrect one is [836](https://travis-ci.org/ghdl/ghdl/jobs/356988756) (see line 1521). I think that these are the changes to blame: https://github.com/ghdl/ghdl/commit/ffca448651c5b7b0fc5fe6bdbbcf1e0dc13419e1#diff-7e056ddc530791272d363e50c1f25647

Because the version is now set in configure, I think that we need to analyse [travis-ci.sh#L103-L104](https://github.com/ghdl/ghdl/blob/master/dist/linux/travis-ci.sh#L103-L104). The output we get now is:

```
/# git clone https://github.com/ghdl/ghdl
...

/# cd ghdl

/ghdl# cat version.tmp
cat: version.tmp: No such file or directory

/ghdl# cat src/version.in
package Version is
   Ghdl_Ver : constant String := "@VER@";
   Ghdl_Release : constant String :=
      "(tarball) [Dunoon edition]";
end Version;

/ghdl# make -f Makefile.in srcdir=. version.tmp
if ! test -d ./.git \
   || ! desc=`cd .; git describe --dirty`; then \
  desc="tarball"; \
        fi; \
sed -e "s/[(].*[)]/($desc)/" \
               -e "s/@VER@/@ghdl_version@/" < src/version.in > version.tmp; \

/ghdl# cat version.tmp
package Version is
   Ghdl_Ver : constant String := "@ghdl_version@";
   Ghdl_Release : constant String :=
      "(v0.35-150-g60f3bb31) [Dunoon edition]";
end Version;
```

Note that the sequence above is executed just once. Then, inside each container, `configure` is executed and last `make`. As a result, the second time `make` is executed, `Ghdl_Ver` in `src/version.in` contains `@ghdl_version@` instead of `@VER@`, and the replacement with sed does not work.

A possible solution is to remove lines [L98-L104](https://github.com/ghdl/ghdl/blob/master/dist/linux/travis-ci.sh#L103-L104) from `travis-ci.sh`. As a result, we will get `GHDL 0.36-dev (tarball) [Dunoon edition]`. However, I think this is undesired, because the nightly builds and/or docker images will produce binaries that are aparently not different from releases. I believe that `tarball` should only be used for releases, i.e., non `-dev` builds.

1138-4EB:
Close #554.

@tgingold , the [fix](https://github.com/ghdl/ghdl/commit/ba14e9da2a0ca776e6be758f5b0fa49f6090e838) you pushed produces an acceptable solution (`GHDL 0.36-dev (tarball) [Dunoon edition]`, [840](https://travis-ci.org/ghdl/ghdl/jobs/366221639)), but not the one we got earlier ([834](https://travis-ci.org/ghdl/ghdl/jobs/354693488)). Because travis builds are exported as docker images, it is desirable to have the `describe --dirty` string/tag.

This PR makes the `"s/[(].*[)]/($$desc)/"` to be conditional: `(.*)` is only replaced if `$desc` is not "empty" (word `tarball`, which is the default in `src/version.in`, is considered the keyword for empty). So:

- When lines 103-104 of the travis-ci.sh are executed, field `Ghdl_Release` in `version.in` is changed, but `Ghdl_Ver` is left untouched.
- Then, inside each container/build, field `Ghdl_Ver` is changed, but `Ghdl_Release` is left untouched.

Besides, if `configure` and `make` are executed locally (with `git` available), both fields, `Ghdl_Ver` and `Ghdl_Release` will be changed at the same time. Shall git not be available, the output will be `GHDL 0.36-dev (tarball) [Dunoon edition]`.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -147,11 +147,14 @@ include $(srcdir)/src/grt/Makefile.inc
 version.tmp: $(srcdir)/src/version.in force
 #	Create version.tmp from version.in, using git date/hash
 	if ! test -d $(srcdir)/.git \
-	   || ! desc=`cd $(srcdir); git describe --dirty`; then \
-	  desc="tarball"; \
-        fi; \
-	$(SED) -e "s/[(].*[)]/($$desc)/" \
-               -e "s/@VER@/$(ghdl_version)/" < $< > $@; \
+	  || ! desc=`cd $(srcdir); git describe --dirty`; then \
+	    desc="tarball"; \
+	fi; \
+	cp $< $@; \
+	if [ "$$desc" != "tarball" ]; then \
+	  sed -i.bak -e "s/[(].*[)]/($$desc)/" $@;  \
+	fi;  \
+	sed -i.bak -e "s/@VER@/@ghdl_version@/" $@; \
 
 version.ads: version.tmp
 #	Change version.ads only if version.tmp has been modified to avoid
--- a/dist/linux/travis-ci.sh
+++ b/dist/linux/travis-ci.sh
@@ -100,9 +100,8 @@ else
     # This is a little bit hack-ish, as it assumes that 'git' is not
     # available in docker (otherwise it will describe as -dirty
     # because this modifies the source file version.in).
-    ghdl_version_line=`grep -e '^ghdl_version' configure`
-    make -f Makefile.in srcdir=. $ghdl_version_line version.tmp
-    cp version.tmp src/version.in
+    make -f Makefile.in srcdir=. version.tmp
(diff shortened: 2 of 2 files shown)


tgingold:
I think I understand the issue.  I will try something a little bit simpler.

tgingold:
Thanks for the early check (and for raising the issue).

PR #557, opened 2018-04-17
move COPYING_DOC.md to subdir doc

1138-4EB:
After talking to GitHub support, it seems that the license is not properly detected now because there are two license files in the root of the repo. This PR moves COPYING_DOC.md to subdir doc, and adapts the links in README.md accordingly.

Note that the change will not be seen inmediately after the merge/rebase. We need to wait until the internal algorithm in GitHub is executed (which is done periodicaly).

Demo repo: https://github.com/1138-4EB/lictest_sub


tgingold:
I'm away this week, I'll look when I come back. Thanks

PR #570, opened 2018-05-19
configure: Fall back to `pwd` if `pwd -W` does not work

ccorn:
This enables configure runs under Cygwin.

The patch also (re)moves an unused `` curdir=`pwd` ``.


tgingold:
Thanks!

PR #573, opened 2018-05-21
Cygwin executable adaptations

ccorn:
Running the testsuite under Cygwin got stuck at `testsuite/gna/issue450`.

First failure was a *File not found* when linking with `-lghdlvpi`. The first commit addresses that by making sure that `SOEXT=.dll` (and other Windows-like settings) under Cygwin.

The error then changed to *Permission denied*. That is addressed by the second commit which switches to `INSTALL_PROGRAM` for anything with `SOEXT`. This makes sure that DLLs get executable permissions. Note that this change affects all platforms: Looking into `/usr/lib` under Linux (`SOEXT=.so`) as well as under MacOS X (`SOEXT=.dylib`) I see that shared libraries should get executable permissions.

Running `testsuite/gna/issue450/testsuite.sh` can now load the module `vip2.vpi`.
(Thereafter it segfaults, and a `gdb` backtrace tells me that that happens in `vpi_register_cb`. Not sure whether I can track that down, in any case that would be a separate issue.)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -375,7 +375,7 @@ install.libghdlsynth.static: libghdlsynth.a install.libghdlsynth.include
 	$(INSTALL_DATA) -p ghdlsynth.link $(DESTDIR)$(libdir)/
 
 install.libghdlsynth.shared: libghdlsynth$(SOEXT) install.libghdlsynth.include
-	$(INSTALL_DATA) -p libghdlsynth$(SOEXT) $(DESTDIR)$(libdir)/
+	$(INSTALL_PROGRAM) -p libghdlsynth$(SOEXT) $(DESTDIR)$(libdir)/
 
 ################ ghwdump #################################################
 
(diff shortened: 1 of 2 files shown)


ccorn:
`configure` has [yet another mention of `mingw`](https://github.com/ghdl/ghdl/blob/master/configure#L330), so I got the idea that it should have a branch for `*cygwin*` there as well.

As it is now (with and without this pull request), `configure` refuses to use the mcode backend under Cygwin.

Note: I use the 64-bit version of Cygwin and the LLVM backend.

On top of this pull request, I have tried a tentative patch to enable mcode under Cygwin:
``` diff
--- a/configure
+++ b/configure
@@ -328,7 +328,7 @@ $MAKE create-dirs
 if test $backend = mcode; then
     case "$build" in
        *darwin*) ortho_flags="Flags_Macosx${mcode64}" ;;
-       *mingw32*) ortho_flags="Flags_Windows${mcode64}" ;;
+       *mingw32* | *cygwin*) ortho_flags="Flags_Windows${mcode64}" ;;
        *linux*) ortho_flags="Flags_Linux${mcode64}" ;;
        *) echo "Unsupported $build build for mcode"; exit 1;;
     esac
```
With that, `./configure` no longer complains about mcode usage, but the build fails:
```
gcc -c -g -o memsegs_c.o src/ortho/mcode/memsegs_c.c
src/ortho/mcode/memsegs_c.c: In function ‘mmap_realloc’:
src/ortho/mcode/memsegs_c.c:58:9: warning: implicit declaration of function ‘mremap’ [-Wimplicit-function-declaration]
   res = mremap (ptr, old_size, size, MREMAP_MAYMOVE);
         ^~~~~~
src/ortho/mcode/memsegs_c.c:58:38: error: ‘MREMAP_MAYMOVE’ undeclared (first use in this function)
   res = mremap (ptr, old_size, size, MREMAP_MAYMOVE);
                                      ^~~~~~~~~~~~~~
src/ortho/mcode/memsegs_c.c:58:38: note: each undeclared identifier is reported only once for each function it appears in
make: *** [Makefile:174: memsegs_c.o] Error 1
```
Until that is fixed, the above case statement better remains unpatched, which is in accordance with this pull request.

ccorn:
Notice: With the committed patches, `sanity` and `vests` test runs are successful; and the only `gna` tests that fail are `issue98` and `issue450`, both successfully loading the VPI and calling its `my_handle_register`, which calls `vpi_register_cb`, which segfaults. The compiler emits a warning when compiling the respective `vpi*.c`, indicating that the function pointer assigned to `cb.cb_rtn` does not match the expected prototype. Fixing that (extending the prototype, ignoring the argument, and returning 0) does not help however, it just makes the warning go away, the segfault still happens.
This is certainly a separate issue; but if you have suggestions, I might be digging further.

tgingold:
Currently win64 is not supported by mcode, so mingw-x64 won't be too.
The amount of work is not very high, but it's very technical.

PR #575, opened 2018-05-22
Update Makefile.in

flip111:
make fails if mkdir command fails. `-` in front of mkdir treats it as a success. This is useful when reusing existing `bin`-dir & friends. Or when building/installing ghdl a second time

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -441,11 +441,11 @@ endif
 
 install.dirs:
 #	Use -p to create parents and to avoid error if existing.
-	$(MKDIR) -p $(DESTDIR)$(prefix)
-	$(MKDIR) -p $(DESTDIR)$(bindir)
-	$(MKDIR) -p $(DESTDIR)$(libdir)
-	$(MKDIR) -p $(DESTDIR)$(prefix)/$(libdirsuffix)
-	$(MKDIR) -p $(DESTDIR)$(incdir)
+	-$(MKDIR) -p $(DESTDIR)$(prefix)
+	-$(MKDIR) -p $(DESTDIR)$(bindir)
+	-$(MKDIR) -p $(DESTDIR)$(libdir)
+	-$(MKDIR) -p $(DESTDIR)$(prefix)/$(libdirsuffix)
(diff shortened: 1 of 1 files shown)


tgingold:
mkdir -p shouldn't fail but adding - doesn't harm.

flip111:
> But, mkdir with -p should not fail if e.g. bin already exists, right?
> 
> Can you submit the original error message that you are seeing in your use case?

You're right, it shouldn't fail. I don't know what was going on here. But i can tell you that before i was getting a message like `mkdir: cannot create directory ‘/home/flip111/.local/bin’: File exists` and after my patch it worked ok.

Since it's already merged i won't investigate what exactly is going on here.

Paebbels:
It's saying `file exists`, not `directory exists`. `-p` can only handle the latter. Linux does not support having a directory and a file of the same name in the same parent directory. So maybe you had a file called `bin`?

In such a case it needs to fail.

flip111:
> It's saying file exists, not directory exists. -p can only handle the latter.

Try on your own system without `-p` flag what messages you get when you try to create a directory that already exist.

> So maybe you had a file called bin?

I had before, but i thought i deleted this file before i retried.

flip111:
I think it's better to reverse this commit for now, until the situation happens again and can be better investigated. Mistakes can be made :confused:

tgingold:
No, often you install the binary in a common place like /usr/local aside with other binaries.

Paebbels:
But if `path/to/ghdl/bin` is a file, and mkdir `path/to/ghdl/bin` will not fail because of this MR, a `cp file path/to/ghdl/bin/` will fail, because `bin` is not a directory and a file cannot be copied into a file. Again, there is no problem if `path/to/ghdl/bin` is a directory, because this is handled by `mkdir -p`.

This will create the exception at the wrong place. Not the `cp` is the problem, the problem is an existing file that hinders `mkdir` to create the right directory structure expected by GHDL or any other tool.

Errors need to be caught as early as possible.

tgingold:
Indeed, now reverted.

PR #587, opened 2018-05-29
add wait statement to bug_report reproduction example

1138-4EB:
As suggested by @keesj in Gitter, this PR adds a `wait` statement to the `ent.vhd` example in the `bug_report.md` template.


1138-4EB:
It is a 'fun' message, because tests had not even started when you received it. Anyway, it should be ready now.

Issue #568, opened 2018-05-08
Add support for GCC 8


andreygursky:
GCC 8.1 has been released on May 2 [1].

[1] GCC 8.1 Released
   https://gcc.gnu.org/ml/gcc/2018-05/msg00017.html

1138-4EB:
@andreygursky, do you mean to use GCC 8 to build GHDL with mcode backend? Or to use GCC 8 as the GHDL backend, building it with any other installed GCC version?

If you tried any of these and got some error, it'd be great if you could post it here.

I just tried replacing [Makefile.in#L219](https://github.com/ghdl/ghdl/blob/master/Makefile.in#L219) with:

```
7.* | 8 | 8.*)      gcc_ortho_lang=ortho-lang-7.c ;; \
```

and I get the following error when building the [gcc-8_1_0](https://github.com/gcc-mirror/gcc/archive/gcc-8_1_0-release.tar.gz) backend in Debian Buster with [these](https://github.com/ghdl/ghdl/blob/master/dist/linux/docker/build/buster#L7) dependencies:

``` bash
../../gcc-srcs/gcc/vhdl/ortho-lang.c: In function 'tree_node* type_for_mode(mach
ine_mode, int)':
../../gcc-srcs/gcc/vhdl/ortho-lang.c:707:61: error: cannot convert 'poly_uint16
{aka poly_int<1, short unsigned int>}' to 'unsigned int' for argument '1' to 'tr
ee_node* type_for_size(unsigned int, int)'
     return type_for_size (GET_MODE_BITSIZE (mode), unsignedp);
```

andreygursky:
> Or to use GCC 8 as the GHDL backend

Yes.

I've just noticed recent PR https://github.com/ghdl/ghdl/pull/567 address building with GCC 8.

vicencb:
Hi,
with #567 ghdl can be built under gcc 8.1 (with gcc 7.3), but the generated compiler is not working :( ...

```sh
$ cat e.vhd
entity e is end entity;
architecture a of e is begin end architecture;

$ ghdl -a e.vhd 

$ ghdl -e e
/usr/lib/ghdl/libgrt.a(run-bind.o): In function `grt_main__Tsec_default_sized_stacksBIP':
(.text+0xc4): undefined reference to `system__secondary_stack__ss_stackIP'
/usr/lib/ghdl/libgrt.a(run-bind.o): In function `grt_init':
[...]

Issue #576, opened 2018-05-23
conditional assignment depending on generic causes overflow


T045T:
I have a project that assigns different things to different parts of a signal depending on a generic - in essence, it zero-pads if the target register is larger than the source, and truncates the source if the target is smaller.

I implemented that using an if statement inside a process (a synchronous one in my actual code, a combinational for the reproduction case).

The code works fine and synthesizes (and runs on an FPGA) without problem using Vivado, but GHDL chokes on the if statement during elaboration:
```
$ ghdl -i --workdir=work *.vhdl && ghdl -m --workdir=work foo_tb && ghdl -r --workdir=work foo_tb
```

```
ghdl:error: overflow detected
  from: [unknown caller]
  from: work.foo_tb(tb).dut.CMP_ELAB at foo_tb.vhdl:10
  from: work.foo_tb(tb).DECL_ELAB at foo_tb.vhdl:18
  from: Elaboration of design
ghdl:error: error during elaboration
```

Am I doing something wrong here, or is this actually a bug in GHDL?

Output of `ghdl -v`:

```
GHDL 0.35 (v0.35-dirty) [Dunoon edition]
 Compiled with GNAT Version: GPL 2017 (20170515-63)
 mcode code generator
Written by Tristan Gingold.

Copyright (C) 2003 - 2015 Tristan Gingold.
GHDL is free software, covered by the GNU General Public License.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

To repro:
```vhdl
library ieee;
use ieee.std_logic_1164.all;

entity foo is
  generic(
    IN_WIDTH : positive := 8;
    OUT_WIDTH : positive := 8);
  port(
    A : in std_logic_vector(IN_WIDTH - 1 downto 0);
    X : out std_logic_vector(OUT_WIDTH - 1 downto 0));
end entity foo;

architecture repro of foo is

begin  -- architecture repro
  set_output: process (A) is
  begin  -- process set_output
    if OUT_WIDTH > IN_WIDTH then
      X(OUT_WIDTH - 1 downto IN_WIDTH) <= (others => '0');
      X(IN_WIDTH - 1 downto 0) <= A;
    else
      X <= A(OUT_WIDTH - 1 downto 0);
    end if;
  end process set_output;
end architecture repro;
```

```vhdl
library ieee;
use ieee.std_logic_1164.all;

entity foo_tb is
end entity foo_tb;

[...]

Issue #333, opened 2017-04-12
Silent overflow between physical types (64 bit integer) and Natural (32 bit integer)


Brian-Drummond:
I came across the following surprise (to me) :  a product of a natural (32 bit integer) and a ratio of two physical quantities (time_in_ms / 1 ms, a 64-bit quantity), exceeding the 32 bit range, is silently truncated to 32 bits before conversion to a real.

Something of the form 
constant CLOCK_FREQ : natural := 30000000;
real(30000000 * 200ms / 1 ms);  -- works
real(CLOCK_FREQ * 200ms / 1 ms); -- fails

Both Modelsim and Synplify have the same issue as well as (GHDL 0.34-dev (2016-09-14-331-g12586eed)), which results in silent incorrect behaviour. 

So this may be  a grey area in the LRM where the "natural" constant imposes a narrowing of the type where a "universal integer" literal doesn't, but it would seem possible and very worthwhile to at least warn of the overflow. 

I would tend to prefer a warning. I'm not sure if silently implementing correct behaviour would be a good thing since synthesis can just silently re-introduce the issue. (In fact, given a literal not a constant, Synplify does reintroduce the issue but it warns about the truncation, so I would regard correct behaviour as acceptable, if less convenient than catching the error in simulation)

Simple testcase with permission from customer...

```
entity Time_test is
end Time_test;

architecture Behavioral of Time_test is

constant c_CLOCK_FREQUENCY : natural := 32000000;
constant c_SWITCH_ON_TIME_ms : time := 200 ms; 
constant c_SWITCH_COUNT_CYCLES : integer := integer(real(c_CLOCK_FREQUENCY  * 
                                    (c_SWITCH_ON_TIME_ms / 1 ms)) / 1000.0);

begin
process is
begin
        report " Switch ON time " & time'image(c_SWITCH_ON_TIME_ms) severity NOTE;        
        report " Switch count " & natural'image(c_SWITCH_COUNT_CYCLES) severity NOTE;
        -- sanity checks on time constraints
        report "Clock frequency =  " & natural'image(c_CLOCK_FREQUENCY) & " Hz" severity NOTE;
        report "Clock period =  " & time'image( 1 sec / c_CLOCK_FREQUENCY) severity NOTE;
        report "Switch period = " & time'image ( 1 sec / c_CLOCK_FREQUENCY * c_SWITCH_COUNT_CYCLES) severity NOTE;
[...]

PR #589, opened 2018-05-29
Remove redundant use-clauses for gcc 8.1 compatibility

truls:
I had to remove some additional redundant use-clauses in order to get GHDL to compile with GCC 8.1. The compiled GHDL seems to work fine for me, but I'm not familiar with ADA or the GHDL internals so please review carefully for any breakage.

This is related to issue #568

### DIFF
--- a/src/ghdldrv/ghdlrun.adb
+++ b/src/ghdldrv/ghdlrun.adb
@@ -231,9 +231,6 @@ package body Ghdlrun is
 
    procedure Run
    is
-      use Interfaces;
-      --use Ortho_Code.Binary;
-
       function Conv is new Ada.Unchecked_Conversion
         (Source => Address, Target => Elaborate_Acc);
       Err : Boolean;
--- a/src/grt/grt-processes.adb
+++ b/src/grt/grt-processes.adb
@@ -1160,7 +1160,6 @@ package body Grt.Processes is
 
    function Simulation return Integer
    is
-      use Grt.Options;
       Status : Integer;
    begin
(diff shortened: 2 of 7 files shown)


1138-4EB:
This seems to fix the mcode variant, but not the LLVM one. See Fedora 28 builds in https://travis-ci.com/ghdl/docker/builds/74811275

1138-4EB:
This seems to fix the mcode variant, but not the LLVM one. See Fedora 28 builds in https://travis-ci.com/ghdl/docker/builds/74811275

tgingold:
Thanks!

1138-4EB:
@Brian-Drummond , might [this comment](https://github.com/ghdl/ghdl/issues/576#issuecomment-393132825) be related to the fact that after the patch in this PR mcode works but LLVM does not?

Brian-Drummond:
On Wed, 2018-05-30 at 08:25 -0700, 1138-4EB wrote:
> @Brian-Drummond , might this comment be related to the fact that
> after the patch in this PR mcode works but LLVM does not?
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.

I can't comment on that patch, what I had in mind was closed issue 333
https://github.com/ghdl/ghdl/issues/333
where after the fix, ghdl (mcode) catches real silent overflows that
Modelsim doesn't. (ghdl-gcc and llvm versions, relying on C-oriented
backends, don't catch them either)

-- Brian

PR #602, opened 2018-06-08
use fedora28 instead of fedora26

1138-4EB:
As commented by @CraigNoble in gitter, 2018-05-29 was the End-of-life for Fedora 26: https://en.wikipedia.org/wiki/Fedora_version_history#Version_history

This PR skips Fedora 27 and jumps to Fedora 28.


1138-4EB:
Thanks!

I am not sure about the `Package: Fedora (rpm)` label, since we are not building any RPM package ATM.

Paebbels:
Changed the label.

Issue #640, opened 2018-09-02
Work around current limitations of `ghdl --list-link`


1138-4EB:
I am wrapping a GHDL simulation in a C program. I have successfully done it by building a `main.c` file with a `int main(int argc, char **argv)` function that calls a function defined in a different file (`wrapper.c`). `ghdl_main` is called from the latter. The build procedure is as follows:

``` bash
$ ghdl -a -fPIC ../screen_ghdl.vhdl ../test.vhdl
$ ghdl --bind -fPIC test_screen
$ gcc -c -fPIC ../wrapper.c -o wrapper.o
$ gcc ../main.c -o ../bin/main -Wl,wrapper.o -Wl,`ghdl --list-link test_screen`
```

Now I'd like to build a shared library `libwrapper.so`, and a program (`shared.c`) that uses it:

``` bash
$ gcc -g3 -fPIC -shared -o libwrapper.so -Wl,wrapper.o -Wl,`ghdl --list-link test_screen`
$ gcc -L$PWD -o ../bin/shared ../shared.c -lwrapper -ldl
```

However, the last command fails:

``` bash
/tmp/ccHFxRna.o: In function `main':
.../hwd-ide/examples/ghdl-so/obj/../shared.c:25: undefined reference to `ghdl_mm'
collect2: error: ld returned 1 exit status
```

Indeed, the three functions defined in `wrapper.c` are not listed with `nm -D libwrapper.so`. But, they are found with objdump:

``` bash
$ objdump -d libwrapper.so | grep "<ghdl_mm>"
000000000000572e <ghdl_mm>:
$ objdump -d libwrapper.so | grep "<get_p>"
0000000000005699 <get_p>:
    613a:	e8 5a f5 ff ff       	callq  5699 <get_p>
$ objdump -d libwrapper.so | grep "<get_m>"
00000000000056f7 <get_m>:
    619f:	e8 53 f5 ff ff       	callq  56f7 <get_m>
```

**Context**

- OS: Fedora 28
- Origin: built from sources (using ghdl/docker) by adding `OPT_FLAGS+=-fPIC` and `LIB_CFLAGS+=-fPIC` to the makefile.

```
GHDL 0.36-dev (v0.35-271-g5cd3de46-dirty) [Dunoon edition]
 Compiled with GNAT Version: 8.1.1 20180712 (Red Hat 8.1.1-5)
 llvm code generator
```

**How to reproduce?**

[...]

PR #670, opened 2018-10-11
Build GHDL and libs with -fPIC

1138-4EB:
This PR forces `-fPIC` to be used when building GHDL, which involves all the libraries in `$prefix/lib/*`. This is required in order to compile VHDL sources to C objects that can then be used to build foreign applications: https://github.com/ghdl/ghdl/issues/640#issuecomment-418470167.

I don't know if this is the best solution for the problem I am trying to solve. I'd be open to any suggestion to avoid making this a fixed setting. However, I am unsure about how to pass this option when calling either `configure` or `make`.

### DIFF
--- a/dist/travis/build.sh
+++ b/dist/travis/build.sh
@@ -145,7 +145,7 @@ echo "travis_fold:start:make"
 travis_time_start
 printf "$ANSI_YELLOW[GHDL - build] Make $ANSI_NOCOLOR\n"
 set +e
-make -j$(nproc) 2>make_err.log
+make LIB_CFLAGS="$LIB_CFLAGS" OPT_FLAGS="$OPT_FLAGS" -j$(nproc) 2>make_err.log
 tail -1000 make_err.log
 set -e
 travis_time_finish


tgingold:
Does `make LIB_CFLAGS=-fPIC` work ?

1138-4EB:
It does not work with `LIB_CFLAGS`, but it does if `OPT_FLAGS` is added. So, I changed the PR in order to modify `dist/travis/build.sh` instead ok `Makefile.in`. This allows to build GHDL with e.g. `OPT_FLAGS=-fPIC LIB_CFLAGS=-fPIC ./dist/travis/build.sh -b llvm-6.0 -p ghdl-llvm-fPIC`.

tgingold:
Why are you setting LIB_CFLAGS if that doesn't work ?  Or do you need both LIB_CFLAGS and OPT_FLAGS ?

According to Makefile.in, you may set LIB_CFLAGS and GRT_FLAGS.

1138-4EB:
I do need both `LIB_CFLAGS` and `OPT_FLAGS`. The first one is required because of https://github.com/ghdl/ghdl/blob/master/Makefile.in#298. At the same time, `GRT_FLAGS` cannot be set from 'outside' because it is overwritten anyway, see https://github.com/ghdl/ghdl/blob/master/Makefile.in#L83. Thus, I am using `OPT_FLAGS` to set `GRT_FLAGS` indirectly.

1138-4EB:
@tgingold, would you mind having a look at this again?

The objective is to build a GHDL which produces libraries that can be integrated with foreign languages such as C. I think that both GHDL's internal libraries (specially `libgrt.a` as commented at [#640](https://github.com/ghdl/ghdl/issues/640#issuecomment-418470167)) and VHDL design files analized with GHDL should be compiled with `-fPIC`. At least, that's what the compiler (GCC) told me when I tried to do it otherwise.

Therefore, any solution that allows to do so by adding some option to `./dist/travis/build.sh -b llvm-6.0 -p ghdl-llvm-fPIC` will be valid for me:

- The original approach was to modify the makefile.
- The current proposal is to modify the build script only.
- It might also be possible to edit [Makefile.in#L83](https://github.com/ghdl/ghdl/blob/master/Makefile.in#L83): `GRT_FLAGS+=$(OPT_FLAGS)`. This would probably allow to set `LIB_CFLAGS` and `GRT_FLAGS` (instead of `OPT_FLAGS`). I have not tested this approach, tho.
- Any other that might involve adding a specific flag to `build.sh` and/or handle this through `configure`.

---

Ideally, it would be possible to build GHDL so that `-fPIC` is used by default for any further analysis/elaboration. That would allow user that are going to merge object with foreign languages not to specify it every time.

tgingold:
I think you will also need to have -fPIC by default (added by a configure flag?), so that std/ieee packages are also compiled with -fPIC.

1138-4EB:
> I think you will also need to have -fPIC by default (added by a configure flag?), so that std/ieee packages are also compiled with -fPIC.

I did not find this requirement, for now. Precisely:

- I set `LIB_CLFGAS` outside of `build.sh`.
- The content is passed to the makefile: `make LIB_CFLAGS="$LIB_CFLAGS"`.
- In the makefile, when the makefile inside the libraries subdir is called, `LIB_CFLAGS` is used to set `GHDL_FLAGS`: `GHDL_FLAGS="--GHDL1=$(PWD)/ghdl1-gcc$(EXEEXT) $(LIB_CFLAGS)"` (see https://github.com/ghdl/ghdl/blob/master/Makefile.in#L261).
- `GHDL_FLAGS` is used to build all the libraries (see https://github.com/ghdl/ghdl/blob/master/libraries/Makefile.inc#L349).

So, regarding all the libraries that are built with GHDL, it should be ok. I don't know if this applies to vendor libraries, since I have never built them.

Nevertheless, I think that I still need to add `-fPIC` when I analize my own VHDL sources. It would be great if there was a configure flag to set it as default. Then, if GHDL is built with `-fPIC`, all the objects generated with it will also comply. Moreover, from absolute ignorance, I wonder if this should be the default. What's the reason not to do so?

However, I am not sure about this last point. ATM, I am running some VUnit testbenches (where I do not set `-fPIC`). Then, I am analyzing some sources with `ghdl -a -fPIC`. Last, I am merging everything together (`*.o` objects from `vunit_out`, `*.o` from the workdir, and some C sources) with `gcc`. And it works. If `-fPIC` was required, objects from `vunit_out` should complain, unless VUnit is setting it. I will run some tests after updating the dependencies, and I will let you know.

PR #703, opened 2018-11-26
Accept LLVM 7 in configure script.

Paebbels:
**Description**  
This changed `configure` to accept LLVM 7.

* fixed indentation in the region of my code change
* allow LLVM 7


tgingold:
I am not sure while the builds failed, but it looks to be unrelated.

Issue #744, opened 2019-01-27
[Feature request] Add pre-built LLVM package for MacOS


twilco:
Thanks for making this awesome tool!

**Is your feature request related to a problem? Please describe.**
It would be nice if there were a pre-built LLVM or GCC GHDL package for MacOS in addition to the already available `mcode` version.  The `mcode` version seems to go quite slowly even in simple designs.  I know I can build from source...but it would be nice if I didn't have to :).  Linux and Windows both have pre-built LLVM packages.

https://ghdl.readthedocs.io/en/latest/getting/Releases.html

Arcnor:
Hi @eine , can you point me / describe what the current status of this is? If it's a matter of modifying Travis / AppVeyor / GitHub actions to build with LLVM I might be able to open a PR for it, but I'm a bit confused as to what the current status of things are.

Thanks in advance.

eine:
Hola @Arcnor!

The current status is that we do test GHDL on macOS, but only with mcode backend and with synthesis features disabled: https://github.com/ghdl/ghdl/blob/master/scripts/ci-run.sh#L400-L401. We generate a zipfile and upload it to nightly assets: https://github.com/ghdl/ghdl/releases/tag/nightly.

- In #867, synthesis features are enabled. As you can see, it fails with `ld: library not found for -lgnat-2019`: https://github.com/ghdl/ghdl/pull/867/checks?check_run_id=1805280440#step:4:896. However, users have reported building it successfully locally. Hence, it must be some issue with the setup in CI.
  - https://github.com/eine/ghdl/commits/osx-hdiutil is an attempt to use hdiutil for extracting the installer of GNAT, instead of using 7z.
- After fixing #867, it would be interesting to add an LLVM build. See issues labeled [OS: Mac OS](https://github.com/ghdl/ghdl/labels/OS%3A%20Mac%20OS) (open and closed ones). For instance #1368.
- Currently, we use a single version of macOS only (`macos-latest`, see https://github.com/actions/virtual-environments/#available-environments). It would be desirable to test on Catalina and Big Sur.
- Homebrew Formulae exists for GHDL (https://formulae.brew.sh/cask/ghdl). It might be interesting to use one of those in our CI workflow, instead of the custom shell scripts. See [ktemkin/homebrew-oss-fpga](https://github.com/ktemkin/homebrew-oss-fpga) and [hdl/packages](https://github.com/hdl/packages). For reference, Windows builds/tests are based on PKGBUILD recipes.

[...]

PR #748, opened 2019-02-03
style: install-ada

1138-4EB:
This is just some style change. On the one hand, execution of `install-ada.sh` is moved before section *Build*. This is more coherent with other platforms, where dependencies are already available when entering section *Build*. On the other hand, multiple calls to `rm` are put together.

> NOTE: this slightly affects the references in #744.


tgingold:
Thanks!

PR #753, opened 2019-02-11
VPI: add support for detecting ports and their directions

xnorme:
**Description** 
The changes enable `vpi_get(vpiType, handle_to_port)` to return `vpiPort` if the signal is a port. It also enables `vpi_get(vpiDirection, handle_to_port)` to return correct port directions(`vpiInput`, `vpiOutput`, or `vpiInout`).
The changes does not change behavior for other types of signals.

:rotating_light: Before submitting your PR, please read [contribute](http://ghdl.readthedocs.io/en/latest/contribute.html#fork-modify-and-pull-request) in the [Docs](http://ghdl.readthedocs.io/en/latest/index.html), and review the following checklist:

- [ ] DO indicate which issues are either fixed or closed by this PR. See [GitHub Help: Closing issues using keywords](https://help.github.com/articles/closing-issues-via-commit-messages/).

**When contributing to the GHDL codebase...**

- [x] DO make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [x] DO make sure you are making a pull request against the **master branch** (left side). Also you should start *your branch* off *our master*.
- [x] DO make sure that GHDL can be successfully built. See [Building GHDL](https://github.com/ghdl/ghdl#building-ghdl).
- [ ] CONSIDER adding a unit test if your PR resolves an issue.
- [ ] CONSIDER modifying the docs, at least in the TODO, if your contribution is relevant to any of the content.
- [ ] AVOID breaking the continuous integration build.
- [x] AVOID breaking the testsuite.

**When contributing to the docs...**

- [x] DO make sure that the build is successful. See [ghdl/ghdl#572 (issuecomment-390466024)](https://github.com/ghdl/ghdl/issues/572#issuecomment-390466024).

**Further comments**

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did, what alternatives you considered, etc.

:heart: Thank you!

PR #759, opened 2019-02-22
Fixing ghdl/ghdl#753 braking VPI

xnorme:
Unfortunately ghdl/ghdl#753 broke the VPI interface. Sorry but I've also added some other VPI related changes in this branch since, and I'm currently not able to cherry pick.

### DIFF
--- a/src/grt/grt-vpi.adb
+++ b/src/grt/grt-vpi.adb
@@ -198,8 +198,14 @@ package body Grt.Vpi is
 
          when vpiModule =>
             Trace ("vpiModule");
+         when vpiDefFile =>
+            Trace ("vpiDefFile");
          when vpiNet =>
             Trace ("vpiNet");
+         when vpiPort =>
+            Trace ("vpiPort");
+         when vpiDirection =>
+            Trace ("vpiDirection");
          when vpiParameter =>
             Trace ("vpiParameter");
          when vpiScope =>
@@ -627,8 +633,7 @@ package body Grt.Vpi is
          exit when Error /= AvhpiErrorOk;
 
          Kind := Vhpi_Handle_To_Vpi_Prop (Res);
-         if Kind /= vpiUndefined and then (Kind = Expected_Kind or
-           (Kind = vpiPort and Expected_Kind = vpiNet)) then
+         if Kind /= vpiUndefined and then Kind = Expected_Kind then
             return Build_vpiHandle (Res, Kind);
          end if;
       end loop;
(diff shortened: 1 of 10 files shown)


src/grt/grt-vpi.adb
```diff
@@ -176,51 +176,66 @@ package body Grt.Vpi is
...
+
+   procedure Trace_Property (V : Integer) is
+      Str: constant access constant String := Vpi_Type_To_String(V);
+   begin
+      if Str = null then
+         Trace(V);
+      else
+         Trace(Str.all);
+      end if;
    end Trace_Property;
```
tgingold: I know this is not definitive work, but why do you need Vpi_Type_To_String here ?  Trace_Property creates a memory leak.

xnorme: I was trying to be DRY here. The string representations are required both for vpi_get_str and Trace. What would be the correct way to get around this?

tgingold: In full Ada, you could simply return the string, but this is not supported in grt.

Simply allocate the strings only once:
```
vpiReset_Name : constant String_Cst := new String'("vpiReset");

...
  return VpiReset_Name;
...
```

I haven't tested that code (I will do it now), but IIRC the string will be pre-allocated.
I know, this is a little bit heavy.

tgingold: This compiles fine.

tgingold:
I wasn't able to reproduce the cocotb issue; any tip ?

xnorme:
Well, the hack I add to pass the testsuite was definitely wrong and could lead to other problems as well. 
I have a modified version of potentialventures/cocotb/pull/796 in my cocotb working branch, but AFAIR it broke my tests with cocotb/master as well. My testbenches use ValidatedBus Monitor and Driver, if that could be relevant.

src/grt/grt-vpi.adb
xnorme: Shouldn't the allocated memory be free once the last reference to it goes out of scope? I would greatly appreciate your help on this as Ada is something I still can't quite get my head around!
Travis is failing and I don't have a clue what change here is causing the VITAL test to fail.

tgingold: No, there is no garbage collection.  In Ada, memory allocated *could* be released when the access type declaration goes out of scope, but this is not the case here.

There is nothing fancy about Ada, in particular in grt.  This could be considered like C code, but with a different syntax.

IIRC, VITAL doesn't use vpi.  I will have a look at the Travis failure.

tgingold: I don't understand your change in grt-avhpi.adb: no need to add NUL here, this is done in grt-vpi.adb just after the call to vhpi_get_str.

benreynwar:
Running my current test I get the following error using this xnorme/vpi-ports branch:

Exception STORAGE_ERROR raised
Exception information:
Exception name: STORAGE_ERROR
Message: stack overflow or erroneous memory access

Using current ghdl/master I get:
vpi_get_str: unhandled property

Using a commit from before the changes to the vpi interface a week ago works fine.

I'll create a more useful minimal testcase tomorrow if that would be helpful.

tgingold:
Yes, a minimal testcase would be useful.

src/grt/grt-vpi.adb
xnorme: I was aware of the NUL termination being added in grt-vpi, but I thought maybe it would make further developments less prone to errors if the NUL is added at the source. I also don't think this affect correctness, and it's highly unlikely for it to cause any noticeable performance hit. But we can definitely remove it if you think it's not a good idea.

xnorme:
@benreynwar cocotb master has instances of calling several `vpi_get_str()`s as arguments to a single function call, e.g. in a log statement. AFAIK this is not correct as the simulator has the liberty of reusing the same buffer for consequential vpi calls (as is the case for ghdl's `vpi_get_str()`). While in your preferred loglevel the messages might not end up on the screen, the vpi's are called anyways, and the pointers to the (same) returned buffered is passed to the callee, and processed by the callee as valid null-terminated strings (which might not be the case after a consecutive vpi call. I think I have addressed that in [this branch](https://github.com/xnorme/cocotb/tree/fix-ports-ghdl). If it resolves the issue we can request a pull on cocotb with a clean patch.

src/grt/grt-vpi.adb
xnorme: Strangely the failing Travis test (gna/ticket30 on macOS) runs perfectly fine for me on macOS/llvm backend:

```
testsuite/gna/ticket30$ ./testsuite.sh 
analyze lib_numeric_tb.vhd
elaborate numeric_tb
simulate numeric_tb --stop-time=10ns --wave=numeric_tb.ghw --sdf=typ==lib_numeric_tb.sdf ( numeric_tb --stop-time=10ns --wave=numeric_tb.ghw --sdf=typ==lib_numeric_tb.sdf )
tperiod_CLK_posedge = -0.045 ns
Remove work library
Test successful

testsuite/gna/ticket30$ clang --version
Apple LLVM version 10.0.1 (clang-1001.0.37.7)
Target: x86_64-apple-darwin18.2.0

testsuite/gna/ticket30$ ghdl --version
GHDL 0.36-dev (20181129-183-g003ffb7c-dirty) [Dunoon edition]
 Compiled with GNAT Version: Community 2018 (20180523-73)
 llvm code generator
Written by Tristan Gingold.
```

benreynwar:
@xnorme When using your fix-ports-ghdl cocotb branch I am able to successfully use the current ghdl master branch.  Thanks!

tgingold:
I fear the problem is in #753: a verilog port is not equivalent to a vhdl port.  A vhdl port is almost like a signal, while in verilog a port is mainly a name used for connections.

I plan to partially revert the #753 change to create the release.  We can discuss about these patches for after the release.

xnorme:
@tgingold I think only ` [06b3ca1](/ghdl/ghdl/pull/753/commits/06b3ca162fce4348b8a5e6bf69e8060a192ee927)` needs to be reverted and the fix for testsuite/gna/issue450 in [testsuite/gna/issue450/vpi2.c](https://github.com/ghdl/ghdl/blob/404695b526aae5a3a21fecf00f9d812314525f42/testsuite/gna/issue450/vpi2.c) should be added instead. We can close this issue, fully revert #753, and then I would PR a clean patch.

tgingold:
Makes sense.

PR #762, opened 2019-02-22
VPI: Adding vpiPort detection (Corrects regression introduced by #753)

xnorme:
* Feixes regression introduced by #753
* Enables `vpi_get(vpiType, handle_to_port)` to return `vpiPort` if the signal is a port. It also enables `vpi_get(vpiDirection, handle_to_port)` to return correct port directions (`vpiInput`, `vpiOutput`, or `vpiInout`).
* [gna/issue450](https://github.com/ghdl/ghdl/tree/master/testsuite/gna/issue450) is modified to look for `vpiPort`s instead of `vpiNet`s.
* `vpi_user.h` is updated to the latest from IEEE Std 1800-2017

### DIFF
--- a/doc/building/gcc/GNULinux-GNAT.rst
+++ b/doc/building/gcc/GNULinux-GNAT.rst
@@ -25,7 +25,7 @@ GCC Backend on GNU/Linux with GCC/GNAT
  --disable-bootstrap --disable-lto --disable-multilib --disable-libssp \
  --disable-libgomp --disable-libquadmath
  $ make -j2 && make install
- $ cd /path/to/ghdl/source/dir
+ $ cd /path/to/ghdl/source/dir/build
  $ make ghdllib
  $ make install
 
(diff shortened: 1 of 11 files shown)


tgingold:
Could you rebase your branch ?  It looks like a new version of the #753 patch rather than changes from it.

I am OK for updating vpi_user.h, but this must be a separate pull request, as this is a large change.

What is different from the #753 change ?  In particular how does it fix the cocotb issue ?

gatk555:
I have a potential application that wraps an arbitrary GHDL simulation for co-simulation with SPICE.  The first step is to use VPI to discover the ports: vpi_iterate(vpiModule, NULL) and vpi_scan() find the top entity, vpi_iterate(vpiPort, top) returns a handle, but then vpi_scan() yields nothing.

It looks likely to be the same problem described here.  Is it worth spending some time to retrieve the associated Pull Request and try and put it into acceptable form?  Unfortunately my previous experience with Ada was brief and decades ago ...

gatk555:
Partial merge in #2757.

Paebbels:
@tgingold anything we can do with this old PR or is it outdated?

PR #763, opened 2019-02-23
Update badges

1138-4EB:
Thanks to some new features in `shields.io`, almost all of the previous `base64` encoded logos in the badges can now be added implicitly through `simpleicons.org`.

- Much encoded strings in the code.
- The functionality is the same. 
- I think that most of them look better now.
- base64 encoded logos were not properly rendered in LaTeX. That's why most of them were inside `.. only:: html` blocks. I don't know how will these behave.


tgingold:
Thanks!

PR #764, opened 2019-02-23
Rename travis-utils.sh to utils.sh; load ansi_color.sh, add buildCmdOpts, travis_start and travis_finish…

1138-4EB:
This PR is mostly about refactoring. These are the changes:

- `travis-utils.sh` is renamed to `utils.sh`. Travis-specific functions and now conditionally defined.
- `travis_start` and `travis_finish` are added to `utils.sh`. If `$TRAVIS` is not defined or empty, these do nothing. Otherwise, the former prints the start of a block, takes a time reference and prints a header. The latter, takes the time difference and prints the end of a block. So, these allow to replace the construct that we use multiple times from 5 lines to 2.
- `ansi_color.sh` is sourced in `utils.sh`.
- `buildCmdOpts` is moved from `travis-ci.sh` oto `utils.sh`. This is because the same computation is done in ghdl/docker, but `travis-ci.sh` files differ. Putting this in `utils.sh` allows ghdl/docker to source it, avoiding code duplication/conflicts between the two repos.


tgingold:
Thanks.

PR #768, opened 2019-02-23
add announcements for 20181129 and v0.36-rc1

1138-4EB:
Update the index of the docs to add the two latest releases to the news.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -19,11 +19,22 @@ LRM) is a plus.
    News
    ****
 
+   23.02.2019 - GHDL v0.36-rc1 was released.
+   =========================================
+
+   The major improvements are:
+
+      * more support of unbounded arrays and records
+      * support of UVVM and Vunit
+
+   29.11.2018 - GHDL 20181129 was released.
+   ========================================
+
    20.12.2017 - A new GitHub organization was created.
    ===================================================
 
-   A new GitHub organization is created and the main repo is moved from github.com/tgingold/ghdl to
-   github.com/ghdl/ghdl. Old refs will continue working, because permanent redirects are set up. However, we suggest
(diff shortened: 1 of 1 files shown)


tgingold:
I plan to do a final v0.36 release.  Should we really announce pre-releases like 20181129 and rc1 ?

1138-4EB:
Whatever you prefer. On the one hand, it's not like we need to keep updating it every month. I think that it is ok to have all of them since until now there are only three releases and six prereleases. On the other hand, yesterday I was working on the 'Releases and sources' of the documentation:

- Current output: https://ghdl.readthedocs.io/en/latest/getting/Releases.html
- New proposal: https://ghdl-devfork.readthedocs.io/en/doc-dynreleases/getting/Releases.html

The most important enhancement is that the two tables ('Downloading pre-built packages' and 'Pre-built packages of older releases') are generated automatically at compile time after retrieving a JSON response from https://api.github.com/repos/ghdl/ghdl/releases.

As a result, the decision not to announce prereleases should be applied to three different sources:

- You need to manually mark prerealeses as such, because we do not have any logic in Travis to do so. So all of them are marked as releases.
  - See:
    - https://github.com/ghdl/ghdl/releases/tag/v0.36-rc1
    - https://github.com/ghdl/ghdl/releases/tag/20181129
    - https://github.com/ghdl/ghdl/releases/tag/2017-03-01
  - This will automatically fix the shields in the README and the index of the docs:
    - [Commits since latest release](https://github.com/ghdl/ghdl/releases)
    - 
- 'Downloading pre-built packages' in the upcoming PR should show the latest non-prerelease or both the latest release and the latest pre-release?
- 'Pre-built packages of older releases' in the upcoming PR should include some identifier to tell prereleases apart. We already have this info, we just need to show it.

tgingold:
OK!

PR #769, opened 2019-02-25
doc: add custom directive 'exec', generate releases shields and tables with json data retrieved from github api

1138-4EB:
As commented in [#768](https://github.com/ghdl/ghdl/pull/768#issuecomment-466784776), this PR updates the `getting/Releases` section of the docs:

> * Current output: https://ghdl.readthedocs.io/en/latest/getting/Releases.html
> * New proposal: https://ghdl-devfork.readthedocs.io/en/doc-dynreleases/getting/Releases.html
>
> The most important enhancement is that the two tables ('Downloading pre-built packages' and 'Pre-built packages of older releases') are generated automatically at compile time after retrieving a JSON response from https://api.github.com/repos/ghdl/ghdl/releases.

A custom sphinx directive (`doc/exec.py`, see https://stackoverflow.com/questions/7250659/how-to-use-python-to-programmatically-generate-part-of-sphinx-documentation/18143318) is added, in order to allow generating parts of the documentation programatically at compilation time.

The logic to retrieve the JSON data, process it and generate RST output is defined in `doc/helpers.py`.

The most tricky part to implement is that SVG shields used in the html version are not properly processed in latex. So the helper functions include logic to generate two versions of each sphinx label. One with `image::` for html and one with `replace::` for latex. At first, I wanted to use the same label for both of them, as they should be non-conflicting because they are defined in differente `only` blocks. But this does not work. So, the labels for latex have a `l` prefix. Fortunately, this procedure does fix the tables for latex, which are currently broken in the PDF (see section 7.1 in https://media.readthedocs.org/pdf/ghdl/latest/ghdl.pdf).

Although not implemented yet, functions `printShieldSrc` and `printTab` can be used to replace other parts of the docs, such as `shields.inc` and `shieldswho.inc`. Nonetheless, I think that this PR is ready to be merged, as those enhancements can be done afterwards.

Last, script `doc/build.sh` allows to build the html and PDF versions of the docs locally, with a couple of docker containers.

---

As commented in [#768](https://github.com/ghdl/ghdl/pull/768#issuecomment-466784776), if we decide not to announce prereleases, these changes should be required in this PR:

> * 'Downloading pre-built packages' in the upcoming PR should show the latest non-prerelease or both the latest release and the latest pre-release?
> * 'Pre-built packages of older releases' in the upcoming PR should include some identifier to tell prereleases apart. We already have this info, we just need to show it.

### DIFF
--- a/doc/build.sh
+++ b/doc/build.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+
+set -e
+
+cd $(dirname $0)/..
+
+$(command -v winpty) docker run --rm -t \
+  -v /$(pwd)://src \
+  -w //src/doc \
+  btdi/sphinx:py3-featured sh -c "$(cat <<-EOF
+pip install -r requirements.txt
+sphinx-build -T -b html -D language=en . _build/html
+sphinx-build -T -b latex -D language=en . _build/latex
+EOF
+)"
+
+$(command -v winpty) docker run --rm -t \
+  -v /$(pwd)://src \
+  -w //src/doc \
+  btdi/latex sh -c "$(cat <<-EOF
+cd _build/latex
+FILE=\"\`ls *.tex | sed -e 's/\.tex//'\`\"
+pdflatex -interaction=nonstopmode \$FILE.tex;
+makeindex -s python.ist \$FILE.idx;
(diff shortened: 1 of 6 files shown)


tgingold:
Thanks!

PR #778, opened 2019-03-12
Fix build for libghdlsynth.so targets with gcc/gnat 8.3

thasti:
**Description** 
When following the instructions at https://github.com/tgingold/ghdlsynth-beta, compilation of the libghdlsynth.so target failed for me because of gnat warnings being treated as errors.
These errors all boiled down to 'use' or 'with' clauses that were considered being 'without effect', therefore this PR removes them.
This happens on a relatively up-to-date system using gcc/gnat 8.3.


tgingold:
Thanks!

PR #801, opened 2019-04-24
add some version to grt.ver (#800)

### LINKED ISSUE #640: Work around current limitations of `ghdl --list-link`

### LINKED ISSUE #800: Allow to provide a custom 'grt.ver' file as an argument to 'ghdl -e'
I am wrapping GHDL in a C application: `ghdl_main` is called from `main`. Several C functions are bind to VHDL through VHPIDIRECT, but I believe that this is irrelevant.

The executable is successfully built with `ghdl -e`. However, when I load it as shared library (e.g., from a Python script), symbols are not found. This is because of `/usr/local/lib/ghdl/grt.ver`. If I modify it to add the name of the functions that I want to be visible (see https://github.com/ghdl/ghdl/issues/640#issuecomment-418443620), it works.

I found no option to provide a custom `grt.ver` script to `ghdl -e`.

I think it would be desirable to provide a script through an arg such as `--version-script`, in order to avoid multiple users in a system being required to have separate installations of GHDL just to avoid conflicts with this file.

**Describe alternatives you've considered**

It is possible to work around this issue by using `gcc` or `llvm` directly, instead of `ghdl -e`. As commented in #640, the output of `--list-link` can be customized through `sed`. Thus, the version script can be removed, or adapted.

Nonetheless, in the current use case the binary is compiled through VUnit, which is not aware of neither `gcc` nor `llvm`. Therefore, I'd like to keep calling `ghdl -e`, to avoid adding 'compiler support' to VUnit.

umarcor:
Close #800 and Close #640.

As commented in #800, this PR adds a version (`ANY`) to `grt.ver`, so it is possible to add other visible symbols to the binary generated with `ghdl -e [args] -Wl,-Wl,--version-script=custom.ver`.

### DIFF
--- a/src/grt/grt.ver
+++ b/src/grt/grt.ver
@@ -1,4 +1,4 @@
-{
+ANY {
   global:
 vpi_chk_error;
 vpi_control;


tgingold:
Thanks!

Issue #803, opened 2019-04-25
Simulation models compiled with -shared do not exit cleanly upon failure


umarcor:
Ref #640, #670, #800, #804, #805, #1053 and #1398.

It is possible to dynamically load a PIE binary or shared library generated by GHDL. This is useful for co-simulation, as it allows to allocate memory buffers in any language compatible with C-alike objects (C/C++, Ada, Python, Golang, Rust, etc.) and have data processed by a GHDL simulation without intermediate files. It is also possible to inspect shared buffers which are used in the VHDL design. See [ghdl/ghdl-cosim](https://github.com/ghdl/ghdl-cosim).

----

Currently, failing simulations that are dynamically loaded do produce an Abortion. This forces any C or Python wrapper/caller to exit inmediately, without running any post-check. See ghdl/ghdl-cosim#15.

> [@tgingold](https://github.com/ghdl/ghdl/issues/800#issuecomment-486936060):
> About abort(): I don't think it is used during normal operations (even simulation failure). A reproducer would be useful.

---

> [@tgingold](https://github.com/ghdl/ghdl/issues/800#issuecomment-486339952)
> For the mcode backend, maybe a shared library of ghdl_mcode should be provided, like python is providing a shared library of the interpreter.
> I will think about it.

---

`-fPIC -pie -fPIE`

umarcor:
This has nothing to do with the version of GLIBC, gcc, or any other. The different behaviour is because on some systems, such as Ubuntu 18, GCC is built with `--enable-default-pie`. As a result, all the objects are compiled as if `-fPIC -pie -fPIE` was used: https://stackoverflow.com/questions/41303584/output-of-make-is-a-shared-object-and-not-an-executable

Unfortunately, prividing the flags when GHDL is compiled as follows is not enough:

```
OPT_FLAGS="-fPIC -Wl,-pie -Wl,-fPIE" \
LIB_CFLAGS="-fPIC -Wl,-pie -Wl,-fPIE" \
./dist/travis/build.sh -b llvm -p ghdl-llvm-fPIC
```

Fortunately, it is possible to achieve it by adding flags to `GRT_FLAGS` in `src/grt/Makefile.in`. I will try to isolate the minimum changes and propose a PR.

umarcor:
[...]

PR #805, opened 2019-04-25
Add configure option '--default-pic' to enable '-fPIC' by default

umarcor:
Currently, GHDL can be built as:

``` bash
OPT_FLAGS="-fPIC" \
LIB_CFLAGS="-fPIC" \
./dist/travis/build.sh -b llvm -p ghdl-llvm-fPIC
```

Then, sources must be analyzed with `-fPIC` in order to generate a PIE binary. E.g., in VUnit:

```
ui.set_compile_option("ghdl.flags", ['-fPIC'])
ui.set_sim_option("ghdl.elab_flags", ['-fPIC', '-Wl,-pie'])
```

---

This PR allows to simplify the commands above to:

``` bash
CONFIG_OPTS="--enable-default-pie " \
./dist/travis/build.sh -b llvm -p ghdl-llvm-fPIC
```

and

``` python
ui.set_sim_option("ghdl.elab_flags", ['-Wl,-pie'])
```

I.e., if GHDL is built with `--enable-default-pie`, `-fPIC` is implicitly added (by default) to any `ghdl -a` or `ghdl -e` command.

This PR sets the default of variable `default_pie` in `configure` to `false`. It is enabled automatically if gcc was built with `enable-default-pie`, as it is now. The new feature is that it can be explicitly set through `--enable-default-pie`.

However, this alone is not enough to achieve the desired result. `PIC_FLAGS` must be added to `GRT_FLAGS`, in order to force all the objects in `libghdl.a` to be position independent.

Note that I changed the values of `default_pie` from `no|yes` to `false|true`, for coherency.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -80,7 +80,7 @@ OPT_FLAGS+=-fprofile-arcs -ftest-coverage
 endif
 
 GNATFLAGS=-gnaty3befhkmr -gnatwa -gnatf $(OPT_FLAGS) $(ADA_FLAGS)
-GRT_FLAGS=$(OPT_FLAGS)
+GRT_FLAGS=$(OPT_FLAGS) $(PIC_FLAGS)
 
 WARN_CFLAGS=-Wall
 
--- a/configure
+++ b/configure
@@ -6,12 +6,12 @@
 # the terms of the GNU General Public License as published by the Free
 # Software Foundation; either version 2, or (at your option) any later
 # version.
-# 
+#
 # GHDL is distributed in the hope that it will be useful, but WITHOUT ANY
 # WARRANTY; without even the implied warranty of MERCHANTABILITY or
 # FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
 # for more details.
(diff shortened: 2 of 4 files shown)


umarcor:
@tgingold, I believe this is ready for reaview now. A test for this PR is available in #803. Running the script on `fedora:29`, `ubuntu:trusty` or `ubuntu:xenial` will not be able to load the binary as a shared library. Instead, if this branch is used (`feat-default-pie`), all of them will work.

Moreover, I think that #670 and #804 could be safely reverted with this PR, because the use cases are already covered with the now visible configure option; which is a much clearner solution, I think.

tgingold:
I have a simple request: don't hijack enable-default-pie, just add a new option like enable-pic

umarcor:
> I have a simple request: don't hijack enable-default-pie, just add a new option like enable-pic

I updated it, but I don't know if that is what you mean.

tgingold:
Thanks.  I will adjust it a little bit so that libgrt is not built with -fpic by default.

Issue #781, opened 2019-03-15
Output path for artifacts generated by ghdl --bind


1138-4EB:
Coming from gitter:

> [@1138-4EB]
> While using `ghdl --bind` I saw that '--workdir' does have no efect on the output path of the artifacts. For example:
> ```
> ghdl --bind -fPIC --std=08 --workdir=./build/obj test_demo
> ```
> The command above will find `test_demo` in `./build/obj`, because the corresponding VHDL file was analysed with `ghdl -a -fPIC --std=08 --workdir=./build/obj test_demo.vhd`. However,  `e~test_demo.lst` and `e~test_demo.o` are created in the current dir. I would expect them to be created in `--workdir`. According to the docs:  *The object files created by the compiler are always placed in the same directory as the WORK library*. Is this a bug or the expected behaviour?

> [@tgingold]
> This is the expected behaviour (maybe not well documented): the final executable is created in the current working directory.

> [@1138-4EB]
> I see... these e~* files are handled as final executables instead of intermediate objects. Would it be possible to optionally set the output dir? I can imagine that this has some effect on `--list-link` too, so it might not be straighforward. If so, please just let me know and I will work arount it with `cd` and `sed` (as in #640). But, if possible, I think it'd be useful to be able to call GHDL without producing artifacts in the current dir.

> [@tgingold]
> You can always `cd` to the destination directory and create the executable here (adjusting `-P` and `--workdir=` options).

---

The point is that I am not creating any executable with GHDL; instead, I am building objects that will be later built in a third-party tool:
- Analyze VHDL sources with `ghdl -a --workdir=build/obj ...`
- Bind the unit with `ghdl --bind --workdir=build/obj ...`
- Generate the list of arguments with `ghdl --list-list ...` and save it to a `list_file.txt`.
- Manipulate the file with sed to make paths portable.
- Save objects and `list_file.txt` in a tarball.

Then:

- Get the sources of a third-party toool, where we want to integrate GHDL by including all the objects and calling `ghdl_main`.
- Extract the tarball.
- Executing `gcc <app-specific args> -o build/bin/app $(cat lis_file.txt)` (or the equivalent with g++/cmake).

As you see, the executable is generated in the very last step, and I can locate it wherever with option `-o`. However, in the steps above I am left with objects generated by GHDL in various locations (which are all different from where the target binary is located):

- `/usr/local/lib`.
- `--workdir`.
- `./` where `ghdl --bind` was executed.

[...]

PR #806, opened 2019-04-27
Update doc: announce v0.36, add info about interfacing to other languages

### LINKED ISSUE #608: Add reference to `-gGENERIC=value` in the docs
Coming from #607. See [this](https://github.com/ghdl/ghdl/issues/607#issuecomment-397800186) comment.

@tgingold where do you think that we should put it? Is it an elaboration or a run option? Does this collide with the `-g` argument related to debugging options?

1138-4EB:
<strike>Depends on #805, because config option `--default-pic` is documented here.</strike>

This PR contains several modifications to the docs:

* announce `v0.36`
* add hint about `--synth` and `ghdlsynth`
* add option `-gGENERIC=VALUE` (Close #608)
* add hint about `--list-link` (#640) (#801)
* add hint about `--bind` (#781)
* move 'Interfacing to other languages' to it's own section, and extend it:
  * 'Linking GHDL to Ada/C'
  * 'Dynamically loading GHDL' (#805)
* fix shields not being shown in contribute nor in licenses (#770)

These changes are available at https://ghdl-devfork.readthedocs.io/en/update-doc-foreign

I am not sure about the following sentence in section 'Wrapping and starting a GHDL simulation from a foreign program': *This function must be called once, and returns 0 at the end of the simulation*. Is it still true? Or does GHDL return exit codes `/=0`?


1138-4EB:
Note that the following question is not solved:

> I am not sure about the following sentence in section 'Wrapping and starting a GHDL simulation from a foreign program': This function must be called once, and returns 0 at the end of the simulation. Is it still true? Or does GHDL return exit codes /=0?

But it is good to have this merged anyway.

tgingold:
I think ghdl returns exit codes /= 0 in case of error.  To be checked.

PR #815, opened 2019-05-05
Synth: Support for constants [WIP]

pepijndevos:
I modified the `dff02` testcase to use a constant as the reset value. This errored with not being able to handle `Iir_Kind_Constant_Declaration`, so at first I copied the variable cases to see what would break.

Next it errored on `Iir_Kind_Enumeration_Literal`, to my surprised. After playing with the debugger a bit, it appears that just adding it to the other constant cases makes the code compile, producing the same output as `dff02`.

I'm not really sure what is going on here. Why is my constant represented as an enum value? Is this because the underlying `std_logic` is an enum?

Is my solution even remotely correct? And would it work for all cases? You mentioned using the enum position in synthesis?

Does anything need to happen for `Iir_Kind_Constant_Declaration` of more complicated values?

Unrelated confusion: When looking at the output of `dff02` I don't see any influence of the reset value when I set it to `'1'`. Is this a bug?


src/synth/synth-decls.adb
```diff
@@ -76,6 +76,10 @@ package body Synth.Decls is
             --  Ignore default value.
             Make_Object (Syn_Inst, Wire_Variable, Decl);
             Create_Var_Wire (Syn_Inst, Decl, null);
+         when Iir_Kind_Constant_Declaration =>
+            null;
+         when Iir_Kind_Interface_Constant_Declaration =>
```
tgingold: I wouldn't handle interface_constant_declaration right now.

pepijndevos:
I tried a bit more, modifying `dff05` to use a constant for the `std_logic_vector`, and this leads to a different and even more puzzling error: `synth_name: cannot handle IIR_KIND_CONSTANT_DECLARATION`. I'm confused how any of this works.

tgingold:
That's normal.

Previously, you didn't handle the constant declaration in synth-decls.adb which 'synthesizes' declarations.

Now, you still don't handle it but then it is referenced in the code and you need to handle the reference.

(Previously the reference was directly replaced by the value as if the constant didn't exist).

pepijndevos:
Okay, so I need to handle `Iir_Kind_Constant_Declaration` in both `Synth_Declaration` and `Synth_Name`. So I'm guessing the reference in `Synth_Name` to the "declaration" is actually the value when it is used, which caused my confusion.

Given that I correctly implement `Synth_Declaration`, should `Get_Value` just work for constants?

Are there any significant differences/pitfalls for implementing `Synth_Declaration` for `Iir_Kind_Constant_Declaration` compared to variables? It appears the code creates a `Wire_Variable` Object and then does `Create_Var_Wire`, is that any different for constants? I found this constant gate, but not sure how to use it.

tgingold:
> Given that I correctly implement Synth_Declaration, should Get_Value just work for constants?

Yes, that would work as long as the constant can be computed.  That would be good enough for now.

> Are there any significant differences/pitfalls for implementing Synth_Declaration for Iir_Kind_Constant_Declaration compared to variables? It appears the code creates a Wire_Variable Object and then does Create_Var_Wire, is that any different for constants? I found this constant gate, but not sure how to use it.

I think this is very different.  A variable can be updated so this has to be tracked.  A constant is never updated.  So all the work for a constant has to be done when it is declared.  Handling a constant reference is easy.

pepijndevos:
Okay, so I'm having a lot of trouble figuring out how stuff works. This is where I'm at:

I looked into `Get_Value` in `synth_context.adb`, where the key operation seems to be `Val := Instance_Map (Sim_Inst.Id).Objects (Info.Slot);`. From what I gather there is a map between synth and simulator objects, and this asks the simulator for the value of a particular slot. So to make this work for constants I need to make a simulator object with a slot that has the value I want.

Right above that is `Make_Object`, which looks promising but very confusing. This function is also used in `Synth_Declaration` as follows: `Make_Object (Syn_Inst, Wire_Signal, Decl);`, but just calling that in the constant case gets me in all sorts of trouble. Just copy-pasting the whole variable case compiles alright, but you suggested constants should be very different.

Could you give any more insight in the steps required?

tgingold:
> I looked into Get_Value in synth_context.adb, where the key operation seems to be Val := Instance_Map (Sim_Inst.Id).Objects (Info.Slot);. From what I gather there is a map between synth and simulator objects, and this asks the simulator for the value of a particular slot. So to make this work for constants I need to make a simulator object with a slot that has the value I want.

Simul.Annotations gives a unique number to every declaration.  This unique number is used for the map.
The simulator is currently used to get the values of static objects, but this is a shortcut and will eventually disappear.
You can use Synth_Expression to compute the value of an expression.  In the case of a constant, you should store it in the map.

I think you should create a new kind of Wire_Kind: Wire_Constant.
> 
> Right above that is Make_Object, which looks promising but very confusing. This function is also used in Synth_Declaration as follows: Make_Object (Syn_Inst, Wire_Signal, Decl);, but just calling that in the constant case gets me in all sorts of trouble. Just copy-pasting the whole variable case compiles alright, but you suggested constants should be very different.

Make_Object fills a slot in the map so it should be used to store the value of the constant.

pepijndevos:
I looked at it for a few more hours, and I haven't figured it out yet. Current status:

So just copying the signal stuff works, but I still don't really get what `Make_Object` and `Create_Var_Wire` do. Inside `Make_Object` I also see `Create_Value_Wire` and I have no idea what they do. I just don't get the data model yet. I added a `Wire_Constant` type, which changed nothing at all.

So I took a step back and looked at `Get_Value` to see if I can reverse-engineer what it needs.
So Inside `Synth_Declaration` I seem to be able to get the value of the constant with `Syn_Inst.Sim.Objects (Slot)`. And inside `Get_value` it does a very similar `Instance_Map (Sim_Inst.Id).Objects (Info.Slot)`, except it uses `Simul.Execution.Get_Instance_By_Scope (Inst.Sim, Info.Obj_Scope)` to seemingly find the parent scope of the object, which is so far not set, and don't know how to correctly set.

So I was thinking, inside `Get_Value` I already have the object scope from which I can get the value. But again I don't get the data model at all. It seems that the objects in the object scope are of type `Iir_Value_Literal` while apparently the parent scope objects are of type `Value_Acc`.

You mentioned that `Synth_Expression` can be used to get the value from an expression, and indeed its signature is `function Synth_Expression (Syn_Inst : Synth_Instance_Acc; Expr : Iir) return Value_Acc`, so I came up with `Synth_Expression(Syn_Inst, Syn_Inst.Sim.Objects(Get_Info(Name).Slot))` which the compiler did not like at all for probably very obvious reasons. I also tried just `Synth_Expression(Syn_Inst, Name)` which just defers the problem to `Synth_Expression_With_Type` which doesn't handle `Iir_Kind_Constant_Declaration`.

That's all for this weekend.

tgingold:
Let me try to clarify.

> So just copying the signal stuff works, but I still don't really get what Make_Object and Create_Var_Wire do. Inside Make_Object I also see Create_Value_Wire and I have no idea what they do. I just don't get the data model yet. I added a Wire_Constant type, which changed nothing at all.

Some declarations are objects: variables, signals, constants and files.  Files are easy are they aren't supported for synthesis.
In an expression, a name can refer to an object, so you need to be able to get the value of it.
Variables and signals are more complex as the value may change after an assignment, while the value of a constant is defined during declaration.

To associate a value with an object, maps are used.  We cannot use a global map as an object can exist at many place: a component or an entity may be instantiated at many places, a subprogram may be activated several times.

This concept of map is defined by the simulation part (built by annotations) and reuse by synthesis.

Make_Object just create an entry in a map for an object.  The entry is a value that currently corresponds to wires.  This makes sense as only signals and variables are currently stored in that map and they correspond to hardware.

> So I took a step back and looked at Get_Value to see if I can reverse-engineer what it needs.
> So Inside Synth_Declaration I seem to be able to get the value of the constant with Syn_Inst.Sim.Objects (Slot). And inside Get_value it does a very similar Instance_Map (Sim_Inst.Id).Objects (Info.Slot), except it uses Simul.Execution.Get_Instance_By_Scope (Inst.Sim, Info.Obj_Scope) to seemingly find the parent scope of the object, which is so far not set, and don't know how to correctly set.
> 
> So I was thinking, inside Get_Value I already have the object scope from which I can get the value. But again I don't get the data model at all. It seems that the objects in the object scope are of type Iir_Value_Literal while apparently the parent scope objects are of type Value_Acc.

Yes, but in synth a value can be a net, a literal or a composite value made of them.  This is defined in synth-values
So it is possible to create a Value_Acc from an Iir_Value_Literal_Acc using Create_Value_Lit.
> 
> You mentioned that Synth_Expression can be used to get the value from an expression, and indeed its signature is function Synth_Expression (Syn_Inst : Synth_Instance_Acc; Expr : Iir) return Value_Acc, so I came up with Synth_Expression(Syn_Inst, Syn_Inst.Sim.Objects(Get_Info(Name).Slot)) which the compiler did not like at all for probably very obvious reasons. I also tried just Synth_Expression(Syn_Inst, Name) which just defers the problem to Synth_Expression_With_Type which doesn't handle Iir_Kind_Constant_Declaration.

Humm, Synth_Expression_With_Type must calls Synth_Name which has to handle Constant_Declaration.  It can simple call Get_Value - which means the value has to be stored.


I was maybe not very exhaustive; do not hesitate to ask for more clarifications.

pepijndevos:
IT WORKS?!?!

Except it seems you added a new falling edge test that created a merge conflict and does not work for me.

Also... I'm not sure this is the nicest way to do things, but I'm just extremely happy I'm making progress.

tgingold:
I will merge it and take care of the conflicts.  That's great!

pepijndevos:
Awesome!!! I was still thinking maybe I should clean it up a bit. It seems the code for std_logic and std_logic_vector is almost the same, so maybe std_logic_vector could use that function, but I guess it's fine.

So what's next? I guess I can just run some random VHDL code and see what breaks, but I don't want to bite of more than I can chew.

You mentioned in an email it's possible to synthesize VHDL and then simulate the netlist to do testing. How would I do this? That is definitely not what testbench.sh is doing at the moment.

Is there a way to gain more insight in the data types in gdb? Sometimes when I print for example an Iir value, it just prints 864 or some other seemingly arbitrary number, which makes it hard for me to understand the content of this value and the things I can do with it. So far I've only figured it out by looking at other code. So I found you can do Get_Type or Get_Default_Value on it and then it gives you... another arbitrary number.

tgingold:
If we want to simulate the result, we need to generate a correct output.  The output is now vhdl, but the identifiers are not yet correct.  This is not difficult to fix, except we must be sure that identifiers don't conflict.  And the vhdl could be simplified even more.

My current goal is to be able to synthesise the 68K00 core from suska.  The vhdl style used is not very complex and this is part of the original code.  One particularity of this code is FF written using this style:

process
begin
  wait until CLK = '0' and CLK' event
  res <= val;
end process;


Supporting that shouldn't be very hard.

To use the debugger, you should source the .gdbinit at the root of ghdl.  It provides additional commands.  To view the content of a vhdl node, use:

(gdb) pt1 n

all the nodes are represented by a number, and the function invoked by pt just print all the fields.

pepijndevos:
I'm looking into the wait thing a bit, and I'm already full of questions:

Could you tell me which codebase you're talking about?
I downloaded https://download.experiment-s.de/Archive/Suska/2K15B/Configware/rtl/vhdl.zip and there is not a single await to be seen.

It seems FF inference happens in Synth.Inference, which is only used from Synth.Environment
What is Synth.Environment doing, and what is a phi context?

I tried your gdbinit trick, and... I'm not sure what I'm looking at.
```
Breakpoint 4, synth.inference.infere (ctxt=0x9cb3e0, val=5, prev_val=3)
    at /home/pepijn/code/ghdl/src/synth/synth-inference.adb:185
185	      pragma Assert (Val /= No_Net);
(gdb) pv ctxt
16#0000_0000_009C_B3E0# e32: 3
(gdb) pt1 val
integer_type_definition [5] "CONVERTIBLE_INTEGER"
  location: *std_standard*:1:1
  resolved_flag: false
  is_ref: false
  signal_type_flag: true
  has_signal_flag: true
  type_staticness: local
  range_constraint: *null*
  type_declarator: anonymous_type_declaration "CONVERTIBLE_INTEGER" [392]
  base_type: integer_type_definition [5] "CONVERTIBLE_INTEGER"
```

I'm trying to figure out what happens if I just run the code, or modify it to use await syntax.
It gives the following error, which seems like it does not understand processes at all, which is obviously not true.
```
synth_statements: cannot handle IIR_KIND_PROCESS_STATEMENT (dff11.vhdl:12:3)

Catchpoint 1, TYPES.INTERNAL_ERROR (vhdl-errors.adb:32) at 0x0000000000508131 in vhdl.errors.error_kind (msg=..., n=842) at /home/pepijn/code/ghdl/src/vhdl/vhdl-errors.adb:32
32	      raise Internal_Error;
(gdb) bt
#0  <__gnat_debug_raise_exception> (e=0x96cb40 <types.internal_error>, message=...)
    at s-excdeb.adb:43
#1  0x000000000065c355 in ada.exceptions.complete_occurrence (x=x@entry=0x9cb540)
    at a-except.adb:931
#2  0x000000000065c365 in ada.exceptions.complete_and_propagate_occurrence (x=x@entry=0x9cb540)
    at a-except.adb:942
#3  0x000000000065c3b8 in <__gnat_raise_exception> (e=0x96cb40 <types.internal_error>, message=...)
    at a-except.adb:984
#4  0x0000000000508131 in vhdl.errors.error_kind (msg=..., n=842)
    at /home/pepijn/code/ghdl/src/vhdl/vhdl-errors.adb:32
#5  0x000000000062e13d in synth.stmts.synth_statements (syn_inst=0x9cb490, stmts=842)
    at /home/pepijn/code/ghdl/src/synth/synth-stmts.adb:823
#6  0x000000000062ed52 in synthesis.synth_entity (parent=2, arch=838, sim_inst=0x9cb250)
    at /home/pepijn/code/ghdl/src/synth/synthesis.adb:225
#7  0x000000000062eee9 in synthesis.synth_design (design=1370)
    at /home/pepijn/code/ghdl/src/synth/synthesis.adb:258
#8  0x000000000064f38a in ghdlsynth.ghdl_synth (args=...)
    at /home/pepijn/code/ghdl/src/ghdldrv/ghdlsynth.adb:119
#9  0x000000000064f3e7 in ghdlsynth.perform_action (cmd=..., args=...)
    at /home/pepijn/code/ghdl/src/ghdldrv/ghdlsynth.adb:128
#10 0x0000000000552174 in ghdlmain.main () at /home/pepijn/code/ghdl/src/ghdldrv/ghdlmain.adb:402
#11 0x0000000000654429 in ghdl_simul () at /home/pepijn/code/ghdl/src/ghdldrv/ghdl_simul.adb:37
```

So I'm guessing that clocked processes and asynchronous processes get converted to other stuff before this point. So the question is where this conversion happens.

The inference code hints at something: It seems rising_edge and falling_edge all get canonicalized to `clk='0' and clk'event` somewhere. And it mentions the following
```
   --  3)
   --  wait until rising_edge(clk);
   --   r <= x;
   --  Which is equivalent to 1) when the wait statement is the only and first
   --  statement, as it can be converted to an if statement.
```

I found there is Vhdl.Canon which seems to canonicalize various things.
My current train of thought is that the correct approach is to add a clause there that canonicalizes the await syntax to an equivalent if statement.
But this is 3K lines of uncharted territory, and no idea if I'm on the right track.

p.s. I wish Github would let you make a PR without any commits, or convert an issue into a PR later, so we can organise the discussion while there is no code written. I guess I could commit my failing await test hehe

tgingold:
> Could you tell me which codebase you're talking about?
> I downloaded https://download.experiment-s.de/Archive/Suska/2K15B/Configware/rtl/vhdl.zip and there is not a single await to be seen.

I think this is the right code.  There are some 'wait' (not 'await') at least in:
68K00/wf68k00ip_top_soc.vhd:        wait until CLK = '0' and CLK' event;

> It seems FF inference happens in Synth.Inference, which is only used from Synth.Environment
> What is Synth.Environment doing, and what is a phi context?

I have reused the name 'phi' from SSA.  If you have never heard about SSA, you should start by looking at the wiki page.  The idea is that a process (which is sequential) should be transformed to assignments.  So if you have code like this:
  if cond then
    s1 <= a;
  else
    s2 <= b;
  end if;
It should be transformed into:
   s1 := cond ? a : s1;
   s2 := cond ? s2 : b;
In order to track which signals/variables have been assigned, the structures in Synth.Environment are used.

> I tried your gdbinit trick, and... I'm not sure what I'm looking at.

Ah OK.  pt1 is used to display vhdl nodes.  If you want to display Net, you could use the subprograms from netlist-dump, like:
(gdb) call disp_driver (5)

> I'm trying to figure out what happens if I just run the code, or modify it to use await syntax.
> It gives the following error, which seems like it does not understand processes at all, which is obviously not true.
> 
> synth_statements: cannot handle IIR_KIND_PROCESS_STATEMENT (dff11.vhdl:12:3)

There are two kinds of statements: the iir_kind_process_statement and iir_kind_sensitized_process_statement.  The wait statement is only allowed in process_statement (and not in the sensitized ones).

> So I'm guessing that clocked processes and asynchronous processes get converted to other stuff before this point. So the question is where this conversion happens.

No, they aren't converted.

> 
> The inference code hints at something: It seems rising_edge and falling_edge all get canonicalized to clk='0' and clk'event somewhere. And it mentions the following
> 
>    --  3)
>    --  wait until rising_edge(clk);
>    --   r <= x;
>    --  Which is equivalent to 1) when the wait statement is the only and first
>    --  statement, as it can be converted to an if statement.
> 
> I found there is Vhdl.Canon which seems to canonicalize various things.

Yes, but most of the vhdl.canon code is disabled in ghdlsynth.adb:

      --  Do not canon concurrent statements.
      Vhdl.Canon.Canon_Flag_Concurrent_Stmts := False;

> My current train of thought is that the correct approach is to add a clause there that canonicalizes the await syntax to an equivalent if statement.

No, don't do that.  Replacing the if statement is only valid for synthesis, but not in general.

You should logically add an if_statement: create the condition and use the same operations as for an if_statement.

> p.s. I wish Github would let you make a PR without any commits, or convert an issue into a PR later, so we can organise the discussion while there is no code written. I guess I could commit my failing await test hehe

You can create an issue for the discussion, and do a PR at the end if you prefer.

Issue #757, opened 2019-02-17
Crash when using static to_string for real type


bradleyharden:
**Description**
Using the function `to_string` for a real value crashes GHDL

**Expected behaviour**
`to_string` should produce a string

**Context**
```
******************** GHDL Bug occurred ***************************
Please report this bug on https://github.com/ghdl/ghdl/issues
GHDL release: 0.36-dev (20181129-170-gf8437ed8) [Dunoon edition]
Compiled with GNAT Version: 7.3.0
Target: x86_64-linux-gnu
In directory: /home/bharden/
Command line:
/usr/local/bin/ghdl1-llvm --std=08 -P/usr/local/lib/ghdl/ieee/v08/ -P/usr/local/lib/ghdl/ -c -fpic -o test.o test.vhd
Exception TYPES.INTERNAL_ERROR raised
Exception information:
raised TYPES.INTERNAL_ERROR : errorout.adb:594
Call stack traceback locations:
0x492250 0x4a2e08 0x4ab034 0x4ab1ed 0x4ab745 0x534bce 0x549739 0x54a7ab 0x5258f8 0x526f28 0x527093 0x527ce4 0x5281ca 0x5284eb 0x558b1d 0x55f6c4 0x51d33f 0x62e9a9 0x62f561 0x630e13 0x408617 0x7f7ce5c2db95 0x4079e8 0xfffffffffffffffe
******************************************************************
```
- OS: Ubuntu 18.04
- Origin:
  - [ ] Package manager. Repo:
  - [ ] Released binaries. Tarball:
  - [x] Built from sources. Commit SHA: ~~f725da~~ f8437ed8

**How to reproduce?**
```
entity test is
end entity;

architecture tb of test is
begin
  process
  begin
    report to_string(3.0);
    wait;
  end process;
end architecture;

ghdl -a --std=08 test.vhd
```

**Checklist**
Before submitting your issue, please review the following checklist:

- [x] Add `GHDL Bug occurred` log block
- [x] Add a MWE
- [x] Try the latest version

Paebbels:
GitHub doesn't find your referenced SHA f725da. Can you confirm your hash and/or the used GHDL version.

bradleyharden:
Sorry. Fixed

bradleyharden:
I just realized, I compiled with the standard "llvm" package in Ubuntu 18.04. It looks like that is version 6. Could that be the source of the problem? Should I recompile with version 5?

[...]

PR #824, opened 2019-05-20
Add Version_String to bug message

1138-4EB:
As commented in [#757](https://github.com/ghdl/ghdl/issues/757#issuecomment-465747978), this is an attempt at adding `Version_String` (which contains the name of the backend) to the 'Bug occurred' message block. To do so, `Version_String` is defined in package `Version`, instead of `Ghdlmain`.

Regarding the suggestion by @Paebbels, I think it is not straighforward to retrieve the version of the backed from ADA, because the strings are hardcode (e.g. `"llvm code generator"`). I believe it would be neccesary for `configure` to add the info to `version.in`, just as we do for `Ghdl_Ver` and `Ghdl_Release`. @tgingold, what do you think?

> Note: in this PR some commented lines from `bug.adb` are removed, because they seemed to be redundant.

### DIFF
--- a/src/bug.adb
+++ b/src/bug.adb
@@ -71,15 +71,11 @@ package body Bug is
         ("Please report this bug on https://github.com/ghdl/ghdl/issues");
       Put_Line_Err ("GHDL release: " & Ghdl_Ver & ' ' & Ghdl_Release);
       Put_Line_Err ("Compiled with " & Get_Gnat_Version);
+      if Version.Version_String /= null then
+         Put_Line_Err (Version.Version_String.all);
+      end if;
       Put_Line_Err ("Target: " & Standard'Target_Name);
       Put_Line_Err (GNAT.Directory_Operations.Get_Current_Dir);
-      --Put_Line
-      --  ("Program name: " & Command_Name);
-      --Put_Line
-      --  ("Program arguments:");
-      --for I in 1 .. Argument_Count loop
-      --   Put_Line ("  " & Argument (I));
-      --end loop;
       Put_Line_Err ("Command line:");
       Put_Err (Command_Name);
       for I in 1 .. Argument_Count loop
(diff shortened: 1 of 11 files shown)


tgingold:
Why not, but this sets the version only in the driver (ghdl), but not in the code generator (ghdl1 or ghdl1-llvm).  I am not sure whether this is that useful.

1138-4EB:
> Why not, but this sets the version only in the driver (ghdl), but not in the code generator (ghdl1 or ghdl1-llvm). I am not sure whether this is that useful.

I did not find `GHDL Bug occurred` anywhere else in the codebase. Where is the bug block generated for `ghdl1` or `ghdl1-llvm`?

tgingold:
Yes, the same code generate the bug box.  But in your path, the version_string in not set in ghdl1 or ghdl1-llvm.  The entry point is in src/vhdl/translate/ortho_front.adb

1138-4EB:
Version_String was/is not set/required in ghdl1 or ghdl1-llvm because `--version` is not an accepted option for the generated binaries. So, when a bug is produced during `ghdl -a`, `ghdl -e`, etc. this PR is valid. But, when a bug is produced during simulation, this PR will only have effect if mcode is used. GCC and/or LLVM will produce a binary with an empty Version_String in the Bug occurred block. Is this correct?

If so, I assume that there are two possible enhancements:

- Add `--version` to the binaries generated by ghdl1 or ghdl1-llvm.
- (implicit) Reuse the Version_String in the Bug occurred block.

tgingold:
When a bug is produced during `ghdl -a` or `ghdl -e`, generally it is the code generator (ie `ghdl1`) that bugs and not the driver.  If you look at the bug messages of #757, you immediately know which backend it is according to the program name.

I don't think we should add `--version` to binaries generated, as again this isn't very useful and you can in fact mix objects generated by ghdl1 and ghdl1-llvm.

I am not against this PR, I just think it isn't very effective.

1138-4EB:
> If you look at the bug messages of #757, you immediately know which backend it is according to the program name.

> I am not against this PR, I just think it isn't very effective.

I will have a look at how to add the version number then, which would make a difference.

PR #839, opened 2019-06-14
style(setup.py): use shutil's copy and copytree

1138-4EB:
Instead of using a custom function to copy a directory recursively while ignoring files with extension `*.o`, this PR uses `shutil.copytree`. Also, `shutil.copy` is used instead of `distutils.file_util.copy_file`. Last, all unused imports are removed.

### DIFF
--- a/setup.py.in
+++ b/setup.py.in
@@ -1,11 +1,9 @@
 #!/usr/bin/env python
 
-from distutils.core import setup, Extension
-import distutils.file_util
-import distutils.dir_util
+from distutils.core import setup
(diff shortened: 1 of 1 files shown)


tgingold:
What about not copying the lib files ?  They are already installed.

1138-4EB:
> What about not copying the lib files ? They are already installed.

I thought about it. Let's keep this PR as a simple style modification and discuss it in #840. Does it sound good to you?

tgingold:
Ok.

PR #845, opened 2019-06-16
Fixes in Vendor Compilation Scripts

preusser:
**Description**
This patch fixes a path issue in the `vendors/compile_*` scripts, which failed as they were unable to source the `ansi_color.sh` include that was apparently relocated to the `dist/` directory. Additionally, the meaningless shebangs of the sole shell includes `ansi_color.sh` and `config.sh` were removed.

### DIFF
--- a/dist/ansi_color.sh
+++ b/dist/ansi_color.sh
@@ -1,5 +1,3 @@
-#!/bin/sh
-
 enable_color() {
   ENABLECOLOR='-c '
   ANSI_RED="\033[31m"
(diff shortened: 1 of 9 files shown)


tgingold:
I am OK with removing the shebangs.

But it looks like you are using the compile scripts from the git repo and not the installed ones.

While this is a possibility, the other one should be kept.

Maybe we could simply move the ansi_color.sh to the compile-* scripts dir; or even get rid of ansi_color.sh.  (I personally think that the compile scripts are slightly too complex).

preusser:
You are right. I was not aware of the scripts being installed in a different directory structure. I am sorry for this oversight. So, I will just close the request.

1138-4EB:
> Maybe we could simply move the ansi_color.sh to the compile-* scripts dir; or even get rid of ansi_color.sh. (I personally think that the compile scripts are slightly too complex).

`ansi_color.sh` is used by multiple scripts in `dist/travis`, not only by the vendor compilation scripts.

PR #846, opened 2019-06-17
fix: move src/xtools to python/xtools

1138-4EB:
In #840, I forgot to move `xtools` to the Python subdir.


tgingold:
The pnodes.py tool is referenced by some Makefile(s), so I fear you need to also modify the makefiles.

1138-4EB:
Done. However, tests did not fail, which means that none of those files was being used during the build/tests in Travis. Should be add some further test to ensure that all the makefiles are used?

tgingold:
Yes, the files created are checked in so that the build is (slightly ?) easier.
Maybe we could call the makefile (and check for no-change) in one builder.

1138-4EB:
Thanks for the explanation. Let's merge this and I will open a new issue to track the enhancement of the test set.

tgingold:
Thanks!

PR #856, opened 2019-06-24
fix(makefile): conf.py does not exist anymore

1138-4EB:
See https://travis-ci.com/ghdl/docker/jobs/210330285

Maybe the rule itself can be removed, since generating a tarball for libghdl is not required now.


tgingold:
Sure.  Thanks!

PR #862, opened 2019-07-07
Add synth testsuite to CI

1138-4EB:
In this PR:

- Option `--synth` is added to `dist/travis/build.sh` and `dist/travis/test.sh`.
- A Debian Buster mcode job is added, where GHDL is buit with `--enable-synth` and the synth testsuite is executed.
- Fedora jobs are updated from 28 to 30.
- Debian jobs are updated from Stretch to Buster.

PR #865, opened 2019-07-09
fix: *gcc* build and run docker images are now named *gcc

1138-4EB:
I did several modifications in ghdl/docker to simplify docker image names. Now:

- Images are not tagged `*mcode` and `*mcode-gpl`, the former is used only. It is fixed for this repo in #862.
- `ghdl/build:*gcc` and `ghdl/run:*gcc` images no longer have the version as a suffix. This PR is to fix this.

### DIFF
--- a/dist/travis/travis-ci.sh
+++ b/dist/travis/travis-ci.sh
@@ -2,9 +2,9 @@
 # This script is executed in the travis-ci environment.
 
 build_img_ghdl() {
-    travis_start "build_run" "$ANSI_BLUE[DOCKER build] ghdl/ghdl:${IMAGE_TAG}$ANSI_NOCOLOR"
-    docker build -t ghdl/ghdl:$IMAGE_TAG . -f-<<EOF
-FROM ghdl/run:$IMAGE_TAG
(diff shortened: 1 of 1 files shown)


tgingold:
Now merged, but still failing.

1138-4EB:
> Now merged, but still failing.

Yes. The failure with `*+gpl` is fixed in #862, not here. Let me explain:

Until yesterday, the same image was tagged `ghdl/build:stretch-mcode` and `ghdl/build:stretch-mcode-gpl`. The same applied to `ghdl/run:*`. I changed it, because it didn't make sense to have different names for the same content.

As a result, the `IMAGE=stretch+mcode+gpl` travis job now fails because it tries to use image `ghdl/build:buster-mcode-gpl`, which does not exist anymore. In #862, this is changed so that `gpl` or `synth` are provided in a different envvar. See https://github.com/ghdl/ghdl/blob/21bb312c6822e672842a4f3a7f194b867ee66a7a/.travis.yml#L13-L16

If you prefer, I can open a separate issue to fix the `gpl` jobs, which is unrelated to adding a job for `synth`.

tgingold:
We can keep everything in #862, but it cannot be merged now due to a conflict.

1138-4EB:
Let's continue in #862...

PR #868, opened 2019-07-12
doc - Adding TCE as an example user of GHDL.

TopiLeppanen:
Hello,
TCE is an **open-source** TTA-based ASIP toolset. The development is led by the Customized Parallel Computing (CPC) group at the Tampere University, Finland. The toolset provides a complete retargetable co-design flow from high-level language programs down to synthesizable processor RTL (VHDL and Verilog backends supported). Therefore, GHDL is one of the more important tools in our research. We noticed the "Who uses GHDL?"-page in your documentation and thought that perhaps you'd be open to expand it to include our toolset.
More information about the toolset: <http://openasip.org/>

BR. Topi Leppänen


tgingold:
Thanks!

PR #870, opened 2019-07-16
synth: add > and >= operators

pepijndevos:
It took a while to find all the correct spots, but after copy-pasting a bunch of stuff from `<`, `>` and the rest of the family now works as well.

I'm guessing there might be a few more cases that are not handled, but I guess this are the common ones at least.

Let me know if you'd like a testcase for this. I wasn't sure where to put it and what to put inside it.
Maybe a new folder in `testsuite/synth/operator01` or something.
There are a lot of operators and types to cover, so maybe it'd be worth making a template that gets generated for every combination?
I think I saw some HDL project that did this, can't remember where.


tgingold:
That's great.  Testcases would be welcome.  There are a few examples.

1138-4EB:
Just my two cents, so we don't get lost in translation.

I think that @tgingold meant to say that 'there are (too) few examples', instead of just acknowledging that there are 'a few' already.

@tgingold, I think that @pepijndevos was asking for some specific guidance about how to name additional tests.

@pepijndevos, I believe you can follow your own criteria nonetheless. @tgingold will later rename the tests on his own, if he needs to.

tgingold:
Yes, for the directory, you can use `operator01` or `compare01`.  It must just finish with digits.
The current scheme is to write a small vhdl file with a testbench.  The testbench is run with the original file and with the synthesized file.

PR #876, opened 2019-07-21
Add port width utility function for yosys

pepijndevos:
To create a black box cell in yosys I need to get the width of the port. I don't think this is exposed in another way, so I added an utility function.

This change is required for https://github.com/tgingold/ghdlsynth-beta/pull/25

p.s. Somehow the gates were also outdated again, because when running make it added 3 new entries.


tgingold:
Thanks!

PR #878, opened 2019-07-31
Better support for ranged integers and integer literals.

pepijndevos:
I have some code like the following that I'm trying to get to work

```vhdl
signal counter : integer range 0 to 63;
--[...]
if counter < 2 then
--[...]
counter <= counter - 1;
```

So first problem is that integer literals seem to be 64 bits.
The logic that makes `Const_UB32` just checks for the width and then fails.
But for small constants, it should be safe to use, so I changed the check to just check if the constant fits in 32 bits.

The next problem is that the comparison then tries to compare a 6-bit value with a 32-bit value, and gives an error. I am not sure how to resolve this, probably one of the values needs to be extended, or the literal needs to match the range. **UNSOLVED** So I changed my code to define constants with the correct range.

```vhdl
  constant two : integer range 0 to 63 := 2;
```

Which leads to the next problem, that a bunch of operators are not defined. Ok, I know how to do this by now.

That lead to the next problem, where `Iir_Predefined_Integer_Minus` uses `Synth_Vec_Dyadic (Id_Sub)` which calls `Create_Res_Bound` which crashes with `CONSTRAINT_ERROR : synth-expr.adb:651 discriminant check failed`. It seems from the debugger that it expects an entirely different value type. I googled it, but this goes beyond my current Ada knowledge, so **UNRESOLVED**.

But with these changes, comparing ranged integer literals works. What does not work is comparing a ranged integer with a literal, and subtracting ranged integers.

### DIFF
--- a/src/synth/netlists-disp_vhdl.adb
+++ b/src/synth/netlists-disp_vhdl.adb
@@ -623,16 +623,16 @@ package body Netlists.Disp_Vhdl is
             Disp_Template
               ("  \o0 <= std_logic_vector (resize (\ui0 * \ui1, \n0));" & NL,
                Inst, (0 => Get_Width (Get_Output (Inst, 0))));
-         when Id_Ult =>
+         when Id_Ult | Id_Slt =>
             Disp_Template ("  \o0 <= '1' when \ui0 < \ui1 else '0';" & NL,
                            Inst);
-         when Id_Ule =>
+         when Id_Ule | Id_Sle =>
             Disp_Template ("  \o0 <= '1' when \ui0 <= \ui1 else '0';" & NL,
                            Inst);
-         when Id_Ugt =>
+         when Id_Ugt | Id_Sgt =>
             Disp_Template ("  \o0 <= '1' when \ui0 > \ui1 else '0';" & NL,
                            Inst);
-         when Id_Uge =>
+         when Id_Uge | Id_Sge =>
             Disp_Template ("  \o0 <= '1' when \ui0 >= \ui1 else '0';" & NL,
                            Inst);
          when Id_Eq =>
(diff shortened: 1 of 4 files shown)


src/synth/netlists-disp_vhdl.adb
```diff
@@ -623,16 +623,16 @@ package body Netlists.Disp_Vhdl is
...
             Disp_Template ("  \o0 <= '1' when \ui0 <= \ui1 else '0';" & NL,
                            Inst);
-         when Id_Ugt =>
+         when Id_Ugt | Id_Sgt =>
             Disp_Template ("  \o0 <= '1' when \ui0 > \ui1 else '0';" & NL,
                            Inst);
-         when Id_Uge =>
+         when Id_Uge | Id_Sge =>
             Disp_Template ("  \o0 <= '1' when \ui0 >= \ui1 else '0';" & NL,
                            Inst);
```
tgingold: I think you should handle Id_S* and Id_U differently otherwise the sign won't be considered.
Use `\si0` and `\si1` for signed conversions.

src/synth/synth-context.adb
```diff
@@ -358,9 +358,11 @@ package body Synth.Context is
...
-               return Build_Const_UB32
-                 (Build_Context, Uns32 (Val.Scal), Val.Typ.Drange.W);
+            elsif Val.Scal <= Int64(Uns32'Last) then
+               return Build_Const_UB32 (
+                 Build_Context,
+                 Uns32 (Val.Scal),
+                 Width'min(32, Val.Typ.Drange.W));
             else
                raise Internal_Error;
             end if;
```
tgingold: Can you create a different PR for this ?  I fear there is an issue somewhere else; I should investigate.

pepijndevos:
I applied your requested changes and made a new PR to track the literal width problem.

Do you have any idea about the discriminant check bug?

tgingold:
Thanks!

PR #880, opened 2019-08-02
[WIP] Add support for PSL assumptions, used in formal verification

pepijndevos:
GHDL currently supports normal VHDL assertions and a subset of PSL commands, such as `cover`. Currently it does not support `assume`. During simulation, `assume` is treated as `assert`, but in formal verification it is used to constrain the inputs.

The current state of affairs in open source formal verification is that SymbiYosys only supports SystemVerilog. And even with the commercial Verific frontend, PSL in VHDL is not supported. This leads to a weird situation where code is written in VHDL, but testbenches have to be written in SystemVerilog.

My understanding is that none of the commercial players really support formal verification in VHDL very well, so by adding `assume` to GHDL, combined with the Yosys plugin, makes it a unique tool for formal verification in VHDL.

Currently, I've updated the parser by copying code from `assert` and `cover`, and tried to update the IR as well, but I'm not 100% correct if I got all the properties right. It compiles, and that's how far I've come.

Now I'll work out the rest of the bits. These are all the files that mention `Psl_Assert`, but not `Psl_Assume`. I hope I can continue to get by by copying from `assert`, as I don't have a deep understanding of what any of these files do.

- [x] src/vhdl/simulate/simul-elaboration.adb
- [x] src/vhdl/simulate/simul-simulation-main.adb
- [x] src/vhdl/translate/trans-chap9.adb
- [x] src/vhdl/translate/trans-rtis.adb
- [x] src/vhdl/vhdl-annotations.adb
- [x] src/vhdl/vhdl-canon.adb
- [x] src/vhdl/vhdl-sem_psl.adb
- [x] src/vhdl/vhdl-sem_stmts.adb


pepijndevos:
`sem_psl` notes that for `assert`
>  This is a simple assertion.  Convert to a non-PSL statement, as the handling is simpler (and the assertion doesn't need a clock).

For assume it'd be very nice if you could have assumptions without a clock. This is useful for example to assume that the design starts in reset.

But this would mean adding a "regular" assume, as a parallel to regular assert, at least internally. Or supporting uncloocked assumptions/assertions properly. Any thoughts on this?

tgingold:
One comment: PSL has no statements but directive.  So it should be psl_assume_directive (and psl_assert_directive).  Yes, assert has to be renamed.

I think you always need clock as I am not sure how clockless sequences/properties can be defined.
For the reset, you don't need to use assume but restrict.  Restrict is used to specify a subset of all the states to be checked (so you don't care about any path that doesn't start with reset), while assume is more for external interaction (you assume that every strobe is answered by an ack).

pepijndevos:
I have absolutely no idea what these translate files do, and they do some weird stuff I'm not sure I can just copy.

tgingold:
They convert the ast to a machine code ast; that's for simulation.
You should do almost the same as psl_assert, except you should call a different procedure in case of failure.

pepijndevos:
Ok, `git grep -l Psl_Assert_Statement | xargs sed -i 's/Psl_Assert_Statement/Psl_Assert_Directive/g'` to the rescue.

I'll have a look at these translate functions later. What do you mean with a different failure function? In terms of simulation, assert and assume do the same thing, right?

tgingold:
Right, but you need a specific message for Assume (as Assert has its own message).  Maybe you'd better to simply ignore assumption during simulation at first.

pepijndevos:
While I work on that, another question: What would be a good way to test this?
I found some tests related to specific issues about PSL, but I'm not sure where to add this.

pepijndevos:
I obviously messed up the parser. Needs more copy-pasting I guess. It seems to think my assume is a signal assignment inside a psl comment. Wut.

```
psl2.vhdl:28:10:error: '<=' is expected instead of "a2"
psl2.vhdl:28:12:error: ';' expected at end of signal assignment
psl2.vhdl:28:12:error: (found: ':')
psl2.vhdl:28:12:error: unexpected token ':' in a concurrent statement list
```

The code is basically `testsuite/gna/ticket24/psl.vhdl` but with assert replaced by assume.

tgingold:
To test it, you should create a new testcase in issue880/ directory.

Be careful, as according to the PSL specification, assume directive has neither message nor severity expressions.

For the parse error, you may have messed with the `;`.  At worst, use gdb to set a breakpoint and debug the parser.

pepijndevos:
Alright, I fixed the parse issue and added a simple testcase. However, I think you said it is supported to have these commands outside `-- psl` comments, but if I remove them it gives lots of errors. Why is this?

tgingold:
Thanks.  I think this is the most complex PR (up to now!).

pepijndevos:
Don't mistake size for complexity ;)

Issue #888, opened 2019-08-09
Fedora packaging: upstream patches, copr and coordination


1138-4EB:
While having a look at Fedora packaging for ghdl (https://src.fedoraproject.org/rpms/ghdl/tree/master), I found some patches that are not required anymore, and some other which are suitable to be upstreamed. Precisely:

- There is a patch for LLVM8 (ghdl-llvm8.patch), which is already supported by latest ghdl.
- There is a patch file (`ieee-mathreal.patch`) that is not used anymore.
- ghdlver is incorrectly set to 0.35dev; it should be updated to 0.37dev.
- A dummy UPF package is added (`upf.patch`), which was proposed in #436, but it is not merged yet.

On the one hand, I am preparing a PR to fix all these issues at once: https://github.com/1138-4EB/rpm-ghdl/commits/fix

On the other hand, there are 12 patches related to GCC 9 that are likely to be interesting/useful in order to add support for it here. Precisely, I think we are hitting some problem with GCC 9 in #883.

---

Moreover, since Fedora's forge supports opening 'Remote PRs' (e.g. from GitHub), I think it would be interesting to move or fork 1138-4EB/rpm-ghdl to ghdl/rpm. This would allow us to keep a better coordination.

---

We can also use copr (https://copr.fedorainfracloud.org/) to build RPM packages for Fedora rawhide, Fedora latest (30) and Fedora previous (29); for aarch64, i386 and x86_64. Potentially, other archs (ppc64le) and platforms can be added to (epel-6, epel-7, mageia-6, mageia-7, mageia-8, opensuse leap, opensuse tumbleweed...).

Note that the objective is not to replace our current CI jobs at ghdl/docker. This is specially so because our container-based CI jobs are very optimized and they take around 35'-40' instead of 45'-120'. However, in the long term, we might want to distribute 'nightly' RPMs through https://copr.fedorainfracloud.org/coprs/ghdl/ghdl, as a complement to 'stable' official releases; instead of the release tarballs for Fedora that we publish now.

Overall, I think that copr is a good playground to try the integration of ghdl, yosys and ghdlsynth, as commented in tgingold/ghdlsynth-beta#31. I.e., we can create non-official packages for ghdl and yosys, and play with different approaches to provide ghdlsynth.

---

/cc @tsailer

tgingold:
I doubt that the gcc patches fix the crash.  They appear to be small changes for configuration.

But yes, It would be nice to improve the coordination with rpm and other integrations.

[...]

PR #889, opened 2019-08-09
openieee: add dummy UPF package

1138-4EB:
Close #436.
Ref #888.

This PR is an update/rework of #436. As requested by @tgingold, the ieee and ieee2008 versions look the same, so sources are added once only. Furthermore, they are added to openieee. As a result, patching the makefile is not required.

/cc @tsailer


tgingold:
I am not sure why patching the Makefile(s) is not required: is the upf compiled ?

1138-4EB:
> I am not sure why patching the Makefile(s) is not required: is the upf compiled ?

My bad. The Makefile does not need to be patched in neither of the two places where #436 did it. But these files do need to be added to `IEEE_SRCS` when `enable_openieee` is `true`. This is now fixed.

tgingold:
Yes, upf package should be available in all cases.

PR #892, opened 2019-08-13
Support for PSL assert and assume in synthesis

pepijndevos:
This adds support for synthesis of PSL assert and assume keywords, so they can be used in SimbiYosys formal verification.

This passes my little update to the testbench, but might very well break once I start to use it for real testing. I've also not added cover yet. So it's ok to merge, but I'll likely have more changes in the coming days.

### DIFF
--- a/src/synth/synth-stmts.adb
+++ b/src/synth/synth-stmts.adb
@@ -1310,7 +1310,7 @@ package body Synth.Stmts is
       return Res;
    end Synth_Psl_NFA;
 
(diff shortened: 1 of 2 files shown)


tgingold:
I will merge it, but it would be nice to factorise code (as a follow up).

1138-4EB:
The `buster gpl` job failing in Travis is my bad. Do not care about it. I'll open a PR with a fix soon.

PR #893, opened 2019-08-15
add synthesis support for logic operators on numeric types

pepijndevos:
This makes it so you can do logic operators on `unsigned` ans `signed` without an explicit cast.
I'm not sure if there is any meaning in having them as separate IIR, but it's what the rest of the code does too.
I don't think we're handling mixed `signed` and `unsigned` operations, but who in their right mind does that anyway.

Example:
```vhdl
entity alu is
  port (
    a : in unsigned(7 downto 0);
    b : in unsigned(7 downto 0);
    y : out unsigned(7 downto 0)
  );
end alu;

architecture rtl of alu is
begin
  y <= a and b;
end rtl;
```

p.s. I accidentally committed a change to the psl testcase. I can revert and make a new pr for that if it bothers you. I don't even think `hello.vhdl` is used any more.

### DIFF
--- a/src/synth/synth-disp_vhdl.adb
+++ b/src/synth/synth-disp_vhdl.adb
@@ -25,6 +25,7 @@ with Name_Table;
 with Vhdl.Prints;
 with Vhdl.Std_Package;
 with Vhdl.Ieee.Std_Logic_1164;
+with Vhdl.Ieee.Numeric;
 with Vhdl.Errors; use Vhdl.Errors;
 with Vhdl.Utils; use Vhdl.Utils;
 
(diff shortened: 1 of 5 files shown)


tgingold:
Thanks!

src/synth/synth-disp_vhdl.adb
```diff
@@ -101,6 +102,16 @@ package body Synth.Disp_Vhdl is
                end if;
                Put_Line (";");
                Idx := Idx + 1;
+            elsif Btype = Vhdl.Ieee.Numeric.Numeric_Std_Unsigned_Type
+               or Btype = Vhdl.Ieee.Numeric.Numeric_Std_Signed_Type then
+               Put ("  wrap_" & Pfx & " <= std_logic_vector(" & Pfx);
+               if Desc.W = 1 then
+                  --  This is an array of length 1.  A scalar is used in the
+                  --  netlist.
+                  Put (" (" & Pfx & "'left)");
```
pepijndevos: @tgingold I think this is actually broken for the 1-bit case because it'll try to assign a logic value to a logic value, but wrap it in std_logic_vector

PR #897, opened 2019-08-20
ignore restrict in simulation

pepijndevos:
Code that includes a PSL restrict derivative currently synthesizes perfectly, but crashes in simulation with a `cannot handle IIR_KIND_PSL_RESTRICT_DIRECTIVE`. This PR simple makes them a no-op in various places.

### DIFF
--- a/src/vhdl/translate/trans-chap9.adb
+++ b/src/vhdl/translate/trans-chap9.adb
@@ -989,9 +989,9 @@ package body Trans.Chap9 is
             when Iir_Kind_Process_Statement
                | Iir_Kind_Sensitized_Process_Statement =>
                Translate_Process_Declarations (El);
-            when Iir_Kind_Psl_Default_Clock =>
-               null;
-            when Iir_Kind_Psl_Declaration =>
+            when Iir_Kind_Psl_Default_Clock
+               | Iir_Kind_Psl_Restrict_Directive
+               | Iir_Kind_Psl_Declaration =>
                null;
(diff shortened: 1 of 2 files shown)


tmeissner:
Why not simply map restrict to assert in simulation, like is is done for assume?

pepijndevos:
Spec says so, I assume. Or at least [this](https://zipcpu.com/tutorial/class-vhdl.pdf) tutorial does.

image

tmeissner:
Okay, I've checked with Mentor Questa Sim, that gives me an error message `PSL directive 'restrict' not supported`. So I'm fine with ignoring restrict, maybe with a warning message (when it can be easily done)?

tgingold:
According to PSL:
```
NOTE—Verification tools are not obligated to verify that the restricted sequence holds.
```

Is it the main difference between restrict and restrict_guarantee ?

tmeissner:
I don't know, in the PSL book I have they write:

> Similar to assume_guarantee, this directive instructs a verification tool to restrict the verification to traces matching a SERE, and in addition, to verify that the restriction holds in the driving block or blocks.

It's seems it's like a combination of a restrict and an assert directive on the same SERE:
```
restrict {a; b; c};
assert always {a; b; c};
```

I think, both of restrict make no real sense in simulation. First would be like a filter, that all assertions only are evaluated when the restricted signal complies to the SERE. Second would also check for the SERE to be hold.

PR #898, opened 2019-08-20
Expose source location to ghdlsynth

pepijndevos:
Adds a `Get_Source` utility function needed for https://github.com/tgingold/ghdlsynth-beta/pull/35


tgingold:
You don't need to re-allocate the string.  I will fix that.

tgingold:
Labels are now set on assert/assume gates.  So maybe the message is already better.

pepijndevos:
`Assert failed in alu: op_or` yea that's useful. Actually better than showing a line number. Thanks!

Regardless, I think this is still a useful change. Including fixing our instance naming convention to match Verilog, if possible. But that's a separate issue.

tgingold:
Yes, I still plan to add locations.

PR #900, opened 2019-08-20
Support for 2008 unary reduce operators

pepijndevos:
In VHDL-2008 there is support for unary logic operations, stuff like `and foo`. This was already supported in the parser and used in synthesizing PSL assumptions, so now I connected the dots.

For now it only supports AND/OR on std_logic_vector. I'll add support for signed/unsigned and xor and other operations later. Probably after dinner :)


tgingold:
Thanks!

PR #902, opened 2019-08-28
Integer operators

marph91:
This PR implements two missing integer operators for synthesis, i. e. `inequality` and `remainder`.


tgingold:
+1 for the test!

tgingold:
Thanks!

PR #904, opened 2019-08-30
synth: add physical division

marph91:
This PR implements the division of two physical numbers, which yields an integer.


tgingold:
Thanks.  You could put several files in a test directory (physical_division could have be put in oper01).

PR #905, opened 2019-08-31
Fix UPF

1138-4EB:
Package UPF was added in #889. Compilation was not correct and it was 'fixed' in https://github.com/ghdl/ghdl/commit/b814dbe04aa1658dc294b7db1f9005c497205f25.

The package was expected to be available in all cases (see https://github.com/ghdl/ghdl/pull/889#issuecomment-520115723), but it is currently only available when `--enable-openieee` is used. This PR fixes it, so UPF is always compiled.

Furthermore, a test is added to `sanity/006upf`. <strike>This shows that, even though compilation is successful, the package seems not to be properly built:

```
sanity 000hello: ok
sanity 001hello87: ok
sanity 002hello2008: ok
sanity 004all08: ok
sanity 005examples: ok
sanity 006upf: failed
analyze test.vhdl
ghdl: cannot load package "upf"
```
</strike>


tgingold:
Thanks!

PR #907, opened 2019-08-31
fix llvm build with synth enabled

sharkcz:
Add missing directory now that llvm backend enables the synth feature.


tgingold:
Thanks!

PR #910, opened 2019-08-31
fix(configure): ignore line ending when comparing ghdl_version and libghdl_version

1138-4EB:
On Windows (MSYS/MINGW), building after cloning the repo fails because `python/libghdl/version.py` is automatically converted to CRLF (as all other files in the repo). This PR 'fixes' the check with `cmp` in configure, so that as many bytes as the length of `ghdl_version` are compared only.


tgingold:
Thanks.  Worth adding a comment.

PR #911, opened 2019-09-01
Fix configure

1138-4EB:
This PR contains several fixes for 'configure':

- A comment is added, as suggested in #910.
- Indentation mixes tabs and spaces. It is normalized to 2 spaces.
- On MSYS2/MINGW I was getting errors because of mistaches between `srcdir` and `abs_srcdir`. One of the underlying issues was that `cmp` was not available. I added a check to print a better error message.
- On MSYS2/MINGW, `$OS` is set to `Windows_NT`. Currently, configure forces MSYS2/MINGW to use Windows paths. I fixed it so that `pwd` is used. This is required in order to execute `configure` successfully in GitHub Actions jobs.

### DIFF
--- a/configure
+++ b/configure
@@ -164,6 +164,11 @@ if ! grep $pat $srcdir/configure > /dev/null 2>&1; then
   exit 1
 fi
 
+if ! `cmp --help > /dev/null 2>&1`; then
+  echo "Sorry, you need 'cmp' to configure GHDL.";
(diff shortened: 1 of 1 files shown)


1138-4EB:
I removed the problematic modification to `Makefile.in`. It's all green now.

tgingold:
Thanks.

PR #921, opened 2019-09-11
fix: skip testsuite/sanity/005examples when 'doc' not found

### LINKED ISSUE #915: The testsuite should not depend on the docs.
Package system managers' build commands, such as `makepkg`, `git clone` a new copy of this repo in a clean temporal path, checkout a specific version, proceed with the build, and remove the temporal dir. I'm using the `prepare()` step, which is executed just before the build, to copy subdir `testsuite` out of the temporal path; so it can be executed later, after the path is removed.

The procedure works as expected, but during execution I found that [testsuite/sanity/005examples](https://github.com/ghdl/ghdl/tree/master/testsuite/sanity/005examples) depends on the docs. Precisely, `doc/using/QuickStartGuide.rst` is parsed to extract `.. code :: VHDL` blocks (a similar procedure to [issue-runner](https://github.com/1138-4EB/issue-runner)). I feel this is not as 'clean' as it could be. Possible alternatives:

1. Invert the dependency. I.e., put VHDL sources in `testsuite/sanity/005examples` and make `doc/conf.py` copy them.
2. Create a 'neutral' subdir named `examples` and put VHDL files there. Then, import them from the testsuite and/or from the docs.
3. Add a target to the Makefile that allows to 'export' the testsuite to a (standalone) dir.

I did not find this issue before because Python is usually not available in the docker containers where GHDL is built/tested. Now, I saw that 8 tests/issues have python sources:

```
# find testsuite/ -name "*.py"
testsuite/gna/issue283/example.py
testsuite/gna/issue43/run.py
testsuite/gna/issue450/cocotb/top_cocotb.py
testsuite/gna/issue456/gen.py
testsuite/gna/issue456/gen_hex.py
testsuite/gna/issue531/test.py
testsuite/gna/testsuite.py
testsuite/python/units01/show_units.py
testsuite/sanity/005examples/extract_vhdl.py
```

This is ~1.5-2% of the tests. Should we add some comment/note anywhere explaining that Python might be required for some tests?

Other requirements/dependencies of the testsuite that I found on an `archlinux/base` image are `diffutils` and `grep`.

1138-4EB:
This is a work around #915 until we guess how to properly move the examples out of `doc`.

### DIFF
--- a/testsuite/sanity/005examples/testsuite.sh
+++ b/testsuite/sanity/005examples/testsuite.sh
@@ -8,6 +8,12 @@ if ! python -V > /dev/null 2>&1; then
   exit 0
 fi
 
+# Skip the test if ../../../doc is not available.
+if [ ! -d ../../../doc ]; then
+  echo "test skipped, '../../../doc' not found"
+  exit 0
+fi
+
 # Extract examples
 python extract_vhdl.py hello.vhdl heartbeat.vhdl adder.vhdl adder_tb.vhdl < ../../../doc/using/QuickStartGuide.rst
 


tgingold:
I think this is the best approach: the purpose of this sanity check is to ensure that the doc examples are valid.  No need to be regularly tested.

PR #928, opened 2019-09-18
update testsuite

1138-4EB:
In this PR `dist/travis/test.sh` is merged into `testsuite/testsuite.sh`. Both files are similar, but currently `testsuite.sh` does not contain the `synth` group, it does not handle logs, and colours are not used. These features are merged from `test.sh`. In the second commit `test.sh` is removed.


tgingold:
OK!

Issue #927, opened 2019-09-18
synth: support PSL cover directive for formal verification


tmeissner:
**Is your feature request related to a problem? Please describe.**
It would be very nice to have synthesis of the PSL cover directive supported. With that, we could use the `mode cover` with SymbiYosys to do an analyze of reachability. At the moment, you get an error like this:

```
ghdl --synth --std=08 ../syn/WishBoneMasterE.vhd -e wishbonemastere
synth_statements: cannot handle IIR_KIND_PSL_COVER_DIRECTIVE (../syn/WishBoneMasterE.vhd:153:5)

******************** GHDL Bug occurred ***************************
Please report this bug on https://github.com/ghdl/ghdl/issues
GHDL release: 0.37-dev (v0.36-738-g1ff70cdd) [Dunoon edition]
Compiled with GNAT Version: Community 2019 (20190517-83)
Target: x86_64-apple-darwin17.7.0
/Users/torsten/Projects/git.goodcleanfun.de/libvhdl/formal/
Command line:
ghdl --synth --std=08 ../syn/WishBoneMasterE.vhd -e wishbonemastere
Exception TYPES.INTERNAL_ERROR raised
Exception information:
raised TYPES.INTERNAL_ERROR : vhdl-errors.adb:32
Load address: 0x1049ca000
Call stack traceback locations:
0x104a57cea 0x104b5229e 0x104b51bbe 0x104b5216f 0x104b6444c 0x104b644d2 0x104b610ea 0x104b9cf8a 0x104af13bc 0x104ba579d 0x1049cc35e
******************************************************************
```

**Describe the solution you'd like**
It could be implemented the same way as the other directives I assume.

**Describe alternatives you've considered**
There are no real alternatives.

**Additional context**
I've tried to added PSL cover by myself:

https://github.com/tmeissner/ghdl/commit/a5fe2d5696dd27147e057821f03bc985774061eb

But all I got is following sightly different error message:

```
ghdl --synth --std=08 ../syn/WishBoneMasterE.vhd -e wishbonemastere

******************** GHDL Bug occurred ***************************
Please report this bug on https://github.com/ghdl/ghdl/issues
GHDL release: 0.37-dev (v0.36-752-geb3d32a6-dirty) [Dunoon edition]
Compiled with GNAT Version: Community 2019 (20190517-83)
Target: x86_64-apple-darwin17.7.0
/Users/torsten/Projects/git.goodcleanfun.de/libvhdl/formal/
Command line:
[...]

PR #930, opened 2019-09-18
synth: Add support for PSL cover directive

tmeissner:
As mentioned in https://github.com/ghdl/ghdl/issues/927 here is my first pull request to implement support for PSL cover directives with synthesis. I added two test cases to check the additions.

I'm not sure about the change in `src/synth/ghdlsynth_gates.h`as the file header says to not modify that file, as it is generated. I can remove that change if necessary.

@tgingold please check and comment if there are problems with the PR. And thanks for your suggestions.


tgingold:
Thanks!

Issue #903, opened 2019-08-29
ghdl-mcode -e exits successfully but produces no output


antonblanchard:
**Description**
ghdl-gcc and ghdl-llvm both produce a binary when building the FPGA target of microwatt, but ghdl-mcode does nothing.

**Expected behaviour**
I'm expecting an error, or a binary. I likely have a bug in my VHDL, but I'm not sure where to look.

**Context**
- OS: Fedora 30
- Origin:
  - I've tried both the Fedora 30 version, and a recent build of ghdl upstream from today

**How to reproduce?**
A branch with the source is here:

https://github.com/antonblanchard/microwatt/tree/ghdl-silent-fail

To reproduce the issue:

```
GHDL=ghdl-mcode

git clone https://github.com/antonblanchard/microwatt
cd microwatt
git checkout ghdl-silent-fail

mkdir build
cd build

$GHDL -a --std=08 ../decode_types.vhdl
$GHDL -a --std=08 ../common.vhdl
$GHDL -a --std=08 ../wishbone_types.vhdl
$GHDL -a --std=08 ../fetch1.vhdl
$GHDL -a --std=08 ../fetch2.vhdl
$GHDL -a --std=08 ../decode1.vhdl
$GHDL -a --std=08 ../helpers.vhdl
$GHDL -a --std=08 ../decode2.vhdl
$GHDL -a --std=08 ../register_file.vhdl
$GHDL -a --std=08 ../cr_file.vhdl
$GHDL -a --std=08 ../crhelpers.vhdl
$GHDL -a --std=08 ../ppc_fx_insns.vhdl
$GHDL -a --std=08 ../execute1.vhdl
$GHDL -a --std=08 ../execute2.vhdl
$GHDL -a --std=08 ../loadstore1.vhdl
$GHDL -a --std=08 ../loadstore2.vhdl
$GHDL -a --std=08 ../multiply.vhdl
$GHDL -a --std=08 ../writeback.vhdl
$GHDL -a --std=08 ../wishbone_arbiter.vhdl
$GHDL -a --std=08 ../core.vhdl
$GHDL -a --std=08 ../simple_ram_behavioural_helpers.vhdl
$GHDL -a --std=08 ../simple_ram_behavioural.vhdl
$GHDL -a --std=08 ../fpga/clk_gen_bypass.vhd
$GHDL -a --std=08 ../fpga/pp_utilities.vhd
$GHDL -a --std=08 ../fpga/pp_fifo.vhd
[...]

PR #975, opened 2019-10-13
Add record (in)equality

pepijndevos:
A small step in #903 

Was it really this simple... you know how you always get suspicious when it works first try...


tgingold:
Thanks.  That's why testcases are always welcome!

PR #976, opened 2019-10-13
Show error on wait without condition

pepijndevos:
Currently GHDL can synth a `wait` statement only if it is the first statement in a process and has a clock condition. This PR just adds a check that there is an actual condition.

#903: The latest Microwatt uses a `wait` without condition at the end of a debug process. Not sure if this should be supported and how.

### DIFF
--- a/src/synth/synth-stmts.adb
+++ b/src/synth/synth-stmts.adb
@@ -1925,6 +1925,10 @@ package body Synth.Stmts is
 
       --  Handle the condition as an if.
       Cond := Get_Condition_Clause (Stmt);
+      if Cond = 0 then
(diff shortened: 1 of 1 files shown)


src/synth/synth-stmts.adb
```diff
@@ -1925,6 +1925,10 @@ package body Synth.Stmts is
 
       --  Handle the condition as an if.
       Cond := Get_Condition_Clause (Stmt);
+      if Cond = 0 then
```
tgingold: Please, use Null_Node instead of 0.
(I have to investigate why you could use 0 here, that's not expected).

tgingold:
Thanks!

PR #993, opened 2019-10-23
change file permissions from read write to executable

arcturus140:
**Description** 

some vendor compilation scripts are set with 644 permissions. This pull request changes permissions to 755. This allows to execute the compile scripts.


arcturus140:
for some reasons permissions are set to 755 on remote repo, I think this is due to a bug in my Git client GUI app.

1138-4EB:
> for some reasons permissions are set to 755 on remote repo

Where did you see that? It seems to me that this PR is legit and correct. Changes indicate `100644 → 100755`, which is expected.

> I think this is due to a bug in my Git client GUI app.

Which is your host OS? I'm finding differences between Windows and a GNU/Linux docker container:

```sh
# ls -la libraries/vendors/
total 262K
drwxr-xr-x ./
drwxr-xr-x ../
-rw-r--r-- compile-altera.ps1
-rwxr-xr-x compile-altera.sh*
-rw-r--r-- compile-intel.ps1
-rwxr-xr-x compile-intel.sh*
-rw-r--r-- compile-lattice.ps1
-rwxr-xr-x compile-lattice.sh*
-rw-r--r-- compile-osvvm.ps1
-rwxr-xr-x compile-osvvm.sh*
-rw-r--r-- compile-uvvm.ps1
-rwxr-xr-x compile-uvvm.sh*
-rw-r--r-- compile-vunit.ps1
-rw-r--r-- compile-xilinx-ise.ps1
-rwxr-xr-x compile-xilinx-ise.sh*
-rw-r--r-- compile-xilinx-vivado.ps1
-rwxr-xr-x compile-xilinx-vivado.sh*
-rw-r--r-- config.psm1
-rwxr-xr-x config.sh*
-rw-r--r-- ghdl.grcrules
-rw-r--r-- ghdl.skipwarning.grcrules
-rw-r--r-- README.md
-rw-r--r-- shared.psm1
-rw-r--r-- shared.sh

# docker run --rm -t alpine sh -c "apk add git; git clone https://github.com/ghdl/ghdl; cd ghdl; ls -la libraries/vendors"
...
drwxr-xr-x .
drwxr-xr-x ..
-rw-r--r-- README.md
-rw-r--r-- compile-altera.ps1
-rwxr-xr-x compile-altera.sh
-rw-r--r-- compile-intel.ps1
-rw-r--r-- compile-intel.sh
-rw-r--r-- compile-lattice.ps1
-rwxr-xr-x compile-lattice.sh
-rw-r--r-- compile-osvvm.ps1
-rwxr-xr-x compile-osvvm.sh
-rw-r--r-- compile-uvvm.ps1
-rw-r--r-- compile-uvvm.sh
-rw-r--r-- compile-vunit.ps1
-rw-r--r-- compile-xilinx-ise.ps1
-rwxr-xr-x compile-xilinx-ise.sh
-rw-r--r-- compile-xilinx-vivado.ps1
-rwxr-xr-x compile-xilinx-vivado.sh
-rw-r--r-- config.psm1
-rw-r--r-- config.sh
-rw-r--r-- ghdl.grcrules
-rw-r--r-- ghdl.skipwarning.grcrules
-rw-r--r-- shared.psm1
-rw-r--r-- shared.sh
```

It seems that, on windows, all the scripts are checked out as executables. However, on GNU/Linux, some are executable and some are not.

arcturus140:
what I meant is that I intended to set `744` permissions instead of `755`. Turns out there is no easy way to do this: https://stackoverflow.com/questions/10516201/updating-and-committing-only-a-files-permissions-using-git-version-control

`755` permissions is legit. The differences that you see on Windows are due to the file system, as `NTFS` does not support file / folder privileges. Changes are only valid for `EXT` file systems and other.

1138-4EB:
> what I meant is that I intended to set `744` permissions instead of `755`

Oh, I missed that. I was only look at `6 -> 7`.

> The differences that you see on Windows are due to the file system, as NTFS does not support file / folder privileges.

Yes. The point is that it happens to me frequently that I create a script, I use it without any problem on MSYS and when I push it CI fails. So, it is easy but annoying that NTFS just ignores those settings.

arcturus140:
I am not sure but I believe git preserves file permissions from commits on NTFS. Otherwise there would be an tremendous amount of unintended permission changes made :laughing: 

> it happens to me frequently that I create a script, I use it without any problem on MSYS and when I push it CI fails.

This is one reason I switched to Linux, it is more flexible in this regards. I started mounting NTFS partitions with preset permissions because otherwise all files on NTFS had 777 which is really bad for Linux. Similar solutions exist for Windows, but it still has too many limitations. I got fed up and moved everything to EXT4 recently.

arcturus140:
if you create a script, you can use `git update-index --chmod=+x script.sh` to push it for CI.

1138-4EB:
> if you create a script, you can use `git update-index --chmod=+x script.sh` to push it for CI.

Precisely, in order to check/reproduce this PR, I did `git update-index --chmod=+x libraries/vendors/*.sh`. However, I normally forget it. On GNU/Linux, I need to `chmod` in order to test locally, which prevents me from pushing files with wrong permissions. But that's not the case on Windows. As a matter of fact, I think that the reason why this PR is required might be that @Paebbels pushed some scripts after adding and testing them on Windows.

tgingold:
Thanks!

PR #996, opened 2019-10-28
Enhance documentation of PSL support

tmeissner:
This is a work in process of enhancing the documentation of PSL support.

Changes are made to the PSL section under:
IMPLEMENTATION REFERENCES -> Implementation of VHDL -> Implementation of PSL.

You can see the result on the [Read the docs](https://ghdl-tmeissner.readthedocs.io/en/latest/references/ImplementationOfVHDL.html#psl-support) page of my fork.

@1138-4EB, @tgingold & others: hints and tips are welcome :)

PR #1003, opened 2019-11-05
Update doc

1138-4EB:
This PR includes multiple updates to the doc:

- A GitHub Action is added to build (test) the doc after each push or PR. html, latex and man targets are built. This does NOT replace readthedocs (yet).
- The makefile and helper scripts of the docs are updated.
- (sub)sections are reorganized: Usage and References are merged, Getting is moved first, etc.
- Section 'Development' is added. New subsections 'Debugging' and 'Synthesis' are created.
- Section 'Usage / Examples' is added.
  - A new subdir named `examples` is created, as commented in #996. Each of the examples of the existing 'Quick Start Guide' is split to a separate subdir. Test `sanity/005examples` is modified accordingly.
  - Brief descriptions are added about two projects where VHPIDIRECT is used (VUnit and ghdlex).
- Several definitions of GHDL options and the references to them are fixed to avoid Sphinx warnings. Headings and spaces are fixed too.
- Notes about possible enhancements to the docs are added to subsection 'Roadmap'.
- `.. program: ghdl` directives are removed and the standard domain is used instead.

A demo of this PR is available at https://ghdl-devfork.readthedocs.io/en/doc-egs/index.html


tgingold:
Question: why did you use ``.vhd`` instead of ``.vhdl`` ?  I prefer to promote ``.vhdl`` because ``.vhd`` is often interpreted as a VM virtual harddisk...

1138-4EB:
> Question: why did you use `.vhd` instead of `.vhdl` ? I prefer to promote `.vhdl` because `.vhd` is often interpreted as a VM virtual harddisk...

It was only because most extensions have three characters and because that's the extensions used in VUnit. Anyway, I will change it now.

tmeissner:
I would move the examples section at the end of the ghdl usage section, as it contains some examples of features that are explained in the implementation of Vhdl section for example.

1138-4EB:
> I would move the examples section at the end of the ghdl usage section, as it contains some examples of features that are explained in the implementation of Vhdl section for example.

I thought about it, and I am not conviced with the current layout. However, I don't know if moving Examples to the end of Usage is desirable. While I understand your view, my proposal was the opposite:

I expect (new) users to read About GHDL, to Get/Install it, and then jump to some Examples straightaway to see how it works. Hence, they will not need to go through all the options and implementation details before finding something 'runnable'. Furthermore, I am adding many links from the Examples to the descriptions of commands and options. This allows users to 'peek forward' as they find terms they don't know, in order to get a feeling of what they'll find in following sections.

If we move Examples to the end, the (most) impatient will be jumping forth and back, which will make it harder for them to get a broad view of the doc. At the same time, I acknowledge that Examples about PSL or co-simulation with foreign languages are probably 'advanced enough' to not fit at the beginning.

Maybe we should keep both subsections 'Quick Start Guide' at the beginning and 'Examples' at the end? We can keep the same layout for the sources and just change the import/includes that generate the sidebar.

tmeissner:
Okay, I can live with that. Maybe we should divide the examples section in two subsections called basic & advanced examples. All PSL, VHPI and such stuff would be added to advanced examples.

1138-4EB:
@tgingold, I think that this is ready to merge.

tgingold:
Great work!

PR #1024, opened 2019-11-20
ci: fix envvar name to run synth tests

1138-4EB:
The name of the envvar that is checked to decide whether to run the synth test suite is not correct. `ISEXTRA` is checked, and it should be `EXTRA`. As a result, currently issues with the synth test suite are not being checked: https://travis-ci.org/ghdl/ghdl/jobs/614170911#L1270

Actually, this PR fails because there is currently some issue with `master`: https://travis-ci.org/1138-4EB/ghdl/jobs/614291974#L1836

### DIFF
--- a/dist/travis/travis-ci.sh
+++ b/dist/travis/travis-ci.sh
@@ -98,7 +98,7 @@ else
       tests="$tests gna"
     fi
     tests="$tests vests"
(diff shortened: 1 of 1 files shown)


tgingold:
Thanks.  Now I have to fix the synth testsuite.

PR #1061, opened 2019-12-31
Add option for compiling Wishbone VIP for UVVM framework.

m-kru:
Wishbone VIP is located in [UVVM_Community_VIPs](https://github.com/UVVM/UVVM_Community_VIPs) repository. To compile it needs to be copied to [UVVM](https://github.com/UVVM/UVVM) directory. Why and how is explained in this issue [UVVM_Community_VIPs issue 2](https://github.com/UVVM/UVVM_Community_VIPs/issues/2).
When it is not copied, and you compile with `--all` flag, compilation of standard VIPs from [UVVM](https://github.com/UVVM/UVVM) repository success, so this commit does not break anything.


tgingold:
Thanks!

PR #1065, opened 2020-01-03
Homogenise NEWS.md, conf.py cleanup

eine:
This PR includes three style modifications:

- `NEWS.md` is homogenised.
- `conf.py` is cleaned by removing all the commented options that are not used. The file was generated 4 years ago and some of the options are likely to be different now. Moreover, an up to date example configuration file is available at https://www.sphinx-doc.org/en/1.8/usage/configuration.html#example-of-configuration-file.
- An anchor is renamed in `.travis.yml`. The name does not correspond to the content and it can be misleading.


tgingold:
Thanks!

PR #1070, opened 2020-01-07
fix btdi image names

eine:
The names of images that are used in `man.sh` and `make.sh` changed. This PR fixes them so that job 'man' in Travis is not broken.


tgingold:
Thanks!

PR #1071, opened 2020-01-07
fix gnat-gpl download URL (osx)

eine:
Apparently, AdaCore changed the URLs of the tarballs, and osx builds are failing, unless GNAT GPL is already cached in the CI system. This PR updates the URL to fix it. I had to use `wget` instead of `curl`, because I couldn't guess how to make the latter work.


tgingold:
Thanks.

PR #1072, opened 2020-01-07
ci: update installation of GNAT GPL on macOS

eine:
Close #1071 

This PR contains a few non-critical updates to the macOS environments in CI:

- GNAT GPL is updated to 2019. Unlike 2017, this version is only available as a DMG file, not as a tarball. Hence, `7z` is used to extract it, instead of `tar`.
- Xcode in Travis is updated to 10.
- Instead of executing the installation of GNAT GPL in the ci run script, it is done in the YAML files of the CI services. The motivation for this is that it is a dependency that is required in the system, rather than a piece of GHDL. On other platforms, dependencies are handled outside of the build scripts.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -31,7 +31,10 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - run: |
+        brew update
+        brew install p7zip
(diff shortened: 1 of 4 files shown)


tgingold:
Thanks!

PR #1082, opened 2020-01-11
netlists: Handle sdiv and udiv

antonblanchard:
I hit an issue with vhdl that uses divides when using `ghdl --synth`.


antonblanchard:
I haven't had to venture into this part of ghdl before, so a review is welcome. A test case:
```
library ieee;
use ieee.std_logic_1164.all;
use ieee.numeric_std.all;

entity test is
    port(
	a_in     : in  std_logic_vector(31 downto 0);
	b_out     : out  std_logic_vector(31 downto 0)
	);
end test;

architecture rtl of test is
begin
    process(all)
    begin
	b_out <= std_logic_vector(to_unsigned((31-to_integer(unsigned(a_in))) / 4, 32));
    end process;
end;
```

tgingold:
Thanks!

PR #1084, opened 2020-01-12
ci: the URL to retrieve GCC sources has changed

eine:
It seems that tags in GCC's mirror at GitHub were renamed, which broke our CI job. This PR fixes it. By the way, `codeload.github.com` is used.


tgingold:
Thanks.

PR #1085, opened 2020-01-12
testsuite/synth: add test from #1082

eine:
In this PR, https://github.com/ghdl/ghdl/pull/1082#issuecomment-573291835 is added as a test.


tgingold:
Thanks.

PR #1102, opened 2020-01-19
enable synth by default

eine:
In this PR, `enable_synth=false` is changed to `enable_synth=true`, so that synthesis features are enabled by default. Since `libghdl` is also enabled by default, after this PR is merged, any regular GHDL build will work with ghdlsynth-beta, without requiring GHDL to be rebuilt.

Synthesis features require GNAT>2017 and GCC>=8. Users with older versions might need to either update or use `--disable-synthesis` when configuring the build.

Currently, the job with `--gpl` fails if synthesis is enabled too (new default). @tgingold, shall I disable synth in this case or should the GPL variant not crash?


tgingold:
It is because some aggregates are not supported by synthesis.

Disable synth for the gpl build.

eine:
Done. GPL builds pass now. Ubuntu 16 jobs in Travis fail because of the known issue.

eine:
@tgingold, shall we drop/deprecate Ubuntu 16 and use Ubuntu 18 instead?

tgingold:
Yes, that makes sense.

eine:
Done!

tgingold:
Thanks!

PR #1105, opened 2020-01-19
update README

eine:
This is an update to the README:

- The comment about the project being moved to an org is removed.
- Shields for GHA workflows are added.
- A logo is added to the CC shields and the colour is changed.
- The table from subsection "Building GHDL" is removed and users are pointed to the full table in the docs.


tgingold:
Thanks.

Issue #883, opened 2019-08-06
Elaboration error with sid-gcc-9.1.0


sjaeckel:
**Description**
VUnit's `vhdl/verification_components/src/axi_stream_monitor.vhd` can't be elaborated as soon as coverage is enabled.

**Expected behaviour**
That the elaboration also works with coverage enabled.

**Context**

```
...
Compiling into vunit_lib: ../usr/local/lib/python3.7/dist-packages/vunit/vhdl/verification_components/src/axi_stream_monitor.vhd          failed
=== Command used: ===
/usr/local/bin/ghdl -a --workdir=/work/vunit_out/ghdl/libraries/vunit_lib --work=vunit_lib --std=08 -P/work/vunit_out/ghdl/libraries/vunit_lib -P/work/vunit_out/ghdl/libraries/osvvm -P/work/vunit_out/ghdl/libraries/t -g -fprofile-arcs -ftest-coverage /usr/local/lib/python3.7/dist-packages/vunit/vhdl/verification_components/src/axi_stream_monitor.vhd

=== Command output: ===
during IPA pass: profile
/usr/local/lib/python3.7/dist-packages/vunit/vhdl/verification_components/src/axi_stream_monitor.vhd: In function ‘vunit_lib__axi_stream_monitor__ARCH__a__axi_stream_protocol_checker_generate__B1__axi_stream_protocol_checker_inst__COMP_ELAB’:
/usr/local/lib/python3.7/dist-packages/vunit/vhdl/verification_components/src/axi_stream_monitor.vhd:35: internal compiler error: in coverage_begin_function, at coverage.c:656
   35 |     variable axi_stream_transaction : axi_stream_transaction_t(
      | 
0x609efc coverage_begin_function(unsigned int, unsigned int)
        ../../gcc-srcs/gcc/coverage.c:458
0xac5731 branch_prob(bool)
        ../../gcc-srcs/gcc/profile.c:1233
0xbff182 tree_profiling
        ../../gcc-srcs/gcc/tree-profile.c:793
0xbff182 execute
        ../../gcc-srcs/gcc/tree-profile.c:898
Please submit a full bug report,
with preprocessed source if appropriate.
Please include the complete backtrace with any bug report.
See <https://gcc.gnu.org/bugs/> for instructions.
/usr/local/bin/ghdl: exec error
```
- OS: 
  - not relevant (can be provided if required)
  - Docker version 19.03.1, build 74b1e89e8a

- Origin:
[...]

PR #1119, opened 2020-01-29
update for gcc10

sharkcz:
No change has been required for ortho-lang-9.c so I have solved it this way. Build-tested only so far.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -248,8 +248,9 @@ copy-sources.gcc: version.ads
 	 7.*)      gcc_ortho_lang=ortho-lang-7.c ;; \
 	 8.*)      gcc_ortho_lang=ortho-lang-8.c ;; \
 	 9.*)      gcc_ortho_lang=ortho-lang-9.c ;; \
+	 10.*)     gcc_ortho_lang=ortho-lang-9.c ;; \
 	 *) echo "Mismatch gcc version from $(gcc_src_dir)"; \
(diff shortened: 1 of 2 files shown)


eine:
Hi @sharkcz, on the one hand, would you mind updating https://github.com/ghdl/ghdl/blob/master/doc/getting/GCC.rst accordingly? It's outdated independently of your PR, but I'd be glad if you could do it.

On the other hand, some errors have been reported with GCC 9.1.0 on Debian Sid (see https://github.com/ghdl/ghdl/issues/883#issuecomment-518815957). Did you find any problem with GCC 9 on Fedora?

sharkcz:
Sure, will update the docs too.
And no, I am not aware of any issues with gcc9 on Fedora. It passed the whole test-suite when I tried and also I see no issues when simulating Microwatt. Both is on F-30 ppc64le host, so I can't talk about x86 :-)

eine:
> Sure, will update the docs too.

Thanks!

> And no, I am not aware of any issues with gcc9 on Fedora. It passed the whole test-suite when I tried and also I see no issues when simulating Microwatt. Both is on F-30 ppc64le host, so I can't talk about x86 :-)

Actually, I'm testing 9.2.0 on Fedora 31 on amd64: https://github.com/ghdl/docker/runs/406238872?check_suite_focus=true . Hence, it seems that we are good on the Fedora ecosystem. Good!

tgingold:
Thanks!

PR #1142, opened 2020-02-19
disable backtrace on android

umarcor:
This PR disables backtrace on Android, since that feature is not included in the Bionic C library. This allows to follow a regular build procedure to have GHDL available on smartphones and tablets!


tgingold:
Thanks!

PR #1183, opened 2020-03-31
python: fix document URI creation on Windows

MaximilianKoestler:
On Windows, a URI has to begin with file:/// (3 slashes!).
Now, the correct implementation from lsp.py is used to create the URIs.


tgingold:
Thanks!

Issue #1179, opened 2020-03-30
synth: synth_static_dyadic_predefined: unhandled IIR_PREDEFINED_IEEE_NUMERIC_STD_MUL_NAT_UNS


bwimm:
**Description**
Attempting to synthesize the attached code fails.

**Expected behaviour**
Successful synthesis

**How to reproduce?**

```vhd :file: issue.vhd
library IEEE;
use IEEE.std_logic_1164.all;
use IEEE.numeric_std.all;

entity bug is
	port(
		dummy : in std_ulogic
	);
end bug;

architecture behav of bug is
	constant A : positive := 4;
	constant B : positive := 1100;
	constant C : positive := to_integer(A * to_unsigned(B, 11));
begin

end architecture;
```

```
ghdl --synth --std=08 issue.vhdl -e bug
```

Output
```
issue.vhd:15:46:error: synth_static_dyadic_predefined: unhandled IIR_PREDEFINED_IEEE_NUMERIC_STD_MUL_NAT_UNS
```
followed by the CONSTRAINT_ERROR below

**Context**
Please, provide the following information:

- OS:
- Origin:
  - [ ] Package manager: `version`
  - [ ] Released binaries: `tarball_url`
  - [X] Built from sources: `42b91517`

If a `GHDL Bug occurred` block is shown in the log, please paste it here:

```
******************** GHDL Bug occurred ***************************
Please report this bug on https://github.com/ghdl/ghdl/issues
GHDL release: 1.0-dev (v0.37.0-194-g42b91517) [Dunoon edition]
Compiled with GNAT Version: 8.3.0
Target: x86_64-linux-gnu
/home/user/Desktop/tmp/
Command line:
ghdl --synth --std=08 issue.vhd -e bug
Exception CONSTRAINT_ERROR raised
Exception information:
raised CONSTRAINT_ERROR : synth-values.adb:44 access check failed
Call stack traceback locations:
0x55c057971531 0x55c057a5ef09 0x55c057a73782 0x55c057a76853 0x55c057a77626 0x55c057a778a8 0x55c057a46b50 0x55c057a46cce 0x55c057a40626 0x55c057aa56fb 0x55c0579e5b19 0x55c057b5028f 0x55c0577a99a3 0x7f1361dd8b69 0x55c0577a81a8 0xfffffffffffffffe
******************************************************************
```

PR #1200, opened 2020-04-07
synth-static_oper: handle mul uns nat.

bwimm:
**Description** 
Implement ```unsigned * natural``` similar to
```natural * unsigned``` in issue #1179


tgingold:
Thanks!

Issue #1185, opened 2020-03-31
Pass Toplevel generic to ghdl_main()


RocketRoss:
**Description**
I am not getting any valid parsing of a argc, argv pair when calling ghdl_main() in a custom main(). I noted that the documented example of 'ghdl -r --std-08 bugtest -gDEPTH=12' was a little picky, and only worked when '-g...' was at the end.

**Expected behaviour**
The generics default to inexplicit values (Integer = -2147483648, a default I did not specify). I want the generic to take on the value after the '=' in the '-gDEPTH=12'. I then struggled to pass a value for a std_logic generic. What are the limits of '-g'?

I also looked at passing ghdl_main all four arguments (-r, --std-08, bugtest, -gDEPTH=12) instead of just '-gDEPTH=12', and including a leading space on the arguments. No avail.

**How to reproduce?**

```vhd :file: BugTest.vhd
library ieee;

entity BugTest is
	generic (
		depth : integer
	);
end entity BugTest;

architecture RTL of BugTest is
	
begin
	process
	begin
		report "Toplevel generic value is: " & integer'image(depth);
		wait;
	end process;
end architecture RTL;

```

```c :file: main.c
#include <stdio.h>

extern int ghdl_main(char argc, char* argv[]);

int main(int argc, char const *argv[])
{
    char argPtr0[] = " -r";
    char argPtr1[] = " --std=08";   
    char argPtr2[] = " bugtest";
    char argPtr3[] = " -gDEPTH=12";
    
    char* argPtr[] = {/*argPtr0, argPtr1, argPtr2, */argPtr3};

    printf("\nargs: [%s,%s,%s,%s]\n", argPtr[0], argPtr[1], argPtr[2], argPtr[3]);
    printf("ghdl_main return: %d\n", ghdl_main(1, argPtr+3)); //including other arguments causes flag conflicts
    
    return 0;
}
```

```sh :image: ghdl/ghdl:buster-llvm
gcc -c main.c &&
ghdl -a BugTest.vhd &&
ghdl -e -Wl,main.o bugtest &&
./bugtest &&
rm *.o work-obj93.cf bugtest

echo Below works
ghdl -a BugTest.vhd &&
ghdl -e BugTest &&
ghdl -r --std-08 bugtest -gDEPTH=12 &&
rm *.o work-obj93.cf bugtest
```
**Context**
$ ghdl --version
[...]

Issue #1184, opened 2020-03-31
Second call of ghdl_main() aborts


RocketRoss:
**Description**
With a custom main(), the second call of ghdl_main(0, NULL) aborts the program.

**Expected behaviour**
I expect to be able to resimulate. (Ultimately I want to pass in a different (argc, argv)/top-level generic the subsequent times - but I am not getting positive results passing in argc, argv in the first place. So that will be a second query.) 

**How to reproduce?**
Tell us how to reproduce this issue. Please provide a Minimal Working Example (MWE). With sample code it's easier to reproduce the bug and it's much faster to fix it. For example:

```vhd :file: ent.vhd
entity ent is
end entity;

architecture a of ent is
begin
  process begin
    report "Toplevel" severity note;
    wait;
  end process;
end;
```

```c :file: main.c
extern int ghdl_main(char argc, char* argv[]);

int main(int argc, char const *argv[])
{
    for (size_t i = 0; i < 2; i++)
    {
        printf("ghdl_main return: %d\n", ghdl_main(0, NULL));
    }
    return 0;
}
```

```sh :image: ghdl/ghdl:buster-llvm
gcc -c main.c
ghdl -a ent.vhd
ghdl -e -Wl,main.o bugtest
./bugtest
```

**Context**
$ ghdl --version
GHDL 0.36-dev (Ubuntu 0.35+git20181129+dfsg-4ubuntu1) [Dunoon edition]
 Compiled with GNAT Version: 8.3.0
 llvm code generator

**Additional context**
Just get the following in the terminal:
```
ent.vhd:7:16:@0ms:(report note): Toplevel
ghdl_main return: 0
./build.sh: line 5: 11322 Aborted                 (core dumped) ./bugtest
```

umarcor:
@RocketRoss, for now, GHDL cannot be reset/restarted. You need to call the binary multiple times. If you are generating an ELF binary or shared library and loading it dynamically, you need to unload it from memory and load it again. See the following related issues: #1053, #803, #1059.

In VUnit/vunit#568 there is an example that shows how this is handled from Python. Precisely, see dlopen and dlclose in https://github.com/VUnit/vunit/pull/568/files#diff-14ddabc1969914ccb081e8cd8222616d

RocketRoss:
@umarcor, Thanks for taking the time to write such an informative response. I'll run through the links etc.

By "path/to/generated/binary" d'you mean the toplevel from ```ghdl -a ... && ghdl -e ...```?

[...]

PR #1059, opened 2019-12-30
doc: add VHPIDIRECT demo

umarcor:
Ref #1053

This is work in progress to provide an example/demo of how to use VHPIDIRECT features.

Since the ABI is not settled yet, I am unsure about adding `ghdl.h` to the codebase. That's why, in this PR, it is added as an example/demo in the docs.

Types included in this demo:

- [x] Most non-composite types (`character`/`char`, `integer`/`int32_t`, `natural`/`uint32_t`, `positive`/`uint32_t`, `real`/`double`, `boolean`/`bool`, `bit`/`bool`, `time`/`int64_t`, *short* `enum`/`uint8_t`)
- [x] Record/struct
- [ ] Arrays with static bounds (note: we have found no issues with these so far)
- [ ] Unconstrained arrays
  - [x] Strings
  - [ ] With dimensions of type natural.
- [ ] Accesses to unconstrained arrays
  - [x] Strings (i.e. `line`)
  - [ ] With dimensions of type natural.
- [ ] Other accesses.
- [ ] Files

---

As expected, handling unconstrained arrays is neither straightforward nor intuitive. Hence, the main purpose of the demo is to make it easier for new users to use these. The following helper functions are provided:

- `char* ghdlToString(ghdl_NaturalDimArr_t* ptr)`: convert a fat pointer of an unconstrained string, to a (null terminated) C string
- `void ghdlToArray(ghdl_NaturalDimArr_t* ptr, void** vec, int* len, int num)`: convert a fat pointer of an uncontrained array with (up to 3) dimensions of type 'natural', to C types
- `ghdl_NaturalDimArr_t ghdlFromString(char *string)`: convert a (null terminated) C string, to a fat pointer of an unconstrained string
- `ghdl_NaturalDimArr_t ghdlFromArray(void* vec, int* len, int num)`: convert C types representing an unconstrained array with a dimension of type 'natural', to a fat pointer (**not complete yet**
- `char* ghdlAccToString(ghdl_AccNaturalDimArr_t *line)`: convert an access to an unconstrained string, to a (null terminated) C string
- `void ghdlAccToArray(ghdl_AccNaturalDimArr_t* ptr, void** vec, int* len, int num)`: convert an access to an unconstrained array with a dimension of type 'natural', to C types  (**not implemented yet**)
- `ghdl_AccNaturalDimArr_t* ghdlAccFromArray(uint32_t length, size_t bytes)`: convert C types representing an unconstrained array with a dimension of type 'natural', to an access
[...]

PR #1202, opened 2020-04-09
doc: update tips/hints related to VHPIDIRECT

umarcor:
This PR includes multiple additional tips, hints and clarifications as a result of the conversations in #1184, #1185 and #1059 with regard to building and co-executing VHDL with foreign (C) sources through VHPIDIRECT.

/cc @RocketRoss


tgingold:
Thanks!

Issue #1206, opened 2020-04-11
VHPIDIRECT: NULL argv produces 'Aborted (core dumped)'


umarcor:
While preparing some examples about VHPIDIRECT, I found that the following example (https://github.com/umarcor/ghdl-cosim/tree/todo/vhpidirect/wrapping/basic) is successfully built and executed on my host, inside a docker container. However, the exact same example, in the same container, fails when executed on GitHub Actions.

```c
#include <stdio.h>

extern int ghdl_main(int argc, void** argv);

int main(int argc, void** argv) {
  printf("ghdl_main: %d\n", ghdl_main(0, NULL));
  return 0;
}
```

The solution is as simple as NOT passing NULL as argv, but to allocate at least a position (e.g. `char** myargv`). However, I am genuinely curious:

- Why is it different on my host and on GHA, if the container (environment) is the same?
- Shouldn't argv be ignored in ghdl_main, given that argc is 0?
- Has this always been a constraint? Is it undefined behaviour to pass NULL to ghdl_main? Is there anything that GHDL can do to avoid the crash?

Depending on the response to the last point, I might need to update the HINT in https://ghdl.readthedocs.io/en/latest/using/Foreign.html#wrapping-and-starting-a-ghdl-simulation-from-a-foreign-program, since I was convinced that it was safe to use NULL.

tgingold:
I am fixing the crash.  Don't know for GHA.

umarcor:
> I am fixing the crash.

Thanks!

> Don't know for GHA.

It was actually my fault... I was using the same container, but not exactly the same version. GHA was pulling the latest, while I had an older one locally. Hence, this was indeed a regression at some point. Hopefully it is fixed now!

PR #1215, opened 2020-04-12
grt: handle argc=0 and argv/=null in ghdl_main (#1206)

umarcor:
The fix fox #1206 applied in https://github.com/ghdl/ghdl/commit/7a211109152e82843a59ba70b0a080ce91b9b495 was required because it had been broken in https://github.com/ghdl/ghdl/commit/2fc3356ae0d34dae87eb22c94f4b5eaa1873695b 10 days ago. The behaviour now is the same that it was for 15 years.

However, I think it is not completely correct. The underlaying issue is not argc being zero, but argv being null. Currently, if `argc=0` and `argv={"progname"}`, GHDL will ignore `progname`, set it to `null` and use `Default_Progname` instead. This PR changes it so that `Argv(0)` is preserved if `Argv/=null`.

BTW, I replaced:

```sh
$CC -c main.c
analyze hello.vhdl
elab -Wl,main.o hello
```

with

```sh
analyze hello.vhdl
elab -Wl,main.c hello
```

@tgingold, is there any relevant difference between these?

---

At first, I tried to test this with mcode backend, as I thought that VHPIDIRECT is supported with any backend. However, I found that the test won't work, because mcode backend does not support `-Wl,main.o` or `-Wl,main.c` as an argument. @tgingold, can you please clarify? Is it required to use command run for mcode? Should GCC be used instead?

### DIFF
--- a/src/grt/ghdl_main.adb
+++ b/src/grt/ghdl_main.adb
@@ -44,19 +44,17 @@ is
    function To_Argv_Type is new Ada.Unchecked_Conversion
      (Source => System.Address, Target => Grt.Options.Argv_Type);
 
-   My_Argv : Grt.Options.Argv_Type := To_Argv_Type (Argv);
-   Progname : Ghdl_C_String;
+   Args : Grt.Options.Argv_Type := To_Argv_Type (Argv);
+   Progname : Ghdl_C_String := null;
 begin
    --  Ada elaboration.
    Grt_Init;
 
    --  Set the options.
-   if Argc > 0 then
-      Progname := My_Argv (0);
-   else
-      Progname := null;
+   if not (Argc = 0 and Args = null) then
+     Progname := Args (0);
    end if;
-   Grt.Main.Run_Options (Progname, Argc, My_Argv);
+   Grt.Main.Run_Options (Progname, Argc, Args);
 
    --  Initialize, elaborate and simulate.
    Grt.Main.Run;
--- a/testsuite/gna/issue1206/main_notnull.c
+++ b/testsuite/gna/issue1206/main_notnull.c
@@ -1,0 +1,8 @@
+#include <stdio.h>
+
+extern int ghdl_main(int argc, char** argv);
+
+int main(int argc, char** argv) {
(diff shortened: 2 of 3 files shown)


tgingold:
Why not, but it is a little bit strange to use argc=0 and argv[0]="progname".

> BTW, I replaced:
> 
> $CC
>  -c main.c
> analyze hello.vhdl
> elab -Wl,main.o hello
> 
> with
> 
> analyze hello.vhdl
> elab -Wl,main.c hello
> 
> @tgingold, is there any relevant difference between these?

In the latter, main.c is compiled during the link by the gcc driver.  Works but harder to pass specific options.

> At first, I tried to test this with mcode backend, as I thought that VHPIDIRECT is supported with any backend. However, I found that the test won't work, because mcode backend does not support -Wl,main.o or -Wl,main.c as an argument. @tgingold, can you please clarify? Is it required to use command run for mcode? Should GCC be used instead?

VHPIDIRECT is about calling foreign C code from VHDL code.  Here we are talking about calling the simulation engine from C code, so that's the reverse.

umarcor:
> Why not, but it is a little bit strange to use argc=0 and argv[0]="progname".

It's strange, if you know what you are doing; but not so for unexperienced users. Alternatively, we can return a specific error, to let users know that they should do it properly.

> In the latter, main.c is compiled during the link by the gcc driver. Works but harder to pass specific options.

Guess it is ok for this specific case, isn't it?

> VHPIDIRECT is about calling foreign C code from VHDL code. Here we are talking about calling the simulation engine from C code, so that's the reverse.

Then, I should update the NOTE in https://umarcor.github.io/ghdl-cosim/index.html. But I'm not sure where to set the limit. All three examples in https://umarcor.github.io/ghdl-cosim/vhpidirect/examples/quickstart.html are about calling foreign C code from VHDL code. However, since `-Wl,main.c` is not supported, it seems that only the first one can be used with mcode. Is this correct?

tgingold:
Yes, only VHPIDIRECT is supported by mcode, not full cosimulation.

Issue #237, opened 2016-12-19
Array support in VPI


mark-mb:
I was using GHDL and Cocotb to run some tests for a VHDL project, and it looks like GHDL doesn't show arrays on the VPI interface.

For example, with this array:

    TYPE ram_type IS ARRAY(depth - 1 DOWNTO 0) OF STD_LOGIC_VECTOR(data_bits - 1 DOWNTO 0);
    SIGNAL ram : ram_type;

I get this error in Cocotb:

    Send raised exception: r contains no object named ram
      File "/home/markmb/vhdl/cocotb/cocotb/decorators.py", line 197, in send
        return self._coro.send(value)
      File "/home/markmb/Documents/vhdl/test/test_vhdl.py", line 44, in init_test
        mem = dut.mem.r.ram
      File "/home/markmb/vhdl/cocotb/cocotb/handle.py", line 261, in __getattr__
        raise AttributeError("%s contains no object named %s" % (self._name, name))

But I can get the value of all the other signals (std_logic and std_logic_vectors).

Is this a known issue? Is there any intention to add support for arrays in the VPI extension anytime soon?

Paebbels:
How is this related to GHDL? It's a Python error message. Cocotb has also a Python to VPI adapter layer written in C and C++. I'm just saying there are several possibilities, where it can fail.

Can you provide a small, but complete example?

mark-mb:
Sorry for the briefness of the previous message. I now include a smaller test case (below), to replicate the issue. To be clear, I am working with the code from Github for both GHDL and Cocotb

For the code below, I see these elements under `dut`: `num_line`, `data_out`, `is_write`, `data_in`, `clk`. Trying to access `mem` directly fails with the error I presented above.

Tracking Cocotb code, it seems that the call to `vpi_handle_by_name()` with parameters `main.mem` and `NULL` returns `NULL`. If I understood correctly how VPI works, this function is implemented on the GHDL side, and `NULL` means that this element doesn't exist.

VHDL code:

```vhdl
LIBRARY IEEE;
USE IEEE.STD_LOGIC_1164.ALL;

ENTITY main IS
    PORT (clk : IN STD_LOGIC;
        data_in : IN STD_LOGIC_VECTOR(31 DOWNTO 0);
        num_line : IN INTEGER RANGE 15 TO 0;
        is_write : IN STD_LOGIC;
        data_out : OUT STD_LOGIC_VECTOR(31 DOWNTO 0)
    );
END main;

[...]

PR #1249, opened 2020-04-19
WIP: Support records in vpi

benreynwar:
**Description**

The goal is to make it possible to use cocotb with a design that uses nested records and arrays in it's ports, signals and generics. Related to issue #237.

It is a long way from being finished, but it would be useful to get some feedback.  If you have the time and inclination, any comments would be appreciated.

I've added comments to the code where I have questions about how things should work.


src/grt/grt-avhpi.adb
```diff
@@ -27,6 +27,19 @@ with Grt.Vstrings; use Grt.Vstrings;
 with Grt.Rtis_Utils; use Grt.Rtis_Utils;
 with Grt.To_Strings;
 
+-- Acronyms that I don't understand:
+   -- Rti
```
tgingold, line 31: rti: run time information.  That's all the data emitted by translate to describe vhdl types and subtypes.

src/grt/grt-avhpi.adb
```diff
@@ -27,6 +27,19 @@ with Grt.Vstrings; use Grt.Vstrings;
...
+   -- Rtis
```
tgingold, line 32: Array of rti

src/grt/grt-avhpi.adb
```diff
@@ -27,6 +27,19 @@ with Grt.Vstrings; use Grt.Vstrings;
...
+   -- Rtin
```
tgingold, line 33: rti node.  All the ghdl_rtin_* records defines the rti data.

src/grt/grt-avhpi.adb
```diff
@@ -27,6 +27,19 @@ with Grt.Vstrings; use Grt.Vstrings;
...
+
+
+-- Questoins:
+   -- Explanation for the Ctxt objects would be really helpful.
+   -- Why are they used, and what do they represent?
```
tgingold, line 38: Example for a port:
you have an rti that describes the port: its name, its type, its mode...
But you also need its address to get its value.  But a port is part of an entity that can be instantiated many times.  So you need the context of the port, which consists of the entity/architecture and the address of the entity/architecture instance.

src/grt/grt-avhpi.adb
```diff
@@ -27,6 +27,19 @@ with Grt.Vstrings; use Grt.Vstrings;
...
+   -- Does ghdl support a VHPI interface?
```
tgingold, line 39: Not at the C level.  There is AVHPI which is inspired from VHPI but uses Ada style.  But there is no C VHPI interface.  Shouldn't be very hard to do, but there are very few examples.

src/grt/grt-avhpi.adb
```diff
@@ -27,6 +27,19 @@ with Grt.Vstrings; use Grt.Vstrings;
...
+   -- What's the best way to test this work.  Should we add
+   --   a cocotb dependency to the testing?
```
tgingold, line 41: I really prefer to have standalone tests.  That's easier to investigate and do not depends on cocotb evolution.

src/grt/grt-avhpi.adb
```diff
@@ -126,14 +139,18 @@ package body Grt.Avhpi is
             end case;
          when VhpiIndexedNames =>
             case Ref.Kind is
-               when VhpiGenericDeclK =>
+               when VhpiGenericDeclK |
+                    VhpiPortDeclK |
+                    VhpiSigDeclK =>
```
tgingold, line 144: Looks ok.

src/grt/grt-avhpi.adb
```diff
@@ -126,14 +139,18 @@ package body Grt.Avhpi is
...
                   Res := (Kind => AvhpiNameIteratorK,
                           Ctxt => Ref.Ctxt,
+                          -- What does the 'N_' indicate?
```
tgingold, line 147: It's the prefix for names in the vhpiHandleT.  All the names in a record must be unique.

src/grt/grt-avhpi.adb
```diff
@@ -145,6 +162,8 @@ package body Grt.Avhpi is
                   return;
             end case;
             case Res.N_Type.Kind is
+               -- Why only Ghdl_Rtik_Subtype_Array?
+               -- Why not also Ghdl_Rtik_Type_Array?
```
tgingold, line 166: Yes, to be added ?

src/grt/grt-avhpi.adb
```diff
@@ -158,12 +177,54 @@ package body Grt.Avhpi is
...
+                          N_Addr => Ref.N_Addr,
+                          N_Type => Ref.N_Type,
+                          N_Idx => 0,
+                          N_Obj => Ref.N_Obj);
+               when others =>
+                  Error := AvhpiErrorNotImplemented;
+                  return;
+            end case;
+            case Res.N_Type.Kind is
+               when Ghdl_Rtik_Type_Record =>
```
tgingold, line 213: and also record subtypes ? :-)

src/grt/grt-avhpi.adb
```diff
@@ -184,6 +254,9 @@ package body Grt.Avhpi is
       El_Type1 : Ghdl_Rti_Access;
    begin
       case Obj_Rti.Common.Kind is
+         -- Does this mean we only support scanning iterators in
+         -- generics.  That doesn't make much sense.
+         -- What am I missing?
```
tgingold, line 259: Possibly.  In VPI you can directly get the value of a vector.

src/grt/grt-avhpi.adb
```diff
@@ -199,6 +272,12 @@ package body Grt.Avhpi is
       case El_Type1.Kind is
          when Ghdl_Rtik_Type_P64 =>
             if Is_Sig then
+               -- This is currently never run, but why would
+               -- this be different for generics and signals?
+               -- Maybe signals are built as structures of signals,
+               -- but generics are raw?  So signals effectively
+               -- contain lots of thin signal wrappers that we're
+               -- pointing at.
```
tgingold, line 280: Yes, generics contain the raw value, while signals are addresses to an internal structure.

src/grt/grt-avhpi.adb
```diff
@@ -207,6 +286,10 @@ package body Grt.Avhpi is
             declare
                Sizes : Ghdl_Indexes_Ptr;
             begin
+               -- An explanation of what's going on here would be really
+               -- helpful.
+               -- It looks like we're finding out the size of the structure
+               -- but I've failed to understand how.
```
tgingold, line 292: Ah OK.  So that's the for_generate_statement.  There is an array containing the sub-instances, because the block can be repeated many times.  So we need to update the context to point to the corresponding sub-instance.

src/grt/grt-avhpi.adb
```diff
@@ -219,11 +302,17 @@ package body Grt.Avhpi is
                end if;
             end;
          when others =>
+            -- Feels like we supported a strange subset of possible types.
+            -- I'm curious what drove the supported features in this
+            -- implementation.
```
tgingold, line 307: I think it is mainly SDF annotations + what has been asked for VPI.

src/grt/grt-avhpi.adb
```diff
@@ -249,10 +339,46 @@ package body Grt.Avhpi is
...
+              N_Obj => Iterator.N_Obj);
+
+      --  Increment Address.
+      Iterator.N_Addr := Add_Index
+        (Iterator.Ctxt, Iterator.N_Addr, Iterator.N_Obj, El_Type, 1);
+
+      -- Decrement number of elements left in iterator.
+      Iterator.N_Idx := Iterator.N_Idx - 1;
+      Error := AvhpiErrorOk;
+   end Vhpi_Scan_Selected_Name;
```
tgingold, line 379: That looks OK, but I haven't thoroughly reviewed this subprogram.

src/grt/grt-avhpi.adb
```diff
@@ -398,6 +526,7 @@ package body Grt.Avhpi is
                Bt : constant Ghdl_Rtin_Type_Array_Acc :=
                  To_Ghdl_Rtin_Type_Array_Acc (Atype.Basetype);
             begin
+               -- I don't understand what's happening here.
```
tgingold, line 529: Ah.  An array type declaration like `type my_vec is array (0 to 7) of bit` is represented as an anonymous unconstrained array type declaration `type anon_my_vec is array (integer range <>) of bit` followed by a subtype declaration `subtype my_vec is anon_my_vec(0 to 7)`.  So we need to ignore the anonymous array type declaration and pretend that the subtype declaration is a type declaration.

src/grt/grt-vpi.adb
```diff
@@ -68,7 +69,7 @@ package body Grt.Vpi is
    Version : constant String := "0.1" & NUL;
 
    --  If true, emit traces
-   Flag_Trace : Boolean := False;
+   Flag_Trace : Boolean := True;
```
tgingold, line 72: Use `--vpi-trace` to dynamically set this flag.

benreynwar:
Awesome,  thanks a bunch for the help!

tgingold:
If you need more comments, do not hesitate.

Maybe we can start to merge some parts.  For example, if it is already possible to scan through generics of a record type, this specific part could be tested and added.

benreynwar:
I'm trying to understand the RTI structures but am having difficulty understanding what data is stored in the memory that the `Loc` fields point to.  My understanding is that the `Loc` fields tell us the location of relevant data in the memory relative to the address specified by `Rti_Context.Base`, but I'm not sure what structure exactly is stored in this location, and how the stored data is different for the different RTI nodes.

tgingold:
Note that there is the option '--dump-rti' to dump the rtis at runtime: ghdl -r design --dump-rti
That could be useful to understand how rtis work.  This is implemented by grt-disp_rti, which is almost standalone.

1) VHDL types are represented as close as possible to the corresponding C type.  So a scalar is represented as a scalar, an array as an array and a record as a struct.

2) For signals, the scalars and the scalar sub-elements are represented as pointers.  So a bit signal is represented by a pointer, a std_logic_vector by an array of pointer, and a record by a struct of pointers.  The pointer points to ghdl_signal record defined in grt-signals.  The record contains the current value, but also the value for 'event, 'active, 'last_value, 'last_event...

3) for records, there is one RTI for each element, which gives the offset from the record address to the field address. But grt.rtis_utils has many subprograms to handle the rtis, including Record_To_Element.

benreynwar:
Looking at grt-disp_rti is helping.  I don't have any additional questions, and there's nothing that's worth reviewing yet.  I'm just going through grt-disp_rti and making sure I understand what's going on.

benreynwar:
When a new array type is defined that binds an unbound type an anonymous RTI type object is created:

```vhd
type mytype is array(natural range <>) of std_logic_vector(WIDTH-1 downto 0);
```

However when a new record type is defined that binds an unbound type an anonymous type is not created.  Rather the information about the bounds is stored in the element node of the record.

```vhd
type mytype is record
  a: std_logic_vector(WIDTH-1 downto 0);
end;
```

This is a bit of a pain when dealing with the subcomponents of the signal, since there is no RTI type that can be easily referred to.  Instead we have to refer to the unbound type, and the bounds information.  It would be much more pleasant if records created anonymous types in the same way that arrays do.  Is this a change that could be limited to the RTI part of the codebase or would it be more complicated?

benreynwar:
Also, the RTI tree has `Ghdl_Rti_Object` to represent signal/port/constant/generic but doesn't contain any more fine-grained objects to represent the subcomponents.  Is this intentional so that we don't get a crazy number of objects that might effect peformance, so was it just that there was no advantage to it?  Having more fine-grained objects could simplify things when dealing with records and arrays.

tgingold:
> Also, the RTI tree has Ghdl_Rti_Object to represent signal/port/constant/generic but doesn't contain any more fine-grained objects to represent the subcomponents. Is this intentional so that we don't get a crazy number of objects that might effect peformance, so was it just that there was no advantage to it? Having more fine-grained objects could simplify things when dealing with records and arrays.

That's not possible.  The size of arrays is not known when the RTIs are generated.

tgingold:
> type mytype is array(natural range <>) of std_logic_vector(WIDTH-1 downto 0);
> However when a new record type is defined that binds an unbound type an anonymous type is not created. Rather the information about the bounds is stored in the element node of the record.

I am not sure to understand what you means:

```
entity repro1 is
  generic (WIDTH : natural := 4);
end;

architecture behav of repro1 is
  type myarr is array(natural range <>) of bit_vector(WIDTH-1 downto 0);

  type myrec is record
    a: bit_vector(WIDTH-1 downto 0);
  end record;
begin
end;
```

`ghdl -r repro1 --dump-rti`:
...
```
ghdl_rtik_architecture, D=1, sloc=5:14: behav
 filename: repro1.vhdl
 ghdl_rtik_entity, D=1, sloc=1:8: repro1
  filename: repro1.vhdl
  ghdl_rtik_generic, D=1, sloc=2:12; width: natural := 4
 ghdl_rtik_type_array: myarr is array (natural range <>) of bit_vector (3 downto 0)
 ghdl_rtik_type_record: myrec is record
  ghdl_rtik_element: a: bit_vector (3 downto 0)
```

benreynwar:
My understanding is that an anonymous RTI node with kind Ghdl_Rtik_Subtype_Array is created.  The Ghdl_Rtik_Type_Array with name `myarr` then references this anonymous subtype as it's `Element`.  

For the `myrec` type no anonymous subtype array is created, rather the Ghdl_Rtik_Type_Array is used in conjunction with the bounds stored in the layout associated with the Ghdl_Rtin_Element.

The anonymous types don't show in the disp_rti dump because they are not children of any of the blocks.

benreynwar:
> Also, the RTI tree has Ghdl_Rti_Object to represent signal/port/constant/generic but doesn't contain any more fine-grained objects to represent the subcomponents. Is this intentional so that we don't get a crazy number of objects that might effect peformance, so was it just that there was no advantage to it? Having more fine-grained objects could simplify things when dealing with records and arrays.
> That's not possible. The size of arrays is not known when the RTIs are generated.
Makes sense.

benreynwar:
I like the anonymous types that are created by arrays because it makes it easy for a subcomponent of an object to directly point to a type that defines it.

Currently the subcomponent of a signal of type `myrec` cannot simply point to a type that defines it.  It must point to:
1) The unbounded type
2) The RTI node of the Record (which we can use with the context to find the location of the record layout)
3) The RTI node of the Element (from which we can find the offset in the record layout where the bounds are located).

benreynwar:
> >Also, the RTI tree has Ghdl_Rti_Object to represent signal/port/constant/generic but doesn't contain any more fine-grained objects to represent the subcomponents. Is this intentional so that we don't get a crazy number of objects that might effect peformance, so was it just that there was no advantage to it? Having more fine-grained objects could simplify things when dealing with records and arrays.

> That's not possible. The size of arrays is not known when the RTIs are generated.

Thinking about this a bit more.  All the elements in an array could use the same object RTI node.  When accessing data associated with the object, you'd use a context that specified the location within the array.   It would work like using the context to specify position in a for/generate loop.

tgingold:
> I like the anonymous types that are created by arrays because it makes it easy for a subcomponent of an object to directly point to a type that defines it.

Ok, but anyway that's not possible with vhdl08 unbounded arrays: the bounds are then defined by the object, not by the subtype.
So such a change won't fly very far.

(And Ok, vhdl08 is not yet well supported by rtis, but that's a different point).

> Currently the subcomponent of a signal of type myrec cannot simply point to a type that defines it. It must point to:
> 
> 	• The unbounded type
> 	• The RTI node of the Record (which we can use with the context to find the location of the record layout)
> 	• The RTI node of the Element (from which we can find the offset in the record layout where the bounds are located).

Ideally, you always need:
1) the address of the object or sub-element.  In your case, you need simply the address of the element, you don't need the address of the whole record.
2) Whether the object is a signal or not.
3) The RTI for the type
4) Possibly the address of the bounds for the type
5) The context of the RTI (in case you need to reference another RTI from it).

Ok, that's not simple but I fear this is required by vhdl08, and the fact that RTIs are static.

tgingold:
> Thinking about this a bit more. All the elements in an array could use the same object RTI node. When accessing data associated with the object, you'd use a context that specified the location within the array. It would work like using the context to specify position in a for/generate loop.

So you'd like to have one RTI for the whole array object and one for an element of the array object.  Right ?

If so, my answer would be:
you could dynamically create your own RTI for that,
or you can use the vpiHandle structure for that too.

benreynwar:
> Thinking about this a bit more. All the elements in an array could use the same object RTI node. When accessing data associated with the object, you'd use a context that specified the location within the array. It would work like using the context to specify position in a for/generate loop.
> So you'd like to have one RTI for the whole array object and one for an element of the array object. Right ? If so, my answer would be: you could dynamically create your own RTI for that, or you can use the vpiHandle structure for that too.

Yep. Dynamically creating these objects as needed would work fine.  I'd rather keep the logic for navigating the RTI as separate as possible from the vpi/avhpi logic just to minimize my own confusion.

benreynwar:
> I like the anonymous types that are created by arrays because it makes it easy for a subcomponent of an object to directly point to a type that defines it.
> Ok, but anyway that's not possible with vhdl08 unbounded arrays: the bounds are then defined by the object, not by the subtype. So such a change won't fly very far. (And Ok, vhdl08 is not yet well supported by rtis, but that's a different point).

I still don't understand why this couldn't be implemented as an anonymous subtype.  Defining the bounds in the object is equivalent to defining the bounds in an anonymous subtype and then pointing the object to that subtype.  I don't use much of vhdl08 so it's possible I'm missing something.  I have a book on vhdl08 arriving in a couple of weeks :).

tgingold:
> I still don't understand why this couldn't be implemented as an anonymous subtype. Defining the bounds in the object is equivalent to defining the bounds in an anonymous subtype and then pointing the object to that subtype. I don't use much of vhdl08 so it's possible I'm missing something. I have a book on vhdl08 arriving in a couple of weeks :).

No, because the bounds may belong to the object and not to the subtype.

For example:

`type string_acc is access string;`

`a := new string'("Hello");`
`b := new string'("my world");`

So `a` and `b` have two different bounds, and the number of such objects is unknown.  So if we needed to always have bounds in an RTI we would have to dynamically create RTIs.

benreynwar:
> > I still don't understand why this couldn't be implemented as an anonymous subtype. Defining the bounds in the object is equivalent to defining the bounds in an anonymous subtype and then pointing the object to that subtype. I don't use much of vhdl08 so it's possible I'm missing something. I have a book on vhdl08 arriving in a couple of weeks :).

> No, because the bounds may belong to the object and not to the subtype. For example: `type string_acc is access string;` `a := new string'("Hello");` `b := new string'("my world");` So `a` and `b` have two different bounds, and the number of such objects is unknown. So if we needed to always have bounds in an RTI we would have to dynamically create RTIs.

Still trying to get my head around this.  In the example you gave we'll be creating RTIs for `a` and `b`.  Why wouldn't it be practical to create an anonymous subtype for each at the same time?

tgingold:
> Still trying to get my head around this. In the example you gave we'll be creating RTIs for a and b. Why wouldn't it be practical to create an anonymous subtype for each at the same time?

And to store the bounds within this RTI ? The consequence is that you will have to modify the RTIs each time to assign such a variable, because you could change the bounds.  But RTIs are read-only.

Or having RTIs that say that the bounds are within the object ?  I fear that would be much more complex or not very different from the current implementation.

benreynwar:
> Still trying to get my head around this. In the example you gave we'll be creating RTIs for a and b. Why wouldn't it be practical to create an anonymous subtype for each at the same time?

> And to store the bounds within this RTI ? The consequence is that you will have to modify the RTIs each time to assign such a variable, because you could change the bounds. But RTIs are read-only. Or having RTIs that say that the bounds are within the object ? I fear that would be much more complex or not very different from the current implementation.

I understand.   The bounds are determined after elaboration and could change during runtime.  I have to try to remember about non-synthesizable vhdl too :).

tgingold:
Yes.  On top of that, subprograms parameters are often of unbounded types.

benreynwar:
I'm continuing to very slowly work on this.  If anyone is impatient and would like to have a crack themselves, I won't be at all offended if you take over :).

benreynwar:
I don't think this ever got to the point where it was worth merging.  I think it's very unlikely that it wouldn't be much easier for @tgingold to start from scratch if he ever wanted to implement this.

Issue #1278, opened 2020-05-03
The path of VPI modules is not valid on MINGW


umarcor:
The following MWE, extracted from ghdl/ghdl-cosim#21, works on Linux, but not on Windows (MSYS2, v0.37). It seems to be an issue with the format of the path/location passed through `--vpi`.

```c :file: vpi.c
#include <vpi_user.h>
#include <stdio.h>

PLI_INT32 cb_hello(){
  printf("Hello!\n");
  return 0;
}

void entry_point_cb() {
  s_cb_data cb;

  cb.reason = cbStartOfSimulation;
  cb.cb_rtn = &cb_hello;
  cb.user_data = NULL;

  if (vpi_register_cb(&cb) == NULL) {
    vpi_printf ("cannot register cbStartOfSimulation call back\n");
  }
}

// List of entry points called when the plugin is loaded
void (*vlog_startup_routines[]) () = {entry_point_cb, 0};
```

```vhdl :file: tb.vhd
entity tb is
end entity;

architecture arch of tb is
begin
  process
  begin
    report "Hello!" severity note;
    wait;
  end process;
end architecture;
```

```sh :file: run.sh
#!/usr/bin/env sh

set -e

cd $(dirname "$0")

ghdl -a tb.vhd
ghdl -e -o tb tb
ghdl --vpi-compile gcc -c vpi.c -o vpi.o
ghdl --vpi-link gcc vpi.o -o vpi.vpi

set +e

export PATH=$PATH:$(pwd)

echo "> Execute tb"

./tb --vpi=vpi.vpi
./tb --vpi=./vpi.vpi
./tb --vpi="$(pwd)/vpi.vpi"
./tb --vpi="/$(pwd)/vpi.vpi"
./tb --vpi="$(echo $(pwd)/vpi.vpi | sed 's#/#\\#g')"
./tb --vpi="$(echo /$(pwd)/vpi.vpi | sed 's#/#\\#g')"
```

```
loading VPI module 'vpi.vpi'
The specified module could not be found.

D:\ghdl-cosim\vpi\quickstart\debug\tb.exe:error: cannot load VPI module  
loading VPI module './vpi.vpi'
The specified module could not be found.

D:\ghdl-cosim\vpi\quickstart\debug\tb.exe:error: cannot load VPI module  
loading VPI module 'D:/ghdl-cosim/vpi/quickstart/debug/vpi.vpi'
The specified module could not be found.

[...]

PR #1279, opened 2020-05-03
Document and use --vpi-library-dir-unix

umarcor:
As discussed in #1278, the current `sed` command in `testenv.sh` does not work on GHA Windows jobs, because the drive is not `C:`. This PR changes `--vpi-library-dir` to `--vpi-library-dir-unix`. At the same time, the workaround that existed in `dist/msys2-mingw/run.sh` is removed.

By the way, `--vpi-library-dir-unix` is added to the docs, and some redundant headers are removed.

### DIFF
--- a/dist/msys2-mingw/run.sh
+++ b/dist/msys2-mingw/run.sh
@@ -109,7 +109,6 @@ test () {
     env | grep MINGW
   gend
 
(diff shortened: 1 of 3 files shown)


tgingold:
Thanks!

Issue #1288, opened 2020-05-07
PSL: psl.sem_property: cannot handle N_NEXT_E


tmeissner:
First real issue resulting from my work at the [psl_with_ghdl](https://github.com/tmeissner/psl_with_ghdl) project :)

**Description**
The PSL `next_e` operator isn't supported yet. If you use `next_e` in your code, you get an error like this:

```
psl.sem_property: cannot handle N_NEXT_E
```

**Expected behaviour**
The `next_e` PSL operator should be supported and work as defined in the IEEE standard for PSL. Taken from the 2010 version, it is mentioned in section 6.2.1.4.2:

> The next_e family of operators, shown in Syntax 6-24, specify that an FL Property holds at least once within some range of future cycles.

**How to reproduce?**

```vhd :file: sequencer.vhd
library ieee;
  use ieee.std_logic_1164.all;


entity sequencer is
  generic (
    seq : string
  );
  port (
    clk  : in  std_logic;
    data : out std_logic
  );
end entity sequencer;


architecture rtl of sequencer is

  signal index : natural := seq'low;
  signal ch    : character;

  function to_bit (a : in character) return std_logic is
    variable ret : std_logic;
  begin
    case a is
      when '0' | '_' => ret := '0';
      when '1' | '-' => ret := '1';
      when others    => ret := 'X';
    end case;
    return ret;
  end function to_bit;

begin


  process (clk) is
  begin
    if rising_edge(clk) then
      if (index < seq'high) then
        index <= index + 1;
      end if;
    end if;
  end process;

  ch <= seq(index);

  data <= to_bit(ch);


end architecture rtl;
```

```vhd :file: psl_next_e.vhd
library ieee;
  use ieee.std_logic_1164.all;


entity psl_next_e is
  port (
    clk : in std_logic
  );
end entity psl_next_e;


architecture psl of psl_next_e is

  component sequencer is
    generic (
      seq : string
    );
    port (
      clk  : in  std_logic;
      data : out std_logic
    );
  end component sequencer;

  signal a, b : std_logic;
  signal c, d : std_logic;
  signal e, f : std_logic;

begin

  --                              01234567890
  SEQ_A : sequencer generic map ("__-_-______") port map (clk, a);
  SEQ_B : sequencer generic map ("_____-_-___") port map (clk, b);

  --                              01234567890
  SEQ_C : sequencer generic map ("__-_-______") port map (clk, c);
[...]

PR #1289, opened 2020-05-08
Fix for PSL next_e operator

tmeissner:
This small PR fixes the handling of the PSL operator `next_e` as reported in bug #1288. I also added tests for simulation & synth. The CI seems to be happy, so have a look and please merge :)


tgingold:
Thanks!

PR #1297, opened 2020-05-10
Constants in vpi

benreynwar:
This allows accessing constants through VPI.
It also fixes an issue that prevented accessing bit, std_logic or boolean generics.

I wasn't sure where to put the tests.  Should I open issues and then create corresponding directories in the testsuite/gna directory?  For now I made a testsuite/vpi directory and put them in there.  Currently the tests are just checking that we don't get an error when we try to get the value.  They don't actually check the value.

Also, it appears in grt-avhpi.adb that put_value is supported for  generics.  Do we want this?

Would you prefer for me to squash pull requests like this myself, or will you squash it yourself when merging?

### DIFF
--- a/src/grt/grt-vpi.adb
+++ b/src/grt/grt-vpi.adb
@@ -907,13 +907,15 @@ package body Grt.Vpi is
                Append_Bin (Ghdl_U64 (V), 32);
             end;
          when Vcd_Bit
-           | Vcd_Bool
-           | Vcd_Bitvector =>
+           | Vcd_Bool =>
+            Append (Buf_Value, Map_Std_B1 (Verilog_Wire_Val (Info).B1));
+         when Vcd_Bitvector =>
             for J in 0 .. Len - 1 loop
                Append (Buf_Value, Map_Std_B1 (Verilog_Wire_Val (Info, J).B1));
             end loop;
-         when Vcd_Stdlogic
-           | Vcd_Stdlogic_Vector =>
+         when Vcd_Stdlogic =>
+            Append (Buf_Value, Map_Std_B1 (Verilog_Wire_Val (Info).B1));
+         when Vcd_Stdlogic_Vector =>
             for J in 0 .. Len - 1 loop
                Append (Buf_Value, E8_To_Char (Verilog_Wire_Val (Info, J).E8));
             end loop;
--- a/src/grt/grt-vpi.adb
+++ b/src/grt/grt-vpi.adb
@@ -574,6 +574,15 @@ package body Grt.Vpi is
                   return vpiParameter;
                end if;
             end;
+         when VhpiConstDeclK =>
+            declare
+               Info : Verilog_Wire_Info;
+            begin
+               Get_Verilog_Wire (Res, Info);
+               if Info.Vtype /= Vcd_Bad then
+                  return vpiConstant;
+               end if;
+            end;
          when others =>
             null;
       end case;
@@ -595,6 +604,9 @@ package body Grt.Vpi is
                                          Ref => Res);
          when vpiParameter =>
             return new struct_vpiHandle'(mType => vpiParameter,
+                                         Ref => Res);
+         when vpiConstant =>
+            return new struct_vpiHandle'(mType => vpiConstant,
                                          Ref => Res);
          when others =>
             return null;
@@ -859,7 +871,8 @@ package body Grt.Vpi is
       case Vhpi_Get_Kind (Obj) is
          when VhpiPortDeclK
            | VhpiSigDeclK
-           | VhpiGenericDeclK =>
+           | VhpiGenericDeclK
+           | VhpiConstDeclK =>
             null;
          when others =>
             return null;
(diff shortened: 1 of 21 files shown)


tgingold:
> This allows accessing constants through VPI.

That's nice.

You added VhpiConstantDeclK in grt-avhpi.ads, but it duplicates the existing VhpiConstDeclK.  Can you fix that ?

> It also fixes an issue that prevented accessing bit, std_logic or boolean generics.

I am not sure why this has been fixed or why it was not working.

> I wasn't sure where to put the tests. Should I open issues and then create corresponding directories in the testsuite/gna directory? For now I made a testsuite/vpi directory and put them in there. Currently the tests are just checking that we don't get an error when we try to get the value. They don't actually check the value.

We can indeed create a new vpi/ directory or put them in the gna/ directory.

> Also, it appears in grt-avhpi.adb that put_value is supported for generics. Do we want this?

Yes, we need that to support SDF annotations.

> Would you prefer for me to squash pull requests like this myself, or will you squash it yourself when merging?

I can squash myself.

Thanks!

src/grt/grt-vpi.adb
```diff
@@ -894,13 +907,15 @@ package body Grt.Vpi is
...
+            Append (Buf_Value, Map_Std_B1 (Verilog_Wire_Val (Info).B1));
+         when Vcd_Bitvector =>
             for J in 0 .. Len - 1 loop
                Append (Buf_Value, Map_Std_B1 (Verilog_Wire_Val (Info, J).B1));
             end loop;
-         when Vcd_Stdlogic
-           | Vcd_Stdlogic_Vector =>
+         when Vcd_Stdlogic =>
+            Append (Buf_Value, E8_To_Char (Verilog_Wire_Val (Info).E8));
+         when Vcd_Stdlogic_Vector =>
```
benreynwar, line 918: These are the changes that fixed accessing the values for generics for std_logic, generic and boolean.  The problem is that Verilog_Wire_Val is not implemented by constant/generic arrays.  Previously boolean, std_logic and bit were using the array version.  I switched to using the non-array one where constant/generic are supported.

benreynwar:
> That's nice. You added VhpiConstantDeclK in grt-avhpi.ads, but it duplicates the existing VhpiConstDeclK. Can you fix that ?

Done

> We can indeed create a new vpi/ directory or put them in the gna/ directory.

Ok.  I'll leave them in the vpi directory.  That will make things a bit simpler for me to keep track of the vpi tests I'm adding.

eine:
> > We can indeed create a new vpi/ directory or put them in the gna/ directory.
> 
> Ok. I'll leave them in the vpi directory. That will make things a bit simpler for me to keep track of the vpi tests I'm adding.

This requires some additional plumbing...

- The new test suite needs to be explicitly added in https://github.com/ghdl/ghdl/blob/master/dist/ci-run.sh#L432 and https://github.com/ghdl/ghdl/blob/master/dist/ci-run.sh#L437
- Existing test might be moved from `gna` to `vpi`:

```
# grep -l -r ./testsuite/ -e "vpi-compile"
./testsuite/gna/issue1226/testsuite.sh
./testsuite/gna/issue1228/testsuite.sh
./testsuite/gna/issue1233/testsuite.sh
./testsuite/gna/issue1256/testsuite.sh
./testsuite/gna/issue450/testsuite.sh
./testsuite/gna/issue531/testsuite.sh
./testsuite/gna/issue98/testsuite.sh
```

BTW, I think it would be good to use the same naming (issue/PR number).

tgingold:
Thanks!

tgingold:
> BTW, I think it would be good to use the same naming (issue/PR number).

For sure, tests from issues must have the issue in the name.

But it is also possible to add tests for specific features using the feature name (like in synth).

eine:
The point is, existing and future tests from issues that correspond to `vpi` do need to be located in subdir `vpi`? Or will there be tests about `vpi` in `gna` too?

tgingold:
I think it is better if all issues are put in gna/, so that it is easier to find them.

benreynwar:
Another option is to move the vpiXXX tests into the gna folder.  These tests would correspond to broad tests that are used to check what features are currently supported, and to help keep track of what needs to be done.  As features are added, lines in these tests would be uncommented so that they are testing these features.

benreynwar:
Maybe once a vpiXXX tests has all lines uncommented (i.e. all the things it's testing supported) I could create an issue, close it, and then rename vpiXXX to issueXXX?  Alternately I could create the issues up front and use that name to start off with.

eine:
> Another option is to move the vpiXXX tests into the gna folder. These tests would correspond to broad tests that are used to check what features are currently supported, and to help keep track of what needs to be done. As features are added, lines in these tests would be uncommented so that they are testing these features.

I had not understood that the purpose was to keep updating these same tests as features are added. Then, I think it is ok as it is now, because possible duplication with content in `gna` is justified. Sorry for bothering you.

benreynwar:
@eine You didn't bother to me.  It's nice to have multiple people checking my work when I'm working on a project I'm not familiar with :).

Issue #1053, opened 2019-12-07
Enhancements to VHPIDIRECT: C APIs


umarcor:
Coming from https://github.com/ghdl/ghdl/issues/1052#issuecomment-562704835

The purpose of this issue is to track possible future enhancements to GHDL regarding the features for co-simulation through VHPIDIRECT.

- GHDLCInterface: runtime/execution management API for GHDL.
- GHDLPointers: C API to decode/encode GHDL's fat pointers from/to C types.

---

> [@umarcor](https://github.com/ghdl/ghdl/issues/800#issuecomment-487084044):
>I can imagine an API such as:
> ```
> ghdl_main();
> ghdl_runfor(10);
> ghdl_restart();
> ghdl_rununtil(100);
> ghdl_destroy();
> ```
> Is this something I might try to implement or is some deep understanding of GHDL internals required?

> [@tgingold](https://github.com/ghdl/ghdl/issues/800#issuecomment-487257376)
> No, I don't think it require a deep understanding.  This is about run time, so you only need to refer to the code of grt/

umarcor:
@tgingold, it feels that the addition of `Simulation_Step` in https://github.com/ghdl/ghdl/commit/2fc3356ae0d34dae87eb22c94f4b5eaa1873695b, along with exporting it and `__ghdl_simulation_init`, is a step towards supporting `ghdl_rununtil` and `ghdl_runfor`. Although the comment of the commit says it is for SystemC, I want to believe that there is little difference between VHPIDIRECT and cosimulation with SystemC. That is, the former is a subset of the later.

Taking https://github.com/ghdl/ghdl-systemc-fosdem16/blob/master/counter_vhdl.cpp as a reference, does the following make sense as a drop-in replacement for `ghdl_main`?

```c
int my_ghdl_main(int argc, void** argv) {
  grt_init (); // Elab Ada code
  grt_main_options (argv[0], argc, argv);
  grt_main_elab ();
  __ghdl_simulation_init ();
  int ecode;
  do {
    ecode = __ghdl_simulation_step ();
  } while (ecode!=3);
  return ecode;
}
```

umarcor:
Replying to myself: yes, it work, but not as expected. See ghdl/ghdl-cosim#14.

umarcor:
The problem in ghdl/ghdl-cosim#14 turned out to be the usage of a too simple testbench. That PR is now merged: the example is available in ghdl-cosim and referenced in the docs.

Issue #101, opened 2016-06-28
Co-Simulation with CLaSH


jgjverheij:
I am trying to create co-simulation between CLaSH (https://github.com/clash-lang/clash-compiler) and other simulators. With GHDL I want to use the VPI standard, but some (needed) things are not supported (yet)...

I am not familiar with the GHDL source-code and I want to leave this as suggestions for future implementations. 

1. I register the callback "cbStartOfSimulation" in which I want to iterate over the "vpiPorts". This is needed to check if the number and sizes of the input- and output-ports correspond to the "signals" within CLaSH. 
For this I want to use "vpiPort" instead of "vpiNet", to also use "vpiDirection". With the "vpiDirection" i can distinguish between input- and output-signals. The properties "vpiPort" and "vpiDirection" are unknown in GHDL. 

2. The width of the "signals" within CLaSH are user-defined and not restricted to 32 bits. To support more than 32 bits, I use the "vpiVectorVal" property in the "vpi_get_value" and "vpi_put_value" functions. The property "vpiVectorVal" is currently not supported in GHDL. 
A workaround is to use the "vpiBinStrVal" property, but then I have to work with string values...

tgingold:
I confirm that vpiPorts, vpiDirection and vpiVectorVal are not yet supported.

1138-4EB:
Was this issue fixed in #753? Or is `vpiVectorVal` unsupported yet?

tgingold:
I think it is not supported yet.

PR #1299, opened 2020-05-10
WIP: Adding structure to represent a give node/context point in RTI tree.

benreynwar:
Currently the RTI codebase is somewhat difficult to work on because there are lots of factors (RTI node structure, context, bounds location, whether it's in a signal or generic) that are needed to properly describe a point in the RTI tree.

In this pull request I added an RTII (Run Time Instance Information) structure that gathers this data into a single structure.  These structures will be generated dynamically as required. Avhpi handles for signals/ports/generics/constants and their subcomponents will contain one of these RTII structures.

My hope is that this will make the codebase simpler to understand, and easier to implement new features.

This is not a branch that should be merged, I've just put in this pull request to get some feedback on the idea of creating this structure, and to get suggestions for alternative ways of approaching this.

My current plan is:
1) Use RTII to support std_logic_vector and bit_vector in generics and constants.
2) Support array of bit_vector, array of integer, multi-dimensional arrays in signal/port/generic/constants
3) Support records in signal/port/generic/constant
4) Support nested records and arrays.


tgingold:
I think this is a good idea.

I will do a review.

You can already submit your documentation changes.  Making many smaller PR is always easier to handle than a big one.

src/grt/grt-rtiis.ads
```diff
@@ -0,0 +1,104 @@
...
+with Grt.Rtis; use Grt.Rtis;
+with Grt.Rtis_Addr; use Grt.Rtis_Addr;
+
+package Grt.Rtiis is
+
+   -- FIXME: Delete this.
+   -- Just in here so we have a constant that we can harmlessly include
+   -- include in another package to force compilation.
+   FISH : constant Natural := 2;
+
```
tgingold, line 36: You don't need that.  You can always do:
```
with Grt.Rtiis;
pragma Unreferenced (Grt.Rtiis);
```

src/grt/grt-rtiis.ads
```diff
@@ -0,0 +1,104 @@
...
+      -- The RTI of the base type (this type or the unbounded version of it)
+      Rti : Ghdl_Rti_Access;
+      -- The context is used to find the layouts and bound associated with
+      -- types.
+      Ctxt : Rti_Context;
+      -- This is the address of the layout information. It can be determined
+      -- from the Rti and the Ctxt, but we also store it here so we don't have
+      -- to determine it too often.
+      Layout_Addr : Address;
+      -- If the Base_Rti is unbound then this is the address of the
```
tgingold, line 47: Maybe create: `subtype Layout_Address is Address;` to clarify the use of the address ?

src/grt/grt-rtiis.ads
```diff
@@ -0,0 +1,104 @@
...
+      -- 0 is the first dimension.
+      Dim : Ghdl_Index_Type;
+   end record;
+
+   -- A RTII (Run time instance information) represents a node in the
+   --  design tree. Where an RTI node depends on the context and could
+   --  be instatiated many times in the design, an RTII node
+   --  represents one specfic instance of that node.
+   type Ghdl_Rtii is record
+      -- Whether it is a signal/port or a generic/constant.
```
tgingold, line 60: I fear it is confusing to have both `Ghdl_Rtii` and `Ghdl_Rtii_Type`.  What about `Ghdl_Object_Rtii` and `Ghdl_Type_Rtii` ?

src/grt/grt-rtiis.ads
```diff
@@ -0,0 +1,104 @@
...
+      -- The address of the data for this component.
+      Addr : Address;
+   end record;
+
+   function To_Ghdl_Rtii(Rti : Ghdl_Rtin_Object_Acc; Ctxt : Rti_Context)
+                        return Ghdl_Rtii;
+
+   -- Get a indexed child of an Rtii.
+   function Get_Rtii_Child(Rtii : Ghdl_Rtii; Index : Ghdl_Index_Type)
+   return Ghdl_Rtii;
```
tgingold, line 73: There are some style differences: space before '(', return aligned, 2 spaces after '--'.

src/grt/grt-rtiis.adb
```diff
@@ -0,0 +1,350 @@
...
+   begin
+      Kind := Rtii.Typ.Rti.Kind;
+      case Kind is
+         when Ghdl_Rtik_Type_Record =>
+            Base_Record := To_Ghdl_Rtin_Type_Record_Acc(Rtii.Typ.Rti);
+         when Ghdl_Rtik_Subtype_Array =>
+            Base_Record := To_Ghdl_Rtin_Type_Record_Acc(
+               To_Ghdl_Rtin_Subtype_Composite_Acc(Rtii.Typ.Rti).Basetype);
+         when others =>
+            Internal_Error("get_record_child");
```
tgingold, line 244: Create a function to get the base type of a record type/subtype.

src/grt/grt-rtiis.adb
```diff
@@ -0,0 +1,350 @@
...
+            Layout_Addr := Rtii.Typ.Layout_Addr;
+         when others =>
+            Internal_Error("get_array_element_type");
+      end case;
+      Bounds_Addr := Layout_Addr + Ghdl_Index_Type'(
+         Ghdl_Indexes_Type'Size / 8);
+      -- Get the length of this dimension so we can check whether
+      -- the index is valid.
+      Idx_Rti := Get_Base_Type (Base_Array.Indexes (Rtii.Typ.Dim));
+      Extract_Range (Bounds_Addr, Idx_Rti, Rng);
```
tgingold, line 114: Maybe the multi-dim arrays should be flattened (as if they had only one dimension).  That could simplify the code.

src/grt/grt-rtiis.adb
```diff
@@ -0,0 +1,350 @@
...
+         when Ghdl_Rtik_Type_Array =>
+            Base_Array := To_Ghdl_Rtin_Type_Array_Acc(Rtii.Typ.Rti);
+            Layout_Addr := Rtii.Typ.Binding_Layout_Addr;
+         when Ghdl_Rtik_Subtype_Array =>
+            Base_Array := To_Ghdl_Rtin_Type_Array_Acc(
+               To_Ghdl_Rtin_Subtype_Composite_Acc(Rtii.Typ.Rti).Basetype);
+            Layout_Addr := Rtii.Typ.Layout_Addr;
+         when others =>
+            Internal_Error("get_array_child");
+      end case;
```
tgingold, line 148: You should create a function to get layout_addr and base_array from the type rtii.

benreynwar:
Thanks for taking the time to look at this.
I'll make these changes after work today, and also modify the avhpi handles so that they actually use these structures.  That should give us a better idea of whether this is the right direction to go in.

src/grt/grt-rtiis.ads
benreynwar: Do you have a formatting checker you're using in emacs?

tgingold:
Not a complete one in emacs.  I follow the indentation given by emacs but it doesn't deal with anything else.

src/grt/grt-rtiis.adb
benreynwar: Yep.  I need to check how this is handled in VPI and VHPI and then try to stay close to that.

benreynwar:
I've got lots of bugs in fix here :). I expect I'll have something ready for you to take a look at again by the end of next weekend.

tgingold:
That works for me.  Thanks!

benreynwar:
Any hints on why the llvm & gcc tests are failing?

tgingold:
Those errors:
```
C:/projects/ghdl-psgys/src/grt/grt-rtiis.adb:44: undefined reference to `system__concat_3__str_concat_3'
5916C:\Tools\GHDL\master-1212-mingw64-llvm\lib/ghdl/libgrt.a(grt-rtiis.o): In function `grt__rtiis__get_array_child':
5917C:/projects/ghdl-psgys/src/grt/grt-rtiis.adb:198: undefined reference to `system__img_uns__image_unsigned'
```

You are using string concatenation and 'image.  There are implemented using functions from the Ada library, but the executable generated with llvm and gcc backend is not linked with the Ada library.

For 'image, you can use grt-images
For concatenation, you'd better to avoid it or do it manually.

If this is just debug code then it doesn't matter.

benreynwar:
Makes sense.  Thanks.

benreynwar:
I've made most of the changes you suggested.  The only one I haven't implemented so far was adding a `Layout_Address` type, which I plan to do, I'm just waiting until I'm sure I understand exactly how all these addresses are used.

This branch is passing all the tests, which I think is more likely to indicate gaps in the tests, rather than that this is ready to merge :).  My main concern is that I will break the generation of wave dumps.  To check this I'm going to add a test that creates wave dumps and then confirm that these wave dumps are unchanged when the test is run in this branch.

The main change in this branch since last time that you looked at it, is that I'm now using the RTII structure in the Avhpi handle.  It's ready for you to take a look at it now if you'd like, or you can wait until I've tested it more thoroughly.

benreynwar:
As suspected this pull request is currently totally breaking the generation of wave file dumps.
I'm going to create some tests for the GHW dumps in a separate pull request before I come back to this (#1323).

benreynwar:
Just an update, since I haven't made any changes to this issue for ages.  I've been using this branch at work for the last month without issues, so it likely doesn't break too much.
I'm very slowly working on improving the VPI using this branch as the base.  In the mean time I don't think it's worth merging in this branch until I've made sufficient improvements to the VPI support to justify this branch.

benreynwar:
So I've haven't worked on this for ages, and have mostly forgotten where I was at.

What I do remember was that it was really, really hard, and I wasn't making very much progress.   I felt like I was having to reverse engineer the grt structure because it wasn't clearly defined anywhere.  I think a first step is probably making the grt more comprehensible.   I'd be curious to hear what people with a better understanding of ghdl think the proper approach would be.

benreynwar:
Much of what I was doing was creating structures to capture the runtime information.  But I wasn't touching any of the code related to actually creating that data in the first place.  I think a more sensible approach would be to use the same structures both for creating the runtime information in the first place, and for accessing it later for use with VPI or VHPI.

tgingold:
Do you mean grt structures or RTI structures.  I suppose the latter.

I agree they aren't simple to deal with.

They represent the tree at runtime.  But this is a waste of efforts, as the tree is well described by the AST at compile time, and finally the RTI is just a different dump of the AST (with some extra data).  So basically, we need to maintain both the AST and the RTIs.

On top of that, to deal with dynamic elaboration, the RTIs have a mix of offsets and of absolute addresses and we need to deal with the complexity of the offsets at runtime.

A different option would be to only use JIT (or code generation in the same process).  So the AST would be available and just have to be decorated with info (addresses/offsets) from the code generation.  No more RTIs.  I think that would simplify VPI or VHPI.

@Blebowski did some contributions related to the RTIs.  You can ask his feelings as a GHDL contributor.

umarcor:
@benreynwar thanks for bringing this discussion.

Let me ping @marlonjames, who's been working on VHPI.

marlonjames:
My (very slow) incremental work on the VHPI implementation hasn't taken me this far yet, so I don't have any useful input. I agree that it's better to avoid dual maintenance.

benreynwar:
Currently there's not only dual maintenance between the AST and the RTI, but also between where the RTIs are created and the various places that they are used.  There's a lot of logic about interpreting the various absolute and offset addresses that is repeated independently and is confusing.  It would be nice if we could capture all of the logic about the various kinds of offsets in one place.  That's kind of what I was trying to do last year with the RTII structure, but I was only looking at making sense of the data after it had already been constructed in memory.  It would also be nice to share structures with the code that is setting up the memory in the first place.  I'm probably missing a lot of the subtlety.

> A different option would be to only use JIT (or code generation in the same process). So the AST would be available and just have to be decorated with info (addresses/offsets) from the code generation. No more RTIs. I think that would simplify VPI or VHPI.

This sounds like a big change.  @tgingold, you can much better judge whether this is worth it.  Do you think it is?

tgingold:
> This sounds like a big change. @tgingold, you can much better judge whether this is worth it. Do you think it is?


It is a big change, and I think it is worth it (or at least worth investigating it).  But that won't happen soon.

umarcor:
Some might already know, but others might be mislead by my activity. Let me clarify that I'm no expert on the LRM, on Ada or on GHDL internals. The AST, RTI, GRT, etc. are really blurry in my mind. Hence, please excuse me if any of the following comments don't make sense.

> What I do remember was that it was really, really hard, and I wasn't making very much progress. I felt like I was having to reverse engineer the grt structure because it wasn't clearly defined anywhere. I think a first step is probably making the grt more comprehensible. I'd be curious to hear what people with a better understanding of ghdl think the proper approach would be.

My perception is that this feeling is shared by all the people that tries to develop some tool on top of GHDL:

- Cosim
  - VPI (#101, #237, #762, #1249, #1299, #1526)
  - VHPI (#1648)
  - DPI (#1053) ([VHDL/Compliance-Tests: cosim/dpi-ffi](https://github.com/VHDL/Compliance-Tests/tree/master/cosim/dpi-ffi))
- pyGHDL.libghdl (#1437, #1492, #1494, #1557, #1671, #1676)
  - pyVHDLModel (to be used in Symbolator, Sphinx...) (#1553)
  - Python API for analysis (pyGHDL.lsp) and elaboration
  - Python API for runtime (GRT)
- GHW
  - dump (#1695)
  - libghw (bindings for Python, Rust...)

I know not all of those areas are using exactly the same types/tree. But all of them need to somehow understand either the AST or the RTI or some variation/subset of both.

> That's kind of what I was trying to do last year with the RTII structure, but I was only looking at making sense of the data after it had already been constructed in memory. It would also be nice to share structures with the code that is setting up the memory in the first place. I'm probably missing a lot of the subtlety.

It seems that you are proposing some heavy internal refactoring with no other initial purpose than reshaping the codebase. I believe that can be a very exciting activity for several of us to try helping. @Blebowski has been doing some arbitrary cleanups in this sense, at his own will. Ben, Tristan, maybe any of you can point us to where should we look for "cleaning up the code that is setting up the memory"? Related to Tristan's comment about JIT, my perception is that all that offset nightmare is not only a matter of code cleanness, but might also be forced by the complexity of supporting two compilation strategies.

For context, the discussion about switching GHDL to a JIT only solution was brought previously in the context of top-level generics. See #1388. As Tristan commented, switching to JIT would imply getting rid of the GCC backend/build. However, LLVM supports JIT. When we last talked about it, there were two main features missing from LLVM which were supported by GCC: debugging and code coverage. Since then, debugging is supported with LLVM too. Yet, code coverage is still possible with GCC only, and that's the main reason for many users of GHDL to pick that backend, specially in aero-space-mil industries. Therefore, a nice strategy would be to investigate how to do coverage with LLVM and document/advertise it for moving the user base to mcode and LLVM only. As @Blebowski commented in #1695, that would reduce the maintenance burden.

PR #1304, opened 2020-05-12
Adding some comments to the RTI processing.

benreynwar:
This adds a bunch of comments to the RTI processing mostly in grt-rti.adb.

I'm using the term "instantiation data" to refer the data for an RTI node that is used for a particular context of that node.  Not sure if that's a good choice of name.


tgingold:
Thanks!

PR #1320, opened 2020-05-17
rework 'version.tmp'

umarcor:
This PR includes multiple enhancements to how `version.tmp` is generated:

- <strike>Instead of being a make target, `version.tmp` is generated in `configure`</strike>.
- <strike>Instead of requiring `make` to override the version description</strike>, environment variable `GHDL_DESC` is supported <strike>by `configure`</strike>. The precedence is as follows:
  - If this is a cloned repo and git exists, `git describe --dirty`.
  - If `GHDL_DESC` is defined, use it.
  - Default to `tarball`.
- The new environment variable is used in `dist/ci-run.sh` for customizing the description of the packages built in CI: `GHDL_DESC="$(git describe --dirty)@${BUILD_IMAGE_TAG}"`.
- Updating actions/checkout to v2 revealed that Windows build scripts have the same "unshallow" issue as Linux builds. It is fixed here.
- Either with v1 or v2 of actions/checkout, clones on Windows produce a "dirty" description. Note that this is independent from previous changes. To work around it, in this PR autocrlf is set to `input` on Windows jobs.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -4,6 +4,6 @@ jobs:
   linux:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
     - name: build doc
       run: ./doc/make.sh
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -61,6 +61,8 @@ jobs:
       with:
         msystem: MSYS
         update: true
+        install: base-devel git
+    - run: git config --global core.autocrlf input
     - uses: actions/checkout@v2
     - name: Build and (hopefully) install package
       shell: msys2 {0}
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -10,7 +10,7 @@ jobs:
   gpl:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
     - run: |
         TASK=buster+mcode ./dist/ci-run.sh -c --gpl --no-synth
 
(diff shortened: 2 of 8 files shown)


tgingold:
Why is version.ads now generated from configure ?
I am not against that, but if you first configure, then change the sources, the dirty flag won't be present.  So maybe we should remove the `--dirty` option.

umarcor:
> Why is version.ads now generated from configure ?

The proposal is to generate `version.tmp` from configure. `version.ads` is still a Makefile target.

> I am not against that, but if you first configure, then change the sources, the dirty flag won't be present. So maybe we should remove the `--dirty` option.

The issue with the dirty option is producing problems on Windows, because of CRLF conflicts when cloning on Windows and using MSYS2.

At the same time, if users rerun configure after modifying the sources, the dirty flag will be present.

Hence, although what you say is correct, I think that preserving `--dirty` is still useful.

tmeissner:
Mhm, I never re-run `configure`. I only run it when I build ghdl the first time. As any other project I know. If that changes, we have to document that very clearly.

umarcor:
@tmeissner, I'll try to move it back to the Makefile, while preserving other fixes/enhancements. Hope it still works...

tgingold:
Are the windows builds failures expected ?

umarcor:
> Are the windows builds failures expected ?

Something went really wrong upstream and running `pacman -Syu` on a clean install is failing. This is unrelated to this PR. In fact, all of the windows builds are failing because of it. Let's keep this on hold until that's fixed. I'm following https://github.com/msys2/MSYS2-packages/issues/1960

tgingold:
Ok.  Thank you for the reply.

umarcor:
I think that Windows builds are fixed now. I had to split `build()` and `package()` because of an strange hang: https://github.com/msys2/MSYS2-packages/issues/1965#issuecomment-631149718. Apart from that, previous issues on MINGW64 seem to be fixed. Synthesis seems to be quite slow compared to other platforms/backends, but it works.

Moreover, I avoided the requirement for MSYS2 to inherit the PATH by setting autocrlf to `input`.

@tmeissner, I kept it as a makefile target in the end.

tgingold:
Thanks.

PR #1327, opened 2020-05-20
ghwdump should handle zero-length signals

benreynwar:
This adds a test that checks that ghwdump doesn't crash on a signal of zero length.
The solution in the pull request just removes the offending assertion in ghwlib.c.

Presumably there is a good reason that this assertion was there, and a better solution is required.

The error message when the assertion is present is:
```
ghwdump: ../../../src/grt/ghwlib.c:1317: ghw_disp_hie: Assertion `sigs[0] != GHW_NO_SIG' failed.
```

### DIFF
--- a/src/grt/ghwlib.c
+++ b/src/grt/ghwlib.c
@@ -1313,7 +1313,7 @@ ghw_disp_hie (struct ghw_handler *h, struct ghw_hie *top)
 	    ghw_disp_subtype_indication (h, hie->u.sig.type);
 	    printf (":");
 	    k = 0;
-	    assert (sigs[0] != GHW_NO_SIG);
+
 	    while (1)
 	      {
 		/* First signal of the range.  */
--- a/testsuite/gna/issue1326/mytestbench.vhdl
+++ b/testsuite/gna/issue1326/mytestbench.vhdl
@@ -1,0 +1,17 @@
+library ieee ;
+
+entity mytestbench is
+end mytestbench;
+
+architecture arch of mytestbench is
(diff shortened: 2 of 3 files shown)


tgingold:
Thanks.

I am not sure whether we could test with ghwdump as it is not built by default.

benreynwar:
I added a test in this pull request that just built ghwdump in the testsuite, which probably isn't the best approach.  I've also noticed that it's failing on windows in the automated test.  Perhaps this is due to the way I'm building ghwdump.

eine:
> I've also noticed that it's failing on windows in the automated test. Perhaps this is due to the way I'm building ghwdump.

I think this might be some specific issue with `mcode`. The MINGW64 LLVM test is successfull. Only MINGW32 mcode fails.

benreynwar:
The pull request introduced additional failures in the windows test and also fails in MINGW64 LLVM.
```  gna issue1326: failed
  analyze mytestbench.vhdl
  elaborate mytestbench
  simulate mytestbench --wave=dump.ghw ( mytestbench --wave=dump.ghw )
  ./testsuite.sh: line 13:  3383 Segmentation fault      ./ghwdump -ths dump.ghw > dump.txt
```

eine:
@benreynwar, see https://github.com/ghdl/ghdl/runs/691651985#step:6:235 (merge of this PR) and https://github.com/ghdl/ghdl/runs/691742331#step:9:235 (a later commit by Tristan). All the MINGW64 runs were successful today: https://github.com/ghdl/ghdl/commits/master

PR #1328, opened 2020-05-20
WIP: Add a test that compares a ghw dump with a known previous dump to che…

benreynwar:
This adds a test that checks that the ghw dump hasn't changed.
It compares the output of the ghwdump tool so that if there is a regression, it's clearer what the actual issue is.

I've marked it as WIP because something is going wrong with the way I'm building ghwdump that's causing it to fail for windows in the automated testsuite.  Also it should probably not be getting built in the testsuite script.

Any tips on how I should be building ghwdump would be appreciated :).


eine:
> I've marked it as WIP because something is going wrong with the way I'm building ghwdump that's causing it to fail for windows in the automated testsuite.

As commented in #1327, I think that the current issue with master is related to mcode on Windows only. Hence, I believe that you are building ghwdump well enough. However, since ghwdump is not built and tested by default, it might be slightly outdated/broken. We should first fix master, and then this PR should be trivial.

> Also it should probably not be getting built in the testsuite script.
>
> Any tips on how I should be building ghwdump would be appreciated :).

ghwdump is not built by default, but it is a make target: https://github.com/ghdl/ghdl/blob/master/Makefile.in#L451-L464. So, you can add `make ghwdump` after the following lines:

- MINGW32 mcode: https://github.com/ghdl/ghdl/blob/master/dist/msys2-mingw/mcode/PKGBUILD#L21
- MINGW64 llvm: https://github.com/ghdl/ghdl/blob/master/dist/msys2-mingw/llvm/PKGBUILD#L21
- Linux: https://github.com/ghdl/ghdl/blob/master/dist/ci-run.sh#L311

Overall, the build scripts above are meant for CI. Hence, it is ok if we add non-defaults (such as building ghwdump). However, I believe you need to enhance the makefile to add an specific target that installs ghwdump. Or, alternatively, do come `cp` after building it.

testsuite/gna/issue1323/testsuite.sh
```diff
@@ -0,0 +1,25 @@
...
+elab myentity
+
+simulate myentity --wave=dump.ghw | tee mydesign.out
+
+gcc ../../../src/grt/ghwdump.c ../../../src/grt/ghwlib.c -I../../../src/grt/ -o ghwdump
+
+# We're just checking that ghwdump doesn't crash on a zero length signal.
+./ghwdump -ths dump.ghw > dump.txt
+
+if cmp -s dump.txt golden_dump.txt; then
```
Xiretza, line 15:
```suggestion
if diff dump.txt golden_dump.txt; then
```

This will show any differences in the test log (and continues to show nothing if the test passes).

tgingold:
For diff, you should use `diff --strip-trailing-cr` so that it also works on windows.

PR #1356, opened 2020-06-06
WIP: Synthesis of PSL built-in rose() function.

tmeissner:
So, now I try to implement synthesis of the PSL built-in `rose()` function. The code compiles, but the behaviour of the test case I have (not yet checked-in) is strange. Looks like `rose()` only works correct when the signal is one cycle valid after the rising edge 😕 
When the signal is two or more cycles valid, that rose seems to be valid each of these cycles, which is strange ...

The property `rose(a) -> not prev(a) and a;`
works with a = `__-__-__-__`
but not with a = `__-__-__--_`

I'm sure that using the `Build_Posedge()` function is too naive. Furthermore the ghdl-yosys-plugin throws an error:

```
ERROR: wire not found for $posedge
```

### DIFF
--- a/src/synth/synth-expr.adb
+++ b/src/synth/synth-expr.adb
@@ -1927,6 +1927,37 @@ package body Synth.Expr is
 
    end Synth_Psl_Stable;
 
+   function Synth_Psl_Rose (Syn_Inst : Synth_Instance_Acc; Call : Node)
+                              return Valtyp
+   is
+      Ctxt    : constant Context_Acc := Get_Build (Syn_Inst);
+      DffCurr : Net;
+      Dff     : Net;
+      NotDff  : Net;
+      Clk_Net : Net;
+      Expr    : Valtyp;
+      Res     : Net;
+   begin
+      Expr := Synth_Expression (Syn_Inst, Get_Expression (Call));
+
+      Clk_Net := Synth_Psl_Function_Clock(Syn_Inst, Call, Ctxt);
+
+      DffCurr := Get_Net (Ctxt, Expr);
+      Set_Location (DffCurr, Call);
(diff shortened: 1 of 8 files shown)


tgingold:
Yes, you shouldn't use Build_Posedge.  This is only for dff clock.

You should use something like: v and (not prev(v)), where prev() is the dff.

tmeissner:
Ah, okay. I have now created an negated version of the old value and feed that with teh current value into an AND gate:

```ada
   function Synth_Psl_Rose (Syn_Inst : Synth_Instance_Acc; Call : Node)
                              return Valtyp
   is
      Ctxt    : constant Context_Acc := Get_Build (Syn_Inst);
      DffCurr : Net;
      Dff     : Net;
      NotDff  : Net;
      Clk_Net : Net;
      Expr    : Valtyp;
      Res     : Net;
   begin
      Expr := Synth_Expression (Syn_Inst, Get_Expression (Call));

      Clk_Net := Synth_Psl_Function_Clock(Syn_Inst, Call, Ctxt);

      DffCurr := Get_Net (Ctxt, Expr);
      Set_Location (DffCurr, Call);
      Dff := Build_Dff (Ctxt, Clk_Net, DffCurr);
      Set_Location (Dff, Call);

      NotDff := Build_Monadic (Ctxt, Id_Not, Dff);
      Set_Location (NotDff, Call);

      Res := Build_Dyadic (Ctxt, Id_And,
             NotDff, DffCurr);
      Set_Location (Res, Call);

      return Create_Value_Net (Res, Boolean_Type);

   end Synth_Psl_Rose;
```

This seems to work, the test cases are successful. Also my formal checks work correctly.

So, I think this ready to merge 😃

tgingold:
Thanks!

PR #1359, opened 2020-06-07
ci(msys2): define deps in PKGBUILD, cleanup run.sh

umarcor:
With this PR dependencies of MSYS2 packages are defined in PKGBUILD files, instead of installing them in the `run.sh` script. Then, scripts are simplified.


tgingold:
Thanks!

PR #1397, opened 2020-07-17
Partially fix vpi_get_vlog_info

### LINKED ISSUE #1396: vpi_get_vlog_info not fully implemented
This function is used by cocotb in a few ways:
1. To detect the simulator name and version to exempt tests in our regressions that we expect not to pass (soon may be a polished user feature as well).
2. To pass arguments (plusargs) to cocotb for parameterizing tests.

Issues:
- [ ] returns `0` for failure
- [ ] argc/argv not returned
- [ ] `Version` is hardcoded to "0.1" and doesn't track the release number

ktbarrett:
This PR implements some of the missing functionality in the `vpi_get_vlog_info` VPI method: the version returned by the function now follows the release version, and the function returns `1` indicating success. While functionality is still missing, argc = `0` and argv = `NULL` is valid, so returning failure is not necessary.

This leaves one last bit of missing functionality in this function: returning argc and argv to C. While I'd like to see this implemented, I'm not familiar enough with Ada to do so.

This PR partially fixes #1396.

### DIFF
--- a/src/grt/grt-vpi.adb
+++ b/src/grt/grt-vpi.adb
@@ -55,6 +55,7 @@ with Grt.Rtis_Types;
 with Grt.Std_Logic_1164; use Grt.Std_Logic_1164;
 with Grt.Callbacks; use Grt.Callbacks;
 with Grt.Vstrings; use Grt.Vstrings;
+with Version;
 
 package body Grt.Vpi is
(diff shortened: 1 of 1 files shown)


tgingold:
Thanks!

Issue #1255, opened 2020-04-22
Allow running the test suite in tree (without install)


FFY00:
**Is your feature request related to a problem? Please describe.**
Yes, when creating a package for GHDL I should also run the tests, this happens before the install step. It seems like the current testsuite runs with the system GHDL instead of the tree GHDL, requiring us to install before running it.

**Describe the solution you'd like**
I would like to have a new make target, `test`. When invoked it would run the testsuite with the GHDL binary, and the other required files if needed, from the build tree.

```sh
make test
```

**Describe alternatives you've considered**
I could install GHDL in a temporary destination and run the test suite from there, but this is hacky and not the great solution. I could also try to pass the correct environment to the testsuite but this is not trivial and backend dependent (eg. it would be different between a gcc and a llvm build).

**Additional context**
It is common practice to have the build system provide a test target, instead of running the tests from the system.

tgingold:
Just do:
 GHDL=my_ghdl_binary ./testsuite.sh

FFY00:
As I said, this differs between the builds. And I want to test everything, not just the GHDL binary, this includes the libs. Since the Makefile knows where everything is located it should be fairly easy to make add a target to run the testsuite.

eine:
I've been quite busy this week and I yet have to check several of @FFY00's comments and commits regarding packaging for Arch Linux. However, if this issue is to replace `./testsuite.sh` with `make test` in the PKGBUILD file, I'm afraid that there should be no difference. This is because the testsuite expects both passing and failing tests. When I tried to run the testsuite from the PKGBUILD file, I found that such behaviour would make the build fail on the first test which was indeed expected to fail. I'd be glad to know whether this can be solved regardless of which command is used in the testing task.

FFY00:
@tgingold that is great, thanks! Are the in-tree libraries used by the testsuite?

> I've been quite busy this week and I yet have to check several of @FFY00's comments and commits regarding packaging for Arch Linux. However, if this issue is to replace `./testsuite.sh` with `make test` in the PKGBUILD file, I'm afraid that there should be no difference. This is because the testsuite expects both passing and failing tests. When I tried to run the testsuite from the PKGBUILD file, I found that such behaviour would make the build fail on the first test which was indeed expected to fail. I'd be glad to know whether this can be solved regardless of which command is used in the testing task.

[...]

Issue #1399, opened 2020-07-23
Package and/or install the testsuite


eine:
The [testsuite](https://github.com/ghdl/ghdl/tree/master/testsuite) is NOT self-contained. Some tests depend on files located outside of subdir testsuite:

- In #915 and #921, a check was added to some tests, in order to skip them if `doc` is not available.
- While updating MSYS2 and Arch packages in [eine/ghdl-packaging](https://github.com/eine/ghdl-packaging/tree/master) (see #901), I found that issue1326 depends on `src/grt` (https://github.com/ghdl/ghdl/blob/master/testsuite/gna/issue1326/testsuite.sh#L10).

Hence, I propose to add a `install.test` target to the Makefile, which allows to copy the testsuite and all the required assets to some destdir/tarball.

At the same time, it would be useful to add a target for testing openieee builds (i.e., which includes a subset of the tests only). Note that the reduced features of openieee are only an issue with VHDL <2008 (see #1060 and #1255).

PR #1401, opened 2020-07-23
makefile: fix 'test' when builddir != srcdir

eine:
After #1255 makefile targets `test` and `test.*` were added. However, those expect the build directory to be the same as the source directory. This PR adds `$(srcdir)` to fix the targets when that's not true.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -200,7 +200,7 @@ install.mcode: install.mcode.program install.vhdllib install.vpi install.libghdl
 uninstall.mcode: uninstall.mcode.program uninstall.vhdllib uninstall.vpi uninstall.libghdl
 
 test.mcode: ghdl_mcode$(EXEEXT)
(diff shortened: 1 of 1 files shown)


tgingold:
Thanks!

eine:
@tgingold, this fix works nice and it allows to use pacman's `check` function in Arch and/or MSYS2. However, VPI plugins seem to be broken with this procedure. That's because `include` is considered to be relative to the location of the binary (builddir), but it's not true. See https://github.com/eine/ghdl-packaging/runs/908005778?check_suite_focus=true#step:3:1114.

This is related to #1399.

tgingold:
a `make install.vpi.local` is needed before running the tests.
Maybe this should be automatic ?

eine:
Thanks a lot for the hint! I think we should have a `make test.install.local` which creates an standalone `builddir/test` (#1399). Then `install.vpi.local` would be a dependency of `test.install.local`, and this would be a dependency of `test`. But, for now, it works well enough.

PR #1411, opened 2020-07-27
ghdl-ls: Fix exception crashing language server

### LINKED ISSUE #1410: ghdl-ls: Language server crashes on unhandled symbol
**Description**
I am using the GHDL language server with Visual Studio Code on Windows.
In case of running into an unhandles symbol, the language server crashes with an exception and gets restarted. As the file is still the same, the language server is restarted five times in a row and finally dies.

**Expected behaviour**
In case of an unhandled symbol, an error should be logged and the language server should continue running. Currently the language server crashes with an exception.

**How to reproduce?**
#1409 describes how I ran into this situation.

**Context**
- OS: Windows, Visual Studio Code
- Origin: Built from sources: `541f64a6`

```
2020-07-27 08:30:27,559 [ERROR] Uncaught error
Traceback (most recent call last):
  File "lib\site-packages\vhdl_langserver\main.py", line 122, in main
  File "lib\site-packages\vhdl_langserver\lsp.py", line 98, in run
  File "lib\site-packages\vhdl_langserver\lsp.py", line 116, in handle
  File "lib\site-packages\vhdl_langserver\vhdl_ls.py", line 112, in textDocument_documentSymbol
  File "lib\site-packages\vhdl_langserver\document.py", line 180, in document_symbols
  File "lib\site-packages\vhdl_langserver\symbols.py", line 70, in get_symbols_chain
  File "lib\site-packages\vhdl_langserver\symbols.py", line 70, in <listcomp>
  File "lib\site-packages\vhdl_langserver\symbols.py", line 78, in get_symbols
  File "lib\site-packages\vhdl_langserver\symbols.py", line 141, in get_symbols
  File "lib\site-packages\vhdl_langserver\symbols.py", line 70, in get_symbols_chain
  File "lib\site-packages\vhdl_langserver\symbols.py", line 70, in <listcomp>
  File "lib\site-packages\vhdl_langserver\symbols.py", line 81, in get_symbols
AssertionError: get_symbol: unhandled Psl_Assert_Directive
[Error - 08:30:27] Connection to server got closed. Server will not be restarted.
[Error - 08:30:27] Request textDocument/documentSymbol failed.
```

**Additional context**
I fixed this issue locally by replacing the exception with logging an error and will provide a PR for this.

HenningHolmDE:
Previously, the language server crashed on unhandled symbols with an exception. As ignoring unhandled symbols should be fine, logging this as an error to the logger results in a much better user experience.

This fixes #1410.

### DIFF
--- a/python/vhdl_langserver/lsp.py
+++ b/python/vhdl_langserver/lsp.py
@@ -113,7 +113,20 @@ class LanguageProtocolServer(object):
         if fmethod:
             if params is None:
                 params = {}
-            response = fmethod(**params)
+            try:
+                response = fmethod(**params)
+            except Exception as e:
+                log.exception(
+                    "Caught exception while handling %s with params %s:",
+                    method, params
+                )
+                self.show_message(
+                    MessageType.Error,
+                    ("Caught exception while handling {}, " +
+                     "see VHDL language server output for details.").format(
+                        method)
+                )
+                response = None
             if tid is None:
                 # If this was just a notification, discard it
                 return None
--- a/python/vhdl_langserver/symbols.py
+++ b/python/vhdl_langserver/symbols.py
@@ -1,3 +1,4 @@
+import logging
 import libghdl.thin.name_table as name_table
 import libghdl.thin.files_map as files_map
 import libghdl.thin.vhdl.pyutils as pyutils
@@ -6,6 +7,8 @@ import libghdl.thin.vhdl.nodes_meta as nodes_meta
 import libghdl.thin.vhdl.elocations as elocations
 
 from . import lsp
+
+log = logging.getLogger(__name__)
 
 SYMBOLS_MAP = {
     nodes.Iir_Kind.Package_Declaration: {'kind': lsp.SymbolKind.Package, 'detail': '(declaration)'},
@@ -78,7 +81,8 @@ def get_symbols(fe, n):
         return get_symbols(fe, nodes.Get_Library_Unit(n))
     m = SYMBOLS_MAP.get(k, None)
     if m is None:
-        raise AssertionError("get_symbol: unhandled {}".format(pyutils.kind_image(k)))
+        log.error("get_symbol: unhandled {}".format(pyutils.kind_image(k)))
+        return None
     kind = m['kind']
     if kind is None:
         return None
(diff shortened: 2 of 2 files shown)


tgingold:
Why not, but it also hides issues.  Nobody looks at logs.

tgingold:
I will first fix the issue.

eine:
> Why not, but it also hides issues. Nobody looks at logs.

Can we add some very visible warning/notification letting users know that all unhandled symbols are shown in the logs? Maybe it can be an option of the language server (or the client).

HenningHolmDE:
> Why not, but it also hides issues. Nobody looks at logs.

That's always a tough one. On the one hand, I also like my code to fail on me loud enough to wake the dead, so I know that I have to fix something. On the other hand for the avarage user every crash is a reason to go "nah, doesn't work properly". For production code, I would always go with the latter.

It also took me some time (and many crashes) to go from "crashes at random" to "crashes when there is an assertion and no entity" and come up with a useful issue for you to work with - which is probably not for everyone.

HenningHolmDE:
> Can we add some very visible warning/notification letting users know that all unhandled symbols are shown in the logs? Maybe it can be an option of the language server (or the client).

Giving this some thought, I checked about using LSP `window/showMessage` to be a bit more obtrusive, but as currently neither `Document` nor `get_symbols` know about the `VhdlLanguageServer` anymore, achieving this would probably be pretty ugly.
Modifing `LanguageProtocolServer.handle` to catch any exceptions being raised when calling `fmethod` and shout them out to the user without crashing could be an alternative.

HenningHolmDE:
I gave it a try and think this could work as well:
```diff
--- a/python/vhdl_langserver/lsp.py
+++ b/python/vhdl_langserver/lsp.py
@@ -113,7 +113,20 @@ class LanguageProtocolServer(object):
         if fmethod:
             if params is None:
                 params = {}
-            response = fmethod(**params)
+            try:
+                response = fmethod(**params)
+            except Exception as e:
+                log.exception(
+                    "Caught exception while handling %s with params %s:",
+                    method, params
+                )
+                self.show_message(
+                    MessageType.Error,
+                    ("Caught exception while handling {}, " +
+                     "see VHDL language server output for details.").format(
+                        method)
+                )
+                response = []
             if tid is None:
                 # If this was just a notification, discard it
                 return None
```

What do you think about it?

tgingold:
Yes, that looks good!

HenningHolmDE:
Thanks for the positive feedback!

I now reverted my previous commit and implemented the alternative.

tgingold:
Thanks!

HenningHolmDE:
You're welcome!

Issue #1027, opened 2019-11-20
Not able to compile PSL verification unit file and bind to module. 


steamka90:
**Description**
Not able to compile PSL verification unit file and bind to module. 

**How to reproduce?**
I was trying to compile `dff.vhd` and `dff_psl.vhd` file using VHDL-2008 and VHDL-93 mode.

```vhd :file: dff.vhd
library ieee;                                                                                                                                                                                                      
use ieee.std_logic_1164.all;                                                                                                                                                                                       
                                                                                                                                                                                                                   
entity dff is                                                                                                                                                                                                      
  port(                                                                                                                                                                                                            
    reset : in std_logic;                                                                                                                                                                                          
    clk : in std_logic;                                                                                                                                                                                            
    d : in std_logic;                                                                                                                                                                                              
    q : out std_logic                                                                                                                                                                                              
  );                                                                                                                                                                                                               
end entity dff;                                                                                                                                                                                                    
                                                                                                                                                                                                                   
architecture rtl of dff is                                                                                                                                                                                         
  signal q_int : std_logic;                                                                                                                                                                                        
begin                                                                                                                                                                                                              
                                                                                                                                                                                                                   
  dff_proc : process(clk, reset)                                                                                                                                                                                   
  begin                                                                                                                                                                                                            
    if reset = '1' then                                                                                                                                                                                            
      q_int <= '0';                                                                                                                                                                                                
    elsif rising_edge(clk) then                                                                                                                                                                                    
      q_int <= d;                                                                                                                                                                                                  
    end if;                                                                                                                                                                                                        
  end process dff_proc;                                                                                                                                                                                            
                                                                                                                                                                                                                   
  -- drive q_int to output port                                                                                                                                                                                    
  q <= q_int;                                                                                                                                                                                                      
                                                                                                                                                                                                                   
end rtl;  
```

```vhd :file: dff_psl.vhd
vunit dff_psl (dff(rtl))                                                                                                                                                                                           
{                                                                                                                                                                                                                  
  default clock is rising_edge(clk);                                                                                                                                                                               
  async_reset_check:                                                                                                                                                                                               
    assert always {reset = '1'} |-> {not q_int}                                                                                                                                                                    
      report "Async reset activated, q_int /= '0' during reset";                                                                                                                                                   
}   
```
Using VHDL-2008 complication option:
```sh :image: ghdl/ghdl:buster-mcode
ghdl -a --std=08 dff.vhd dff_psl.vhd
```

[...]

PR #1418, opened 2020-07-30
doc: Update and fix PSL documentation (vunits, see #1027, PSL functions)

tmeissner:
Small addition to the documentation of PSL support. As mentioned in #1027 PSL vunits are only supported with VHDL-08 and synthesis. These PR adds that information to the doc.


tgingold:
Thanks!

PR #1434, opened 2020-08-14
py: adjust blank lines to PEP 8 for vhdl_langserver

m-kru:
Why?
1. Conform to PEP 8.
2. Easier to read.
3. Make tools, such as PyCharm, not complain.

### DIFF
--- a/python/libghdl/__init__.py
+++ b/python/libghdl/__init__.py
@@ -5,6 +5,7 @@ from os.path import dirname, join, exists, normpath
 from shutil import which
 from libghdl.version import __version__
 
+
 def _to_char_p(arg):
     return ctypes.c_char_p(arg), len(arg)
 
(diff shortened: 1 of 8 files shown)


eine:
LGTM. Thanks!

BTW, if you expect to invest some time on these Python sources, do you think it is worth to use an automatic formatter? When we had this discussion in VUnit a year ago, we concluded that using black was preferred, precisely in order to avoid discussions about style. See #554. Hence, if you think it would be valuable, adding `black` to CI is trivial.

There are other lint tools/test which we could just copy/reuse: https://github.com/VUnit/vunit/tree/master/tests/lint

m-kru:
@eine definitely. Uncompromising formatters (such as `black`, `rustfmt`, `gofmt`) are great as they format your code in unified way and end unproductive discussions about formatting styles. We need `vhdlfmt` ;).

eine:
> We need vhdlfmt ;).

Not exactly, because I think that comments are not preserved in GHDL, but there is https://ghdl.readthedocs.io/en/latest/using/CommandReference.html?highlight=pretty#pretty-print-pp-html

PR #1439, opened 2020-08-15
Add and use 'black' Python formatter

eine:
As discussed in #1434, in this PR `black python` was executed, to have the code in subdir `python` automatically formatted by [black](https://github.com/psf/black). Moreover, a test is added to CI, which executes `black --check python`. As a result, users modifying the code should execute `black python` (they can optionally set a hook), before pushing changes to a PR. Otherwise, CI might fail. Apart from that, we can forget about formatting/style issues (in Python)!

/cc @m-kru

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -10,6 +10,24 @@ env:
 jobs:
 
 #
+# Format
+#
+
+  fmt:
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@v2
+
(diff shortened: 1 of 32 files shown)


tgingold:
Ok, but some of these files (nodes.py, nodes_meta.py) are automatically generated.

eine:
> Ok, but some of these files (nodes.py, nodes_meta.py) are automatically generated.

Just add `black python` (from the root) or `black path/to/file`, after the files are regenerated, to have them formatted automatically. Is it done in some makefile?

PR #1441, opened 2020-08-18
Provide initial skeleton for libghdl documentation.

m-kru:
Not a lot of useful information, but some questions, that I hope @tgingold will be able to answer.

PR #1445, opened 2020-08-25
Provide initial skeleton for libghdl documentation.

m-kru:
Reopen of https://github.com/ghdl/ghdl/pull/1441

### DIFF
--- a/testsuite/gna/issue1490/ent.vhdl
+++ b/testsuite/gna/issue1490/ent.vhdl
@@ -1,0 +1,7 @@
+entity ent is
+end ent;
+use work.ent.ent; -- ***this second .ent causes the crash***
+architecture foo of ent is
+begin
+end foo;
+
--- a/testsuite/gna/issue1490/testsuite.sh
+++ b/testsuite/gna/issue1490/testsuite.sh
@@ -1,0 +1,9 @@
+#! /bin/sh
+
+. ../../testenv.sh
+
+analyze_failure ent.vhdl
+
+clean
+
+echo "Test successful"


eine:
FTR, I believe this is the most simple (and single) example available using libghdl: https://github.com/ghdl/ghdl/tree/master/testsuite/python/units01

doc/libghdl/ASTTraversing.rst
```diff
@@ -0,0 +1,12 @@
+.. _LIB:ASTTraversing:
+
+AST Traversing
+##############
+
+If someone wants to use `libghdl`, then he probably wants to know how to traverse the AST to get the information about nodes of design units.
```
tgingold, line 6: There are examples of traverse functions in `pyutils.py`.  The function `nodes_iter` is a visitor on all nodes, while `declarations_iter` and `concurrent_stmts_iter` are more selective.

doc/libghdl/ASTTraversing.rst
```diff
@@ -0,0 +1,12 @@
...
+It would be really nice to provide an example showing how to traverse the tree.
```
tgingold, line 7: Have a look at `vhdl_langserver/symbols.py` to see how symbols are gathered from a vhdl file. Or `references.py`.

doc/libghdl/ASTTraversing.rst
```diff
@@ -0,0 +1,12 @@
...
+In the `vhdl_ls` Load_File function is used to parse the file and get the tree root.
+The return value is an integer.
+
+* What does this integer represent?
```
tgingold, line 11: It is not an integer, but a reference to a node (which is an index so indeed an integer).  In Ada the fact that this is an integer is almost hidden, but that is clear in the low-level binding.  (Hence a reason why we should have a high-level and more pythonic API).

The value, if not null, represents an Iir_Kind_Design_File.

doc/libghdl/ASTTraversing.rst
```diff
@@ -0,0 +1,12 @@
...
+* How to use this integer to traverse the AST and get information about nodes?
```
tgingold, line 12: I think you can call the various iterators defined in `pyutils.py`

doc/libghdl/Initialization.rst
```diff
@@ -0,0 +1,21 @@
+.. _LIB:Initialization:
+
+Initialization
+##############
+
+To be able to use `libghdl`, it first must be initialized.
+
+* Why does it must be initialized?
```
tgingold, line 8: There are many data that must be set up, like the name table.  You cannot do anything useful without it.  And then there are also some higher levels constructs that should also be initialized, like the content of the ieee library.

doc/libghdl/Initialization.rst
```diff
@@ -0,0 +1,21 @@
...
+* What does the initialization do?
+* Is there any use case where someone would want to use libghdl without initialization?
```
tgingold, line 10: No, I don't think so.

doc/libghdl/Initialization.rst
```diff
@@ -0,0 +1,21 @@
...
+* If no, then why it does not initialize itself automatically?
```
tgingold, line 11: One reason is because it is a low-level binding.  The other reason is because some initialization (like which ieee library is loaded) depends on some options.

doc/libghdl/Initialization.rst
```diff
@@ -0,0 +1,21 @@
...
+* What functions need to be called during initialization? In Python bindings 3 functions are called:
+
+.. code-block:: python
+
+   libghdl.libghdl_init()
```
tgingold, line 16: This one does Ada elaboration.  Nothing visible and could be automatic (but that would make the build infrastructure more complex).

doc/libghdl/Initialization.rst
```diff
@@ -0,0 +1,21 @@
...
+   libghdl.libghdl__set_hooks_for_analysis()
```
tgingold, line 17: Basically, it defines the back-end, which is invoked when a unit is analyzed.

doc/libghdl/Initialization.rst
```diff
@@ -0,0 +1,21 @@
...
+
+   # Set the prefix in order to locate the vhdl libraries.
+   libghdl.libghdl__set_exec_prefix(
+       *_to_char_p(dirname(dirname(_libghdl_path)).encode('utf-8')))
```
tgingold, line 21: If the comment is not clear enough, please tell me.

But this is 'just' the start of the initialization.  If you look at `vhdl_langsever/workspace.py` you will see the remaining (in `__init__`):
* redirecting errors
* setting flags and options
* end of initialization (load libraries).

doc/libghdl/Overview.rst
```diff
@@ -0,0 +1,17 @@
...
+`GHDL` provides `libghdl` shared library, that is intended to be used by advanced users.
+The library includes a subset of the regular features plus some features to be used by extension tools (**TODO**: It is really desirable to list available features).
+`libghdl` allows developers to use `GHDL` features in any language they want.
+What is needed are only language specific bindings.
+If bindings for the language you want to use do not exist, and you want to implement them, please make them raw bindings.
+Any additional abstraction or functionality should be based on raw bindings.
+
+Here we should write superficially how `libghdl` works.
+
+* Does it return structures such as AST to the calling code, or does it store them internally?
```
tgingold, line 15: The AST are always internally stored.  The reason is that a unit may reference another unit.
At some point, we may want to introduce a notion of context that would contain all the global variables.

doc/libghdl/Overview.rst
```diff
@@ -0,0 +1,17 @@
...
+* Is it thread safe?
```
tgingold, line 16: No!

eine:
@m-kru, did you close this on purpose? Or was it an undesired effect of updating your master branch?

m-kru:
@eine ehhh, again undesired effect.

eine:
Do you know how to get it back? `git checkout -b doc/libghdl cc416a50e817fbdaaa5579420051dffa258a40bf` might work.

PR #1446, opened 2020-08-25
Replace distutils with setuptools for pyghdl and remove unused import

corna:
Currently, using directly setup.py (without pip) fails to include `ghdl-ls`, as `entry_point` is not available in distutils.


tgingold:
Thanks!

PR #1448, opened 2020-08-27
Create a version header to export version strings to C

rlee287:
**Description**

Export the version strings as a C header to allow access to these strings from C extensions using GHDL. This is intended specifically for the VHDL backend at ghdl/ghdl-yosys-plugin#122, but may also be useful for other people.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -172,6 +172,17 @@ version.ads: version.tmp
 #	useless rebuild.
 	if [ ! -r $@ ] || ! cmp $< $@ > /dev/null; then $(CP) $< $@; fi
 
+# Same tasks, but generate a C header instead
+version.tmp_h: $(srcdir)/src/version.in_h force
+#	Create version.tmp_h from version.in, using git date/hash, or envvar GHDL_DESC. Defaults to 'tarball'.
+	if test -d $(srcdir)/.git && desc=`cd $(srcdir); git describe --dirty`; then GHDL_DESC="$$desc"; fi; \
+	$(SED) -e "s/[(].*[)]/($$GHDL_DESC)/" -e "s/@VER@/$(ghdl_version)/" < $< > $@; \
+
+ghdl_version.h: version.tmp_h
+#	Change ghdl_version.h only if version.tmp_h has been modified to avoid
+#	useless rebuild.
+	if [ ! -r $@ ] || ! cmp $< $@ > /dev/null; then $(CP) $< $@; fi
+
 #################### For mcode backend ##############################
 
 all.mcode: ghdl_mcode$(EXEEXT) libs.vhdl.mcode all.vpi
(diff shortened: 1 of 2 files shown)


eine:
@rlee287, the CI failure was unrelated to this PR. You should get all green if you rebase.

tgingold:
Two points:

1. Git can be very (very) slow on Windows, so you'd better to generate the C file from the Ada file, or to first generate a third file and then the Ada and C file from it.  Avoid to invoke git twice.

2. I don't think it is good style to define a const in a `.h` file.  So either generate `#define`s or a C file.

I am not sure to understand why you need this file for the VHDL back-end.  For me, this backend is independent from GHDL.

rlee287:
As to why I'm creating this file: I'm placing a comment at the top of the autogenerated files containing both the Yosys version and the GHDL version, like so:

```vhdl
-- Generated by:
-- Yosys 0.9+3527 (git sha1 925c0f25, clang 10.0.0-4ubuntu1 -fPIC -Os)
-- GHDL 1.0-dev (v0.37.0-991-g8c71ad62) [Dunoon edition]
```

I am writing this file to export the GHDL version strings in a header so that the Yosys plugin can read in this information to produce the comment.

I'll address the two points you raised in the meantime, but suggestions are welcome if you think there is a better way to do this.

eine:
> I am writing this file to export the GHDL version strings in a header so that the Yosys plugin can read in this information to produce the comment.

I think that ghdl-yosys-plugin is *independent* from GHDL. The plugin is a shared library that uses GHDL's shared library and VHDL libraries, but not GHDL itself. So, the plugin depends on the AST/API of GHDL, which is unversioned. You might use the GHDL version (SHA) as an indicator of the version of the AST/API (even though most of the changes do not modify the AST/API). However, that is different from the version of the existing plugin. Hence, you should add that too. The plugin does not have the version hardcoded because, AFAIK, yosys cannot tell the version of the modules. I guess that using the commit SHA is good.

Anyway, I believe that this does not apply to ghdl/ghdl-yosys-plugin#122. My understanding is that your work does depend neither on GHDL nor on the plugin, because you don't use any types or functions defined by them. Yours is an independent feature and codebase that depends on Yosys only. It makes sense to have it in the same repo as ghdl-yosys-plugin because it's a complement from a functionality perspective. But regarding versioning and licensing, your should be able to handle it as you want. Of course, licensing would need to be agreed with Tristan. However, that is because he is the owner of the repo, not because he is the author of GHDL or the plugin.

rlee287:
On further consideration, I think I'll just use the commit SHA of the ghdl-yosys-plugin repo, so that I won't need modifications to the GHDL repo itself for placing a usable version identifier in generated files. I'm closing this pull request accordingly.

Issue #1437, opened 2020-08-14
libghdl documentation


m-kru:
@tgingold I was analyzing `ghdl_ls` and it looks like there is no documentation for `libghdl`. Few questions came into my mind. Is `libghdl` thread safe? Can multiple programs interact with `libghdl` simultaneously? Is it safe to use `ghdl_ls` when there is some simulation running in the background? As `libghdl` is a shared library it should be possible to document its API. It would be really nice if the documentation was placed in the source files, and later fetched to the online docs. My personal feeling is that `libghdl` makes it possible to use the power of GHDL in Python, however it is not easy to start, as without documentation user still does not know what `libghdl` exactly is and how it works. I could even try to prepare some `libghdl` documentation, but I do not know where to start, what are the source files I should be interested in.

eine:
FTR, the docs of GHDL are generated with Sphinx, which was primarily designed and is mostly used for documenting Python projects. Hence, we can generate "nice" docs from docstrings written in the Python sources, and have them integrated in the docs cleanly (much better than how we document GHDL itself). See, for example, the documentation of VUnit's Python interface: http://vunit.github.io/py/ui.html

Why don't we do it for Ada and VHDL sources too? Because there is no built-in mechanism in Sphinx to do so. See #484 and https://github.com/Paebbels/sphinxcontrib-vhdldomain.

m-kru:
To be honest, I am not sure what should be documented, functions from shared library in Ada files or maybe Python bindings. It looks like right now Python bindings do not add any functionality or abstraction, example `libraries.py`:
```Python
from libghdl import libghdl
from ctypes import c_int32

Get_Libraries_Chain = libghdl.libraries__get_libraries_chain

Add_Design_Unit_Into_Library = \
    libghdl.libraries__add_design_unit_into_library

# Use .value
Library_Location = c_int32.in_dll(libghdl, "libraries__library_location")

# Use .value
Work_Library = c_int32.in_dll(libghdl, "libraries__work_library")

Purge_Design_File = libghdl.libraries__purge_design_file

Find_Entity_For_Component = libghdl.libraries__find_entity_for_component

Get_Library_No_Create = libghdl.libraries__get_library_no_create

Find_Primary_Unit = libghdl.libraries__find_primary_unit
```

By the way, these names look a bit unPythonic.

I also do not know if Python bindings should provide some extra functionality or should they only directly call functions from shared library.

m-kru:
[...]

Issue #1373, opened 2020-06-16
Commit e0ca8bf0b0d8049c7e9f7fe53f7ed942fac61453 breaks some vendor library builds


AppMathIsRealMath:
**Description**
I rebuild the xilinx, osvvm, and uvvm vendor libraries after each set of ghdl commits I update. Using gnat-2020 CE I execute a personal script containing

./compile-osvvm.sh -a
./compile-uvvm.sh -a
./compile-xilinx-vivado.sh -a --vhdl2008

This procedure successfully builds these libraries (used by me in emacs running flycheck for VHDL linting).

After commit e0ca8bf0b0d8049c7e9f7fe53f7ed942fac61453 

osvvm builds successfully but with a set of different warnings
uvvm fails with tons of warnings
xilinx-vivado fails with a bunch of new warnings similar to
../../src/vital2000/timing_p.vhdl:152:14: result subtype of a pure function cannot have access subelements
../../src/vital2000/timing_p.vhdl:152:14:note: (you can use -frelaxed to turn this error into a warning)
/home/sam/mnt/tools/xilinx/Vivado/2020.1/data/vhdl/src/unisims/unisim_VPKG.vhd:583:12:warning: declaration of "s" hides constant "s" [-Whide]

As a workaround I've created a branch at the last working commit 8eb418a695000718920c2ef5294c185a2ef40cca

Do the vendor lib build scripts need to be updated to account for commit e0ca8bf0b0d8049c7e9f7fe53f7ed942fac61453?

tgingold:
Should be fixed with the last commit.

AppMathIsRealMath:
Fix confirmed for osvvm, uvvm and xilinx-vivado (I don't build the others). Thanks!

tgingold:
Thank you for the check.

PR #1449, opened 2020-08-27
update python interface test/example

umarcor:
This is an attempt to update testsuite/python/units01 and do something more than just printing the name of the design units. I could successfully get and print the entity name that the architecture belongs to (`Get_Entity_Name`). Then, for the entity, I'd like to get the name, direction and type of the ports. I could use `nodes.Get_Port_Chain(lib_unit)` to get access to the first port (and print the name). But I don't know how to iterate and get the second port.

Note: in the first commit, `show_units.py` was formatted with `black`. Relevant changes are in the second commit.

Ref #1437 #1441 #1373 #1492

### DIFF
--- a/python/libghdl/thin/__init__.py
+++ b/python/libghdl/thin/__init__.py
@@ -1,0 +1,3 @@
+from libghdl.thin.vhdl import nodes
+from libghdl.thin.vhdl import sem_lib
+from libghdl.thin.vhdl import pyutils
--- a/testsuite/python/units01/show_ports.py
+++ b/testsuite/python/units01/show_ports.py
@@ -3,11 +3,7 @@ from sys import argv
 from pathlib import Path
 
 import libghdl
-from libghdl.thin import name_table
-from libghdl.thin import files_map
-from libghdl.thin.vhdl import nodes
-from libghdl.thin.vhdl import sem_lib
-from libghdl.thin.vhdl import pyutils
+from libghdl.thin import name_table, files_map, nodes, sem_lib, pyutils
 
 
 def get_identifier_ptr(n):


umarcor:
Thanks to https://github.com/ghdl/ghdl/issues/1437#issuecomment-678915837, I found https://github.com/ghdl/ghdl/blob/cd8f99e3bb1567c0d879b3eb68d7fe020d080280/src/synth/synth-disp_vhdl.adb#L474-L480.

umarcor:
I updated the PR with the iteration over ports. I can read the name of each port, the mode and the type. I can convert the mode to string, by comparing it with `nodes.Iir_Mode`. But no idea about how to get `std_logic` from `Get_Type(port)` (which returns `0`):

```sh
# python3 show_units.py
entity e1
  port clk in 0
  port rst in 0
architecture behav of e1
```

tgingold:
The comment about analysis is wrong: `Load_File` just parses but does not analyze.

So the `Type` field is not set, but the `Subtype_Indication` is.

If you want to know how to display the tree, have a look at `vhdl-prints.adb`.  It is not fully trivial when you aim at printing as it was in the source.

Should you create a `show_ports.py` test in order to keep `show_units` as simple as possible ?

We should really have a more pythonic and high-level layer above libghdl.

umarcor:
> The comment about analysis is wrong: `Load_File` just parses but does not analyze.

I'll fix it.

> So the `Type` field is not set, but the `Subtype_Indication` is.

It is, and it works! `get_identifier_ptr(nodes.Get_Subtype_Indication(port))`

> If you want to know how to display the tree, have a look at `vhdl-prints.adb`. It is not fully trivial when you aim at printing as it was in the source.

What do you mean with "the tree"? All the processes, and instances in the architecture? That might come in the future, but not yet. My main focus is on evaluating if https://kevinpt.github.io/symbolator can use GHDL's Python interface for generating symbols of VHDL 2008 entities with "complex" port types.

> Should you create a `show_ports.py` test in order to keep `show_units` as simple as possible ?

I will! 

> We should really have a more pythonic and high-level layer above libghdl. 

I agree. Having to compare the port direction with several constants is really unpythonic. However, I am not an expert in Python, so I'm no really sure about what to use, apart from an intermediate function.

umarcor:
Now `show_port.py` is a new file. I added two functions (`get_port_mode` and `get_port_type`) which are expected to return an string. I also added a port of Mode out and Type `std_logic_vector`. As Tristan already knew, getting an string from the subtype of this last one fails.

```
# python show_ports.py ./
· demo.vhdl
  - entity e1
    * clk in std_logic
    * rst in std_logic

raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : no field Identifier
```

Yet, the subtype is not null. Using:

```py
def get_port_type(port):
    "Return the Type of a port, as a string"
    subtype = nodes.Get_Subtype_Indication(port)
    return nodes.Get_Subtype_Indication(port)
```

we get:

```
# python show_ports.py ./
· demo.vhdl
  - entity e1
    * clk in 1292
    * rst in 1294
    * q out 1298
  - architecture behav of e1
```

Reading [vhdl-prints.adb](https://github.com/ghdl/ghdl/blob/master/src/vhdl/vhdl-prints.adb), I printed `nodes.Get_Kind(port)`. That's `127` for all of them (`std_logic` or `std_logic_vector`). I guess I do need to get from that `127` to either `59` or `60`: https://github.com/ghdl/ghdl/blob/master/python/libghdl/thin/vhdl/nodes.py#L73-L74. The problem is that [`Get_Base_Type`](https://github.com/ghdl/ghdl/blob/master/src/vhdl/vhdl-prints.adb#L642) is not defined in `nodes.py`. I think I need the base type, for mimicking https://github.com/ghdl/ghdl/blob/master/src/vhdl/vhdl-prints.adb#L673.

tgingold:
By tree, I mean the AST.

You can also display it by using: `ghdl -a -dp f.vhdl`

tgingold:
First tip: using pyutils.kind_image(x) to display the corresponding name for a kind, like:
`print(pyutils.kind_image(nodes.Get_Kind(n)))`

So yes, 127 corresponds to Interface_Signal_Declaration, which is the usual kind of a port.
But then you need to extract the subtype of it using `Get_Subtype_Indication`.
It could be a simple name or an Array_Subtype_Definition.  From it, extract the type mark and the constraints.

umarcor:
> First tip: using pyutils.kind_image(x) to display the corresponding name for a kind, like: `print(pyutils.kind_image(nodes.Get_Kind(n)))` So yes, 127 corresponds to Interface_Signal_Declaration, which is the usual kind of a port.

It is. However, it does not help with telling `std_logic` and `std_logic_vector` apart. This is the function:

```py
    kind = nodes.Get_Kind(port)
    print("kind:", kind, pyutils.kind_image(kind))

    subtype = nodes.Get_Subtype_Indication(port)
    print("subtype:", subtype)

    mark = nodes.Get_Subtype_Type_Mark(subtype)
    print("mark:", mark)

    has_index = nodes.Get_Index_Constraint_Flag(subtype)
    print("constraint:", has_index)

    clist = nodes.Get_Index_Constraint_List(subtype)
    print("clist:", clist)

    return get_identifier_ptr(subtype)
```

`kind` and `subtype` are properly retrieved for any of the ports:

```
· demo.vhdl
  - entity e1
kind: 127 Interface_Signal_Declaration
subtype: 1292
    * clk in std_logic
kind: 127 Interface_Signal_Declaration
subtype: 1294
    * rst in std_logic
kind: 127 Interface_Signal_Declaration
subtype: 1298
```

But `Get_Subtype_Type_mark(subtype)` produces a crash for ports of type `std_logic`:

```· demo.vhdl
  - entity e1
kind: 127 Interface_Signal_Declaration
subtype: 1292

raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : no field Subtype_Type_Mark
```

Unfortunately, `try` and `except` seem not to work. How to know when to `return get_identifier_ptr(subtype)` and when to continue with getting the mark and constraints?

> But then you need to extract the subtype of it using `Get_Subtype_Indication`. It could be a simple name or an Array_Subtype_Definition.

How do I check whether any of 1292, 1294 or 1298 is an array subtype definition?

> From it, extract the type mark and the constraints.

I could extract the type mark, the "index constraint flag" and the "index constraint list", by looking at:

- https://github.com/ghdl/ghdl/blob/master/src/vhdl/vhdl-prints.adb#L594
- https://github.com/ghdl/ghdl/blob/master/src/vhdl/vhdl-prints.adb#L518
- https://github.com/ghdl/ghdl/blob/master/src/vhdl/vhdl-prints.adb#L497

However, I am still missing several steps:

- https://github.com/ghdl/ghdl/blob/master/src/vhdl/vhdl-prints.adb#L524 Do I need to get `Tm_El` and compare it in the Python code too?
- https://github.com/ghdl/ghdl/blob/master/src/vhdl/vhdl-prints.adb#L530 In Python, `nodes.Get_Constraint_State(mark)` produces `raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : no field Constraint_State`.
- https://github.com/ghdl/ghdl/blob/master/src/vhdl/vhdl-prints.adb#L537
- ...

tgingold:
> How do I check whether any of 1292, 1294 or 1298 is an array subtype definition?

Get_Kind is available for any node of the AST.  So call it on the subtype_indication to distinguish between a simple_name and an array_subtype_definition.

umarcor:
A little step forward, and some questions:

- I used `pyutils.chain_iter` to iterate over the chain of ports, instead of using a `while` loop.
- I could tell `std_logic` (Simple_Name) apart, so that error is solved.
- `Tm_El = nodes.Get_Element_Subtype(mark)` fails with `raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : no field Element_Subtype` and `mark_state = nodes.Get_Constraint_State(mark)` fails with `raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : no field Constraint_State`. Hence, I cannot evaluate any of the following:

```py
has_own_element_subtype = Def_El != Tm_El

if not has_index and not has_own_element_subtype:
   return "EMPTY"

mark_state = nodes.Get_Constraint_State(mark)
if mark_state != nodes.Iir_Constraint.Fully_Constrained and has_index:
    Disp_Array_Sub_Definition_Indexes (Ctxt, Def)

if has_own_element_subtype and (nodes.Get_Kind(Def_El) in nodes.Iir_Kinds.Composite_Type_Definition):
     Def_El_kind = nodes.Get_Kind(Def_El)
     if Def_El_kind == Iir_Kind.Record_Subtype_Definition:
         #Disp_Record_Element_Constraint (Ctxt, Def)
     if Def_El_kind == Iir_Kind.Array_Subtype_Definition:
         #Disp_Array_Element_Constraint (Ctxt, Def, Type_Mark);
     raise Exception("disp_element_constraint")
```

In this case, I know it is not a record type, and using `Disp_Array_Element_Constraint` again feels like a circular path. As a result, I assume the correct way is `Disp_Array_Sub_Definition_Indexes`:

```py
        clist = nodes.Get_Index_Constraint_List(subtype)

        for rng in pyutils.flist_iter(clist):
            #> Disp_Discrete_Range
            #> Disp_Range
            if nodes.Get_Kind(rng) == nodes.Iir_Kind.Range_Expression:
                print(
                    nodes.Get_Direction(rng),  # Dir_To
                    nodes.Get_Range_Origin(rng),
                    nodes.Get_Left_Limit(rng),
                    nodes.Get_Right_Limit(rng),
                    nodes.Get_Left_Limit_Expr(rng),
                    nodes.Get_Right_Limit_Expr(rng),
                )
            # if ...
```

- `Get_Direction` returns `1`. In Ada, it is compared to `Dir_To`. But it seems not to be defined in `nodes.py`.
- `Get_Left_Limit` and `Get_Right_Limit` return `0`, but the `*_Expr` calls return apparently valid handlers:

```py
· demo.vhdl
  - entity e1
    * clk in std_logic
    * rst in std_logic
mark: 1296 Simple_Name
constraint: 1
1 0 0 0 1297 1301
    * q out array_subtype_definition
  - architecture behav of e1
```

I guess I need to get from `1297` to `7` and from `1301` to `0`, since both of them are of kind `Integer_Literal`. How to do that?

Overall, I feel I am rewriting Ada functions/procedures in Python, without any added value. I implemented the parts of the functions that lead to my example, but all other paths should also be covered (in Python). That does not seem correct. Would it make sense to rewrite the procedures in `vhdl-prints.adb` as functions? Instead of printing the content directly, the caller would take care of the resulting string.

tgingold:
The file `vhdl-nodes.ads` list all the nodes and which methods can be called on a node.
So for Integer_Literal, you can call `Get_Value` for the value.

In `vhdl_langserver/document.py`, the function `format_range` indirectly calls printers to reformat sources.

Yes, `Dir_To` and `Dir_Downto` should be declared in `libghd/thin/types.py` (to be created) like they are in `types.ads`.

umarcor:
> So for Integer_Literal, you can call `Get_Value` for the value.

Thanks. With this and some more tinkering, I think this PR is ready to be merged now. Not ideal, but good enough:

```
dir units01:
entity e1
architecture behav

· demo.vhdl
  - entity e1
    * clk in std_logic
    * rst in std_logic
    * q out std_logic_vector(7 downto 0)
  - architecture behav of e1
Test successful
tests are successful
```

---

>  * `Tm_El = nodes.Get_Element_Subtype(mark)` fails with `raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : no field Element_Subtype` and `mark_state = nodes.Get_Constraint_State(mark)` fails with `raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : no field Constraint_State`.

Since I did not understand these yet, I removed them from the PR.

---

>  * `Get_Left_Limit` and `Get_Right_Limit` return `0`, but the `*_Expr` calls return apparently valid handlers

@tgingold, why is this?

> In `vhdl_langserver/document.py`, the function `format_range` indirectly calls printers to reformat sources.

I checked it, but I'm not sure about `self._tree`, `first_line` and `last_line` in:

```py
    hand = formatters.Allocate_Handle()
    formatters.Indent_String(self._tree, hand, first_line, last_line)
    res = [
        {
           "range": {
                "start": {"line": first_line - 1, "character": 0},
                "end": {"line": last_line, "character": 0},
            },
            "newText": formatters.Get_C_String(hand)[:formatters.Get_Length(hand)].decode(Document.encoding),
        }
    ]
    formatters.Free_Handle(hand)
```

> Yes, `Dir_To` and `Dir_Downto` should be declared in `libghd/thin/types.py` (to be created) like they are in `types.ads`.

Since it's an enumeration (https://github.com/ghdl/ghdl/blob/master/src/types.ads#L185), should it be as follows?

```py
from enum import Enum
class Dir(Enum):
     To = 0
     Downto = 1
```

tgingold:
> I checked it, but I'm not sure about self._tree, first_line and last_line in:

_tree is the node of the design file (iirc, but it could also work with the design
unit).  first_line/last_line is the range to be reformatted.
Maybe a new interface could be added to reprint any node.

> Since it's an enumeration (https://github.com/ghdl/ghdl/blob/master/src/types.ads#L185), should it be as follows?

Yes.  There is also an automatic way to generate python enumerations in xtools/pnodespy.py

umarcor:
> Maybe a new interface could be added to reprint any node.

I think we should return a class for each node, and each class should have a "to_string" or "print_str" function that returns a reprint. However, I believe that the print functions are not relevant, just for debugging. The important piece is the class, which would allow users to access fields in a pythonic way. Then, anyone can extend the class to add a custom printing function. Yet, implementing those classes needs some more thinking than I can currently afford.

umarcor:
Two commits of this PR were are merged already, and it now contains a single commit related to https://github.com/ghdl/ghdl/issues/1437#issuecomment-682577798. Let's keep it as a placeholder to work on that and for the more pythonic solution discussed above.

umarcor:
After the hard work during these last months, the purpose of this PR is now superseeded by pyGHDL.dom. Thanks @Paebbels, @tgingold! ❤️ 

For the record, pyGHDL.dom is used in the following demos:

- Open Source VHDL Design Explorer (OSVDE): https://umarcor.github.io/osvb/apis/project/OSVDE.html
- pyVHDLModelUtils: https://umarcor.github.io/osvb/apis/project/pyVHDLModelUtils.html

PR #1451, opened 2020-08-29
update for LLVM 11

sharkcz:
We need this in Fedora, because we already have the LLVM 11 release candidates available in the distro.


tgingold:
Sure & thanks!

PR #1306, opened 2020-05-12
WIP: Add some documentation on the RTI structures.

benreynwar:
I created some documentation on the RTI nodes mostly for my own benefit.  I'm hoping that as I add more comments to the code, this document will be become unnecessary.

Issue #506, opened 2017-12-22
Landing site for GHDL


1138-4EB:
[Demo](https://1138-4eb.github.io/ghdl/) available.

---

- The enhancement of the documentation which started almost a year ago has been recently merged to master (see #482).
- The migration from tgingold/ghdl to the new organization is almost done (#500). GitHub, Travis and RTD are already fixed. Only AppVeyor is to be updated (see [500#issuecomment-353274561](https://github.com/ghdl/ghdl/issues/500#issuecomment-353274561)).
- Travis stages have been recently introduced in the build system (see #489). Also, dockerhub is integrated in the main repo, although credentials need to be configured, yet.

This issue is an attempt to follow these efforts towards making GHDL easier to start with for new users. More precisely, it is a proposal to replace [ghdl.free.fr](http://ghdl.free.fr/), which is mostly deprecated, but has some interesting section 'concepts': http://ghdl.free.fr/site/pmwiki.php?n=Main.Download

This is just a proof of concept to put some content in `ghdl.github.io`. To do so, we would create a new repo in the organization (`ghdl/ghdl.github.io`). Then, we can automate the deployment in travis by adding a (last) stage to the build. In fact, having an organization makes it safer to automate the deployment without compromising the main repo, as explained in [#477](https://github.com/ghdl/ghdl/issues/477#project).

Related to the URL, if @tgingold is okay with it, `ghdl.github.io` can be a redirected to `ghdl.free.fr`, or the other way. Indeed, we can use an external domain but keep the page hosted at GitHub. If not, they can be just separate sites.

---

The [proposed site](https://1138-4eb.github.io/ghdl/) is built from Markdown and reStructuredText sources with [hugo](https://gohugo.io/), as explained in [#477](https://github.com/ghdl/ghdl/issues/477#documentation). Indeed, the structure explained there can be seen in the demo site. Mostly md sources are saved in `ghdl-io`, while mostly rst sources are kept in `doc`.

I wrote the template from scratch with [tachyons](http://tachyons.io/). This means it is not feature-complete, but is is lightweight and simple (no social links, no google-analytics, etc.). The landing page is almost finished. The rest of the pages still need some work (i.e. code highlighting).

Some interesting features:

- The whole site is built in 1.35s in a docker container. A local server can be started to edit the sources and livereload the site.
- The commit and date of the build are shown in the footer.
- The date of the last modification of each file is extracted from git.
[...]

PR #1452, opened 2020-08-31
Documentation: add RTI, add python, use BTD

umarcor:
Ref #1445 #506

This PR includes three modifications related to the docs:

1. The info about RTI contributed in #1306 was not added to the index, so it was unpublished. Now, it is shown in section 'Internals'.

2. Python modules are added to the docs using `autodoc`. A section named 'Python Interface' is added to 'GHDL Usage'. In the main section, `libghdl` is documented. Then, three subsections are added: `libghdl.thin`, `libghdl.thin.vhdl` and `libghdl.thin.vhdl.pyutils`. Note that the packages are shown as siblings, to avoid too many hierarchy levels.

3. Using `autodoc` requires Python `libghdl` package, and that requires the shared library. Hence, in order to document the Python interface, Sphinx needs to be executed in an environment where GHDL and the Python packages are installed. Instead of trying to install those in ReadTheDocs, the proposal in this PR is to migrate to BuildTheDocs and GitHub Pages. That allows to use `ghdl/vunit:llvm` as the base image to run Sphinx.

Migration from ReadTheDocs to GitHub Pages has some implications:

- The theme is slightly changed. However, the "new" one is the same used in [ghdl.github.io/ghdl-cosim](https://ghdl.github.io/ghdl-cosim).
- The location is [ghdl.github.io/ghdl](https://ghdl.github.io/ghdl), instead of [ghdl.readthedocs.io](https://ghdl.readthedocs.io). Optionally, the content might be pushed to [ghdl.github.io](https://ghdl.github.io), but I think it is better to reserve that for a landing page (#506).
- The theme/build procedure based on BuiltTheDocs and GitHub Pages does not support hosting multiple versions of the docs automatically (yet). Nonetheless, I believe it is in the roadmap, and the stable docs of GHDL will still be available in RTD.
- These changes should not break RTD. However, all the sections in 'Python Interface' will be empty (due to libghdl not being available).

There is a demo of this proposal available: [umarcor.github.io/ghdl](https://umarcor.github.io/ghdl/).

In fact, the most significant advantage of using BTD and GitHub Pages is that contributors will no longer need to register/setup an RTD account in order to test their PRs. Pushing to their fork will update `USERNAME.github.io/ghdl` automatically.

---

@tgingold, three items from `libghdl.thin.vhdl.canon` and `libghdl.thin.vhdl.iee` seem to be undefined in the shared library and produce errors when running autodoc:

- `vhdl__canon__canon_extract_sequential_statement_chain_sensitivity`
- `vhdl__ieee__std_logic_1164__rising_edge`
- `vhdl__ieee__std_logic_1164__falling_edge`

I commented them out. See https://github.com/ghdl/ghdl/pull/1452/files#diff-620ea71d748508d4fd04184e0450cc07 and https://github.com/ghdl/ghdl/pull/1452/files#diff-12a47c36ad701763ecf086132121a415.


tgingold:
Thanks!

PR #1459, opened 2020-09-08
termux: update gnat, install llvm and make explicitly

### LINKED ISSUE #1458: Termux build script not working with current version.
**Description**
Termux build script does not work, because configure reports missing gnat9.

**Possible Fix**
I had to manually install:
pkg install libgnat-10
pkg install llvm
pkg install make

This should be added into termux.sh

umarcor:
Fix #1458

I cannot test it in a "clean" environment, but it works in my existing termux installation.

/cc @BenjaminRenz

### DIFF
--- a/dist/termux.sh
+++ b/dist/termux.sh
@@ -3,8 +3,8 @@
 cd $(dirname "$0")/..
 
 curl -fsSL https://its-pointless.github.io/setup-pointless-repo.sh | bash -
-pkg install gnat-9
-setupgcc-9
+pkg install gnat-10 llvm make
+setupgcc-10
 
 mkdir -p build-termux
 cd build-termux
--- a/dist/termux.sh
+++ b/dist/termux.sh
@@ -6,6 +6,9 @@ curl -fsSL https://its-pointless.github.io/setup-pointless-repo.sh | bash -
 pkg install gnat-10 llvm make
 setupgcc-10
 
+# Temporal fix. See https://github.com/its-pointless/gcc_termux/issues/100
+ln -s $PREFIX/bin/gnatmake-10 $PREFIX/bin/gnatmake
+
 mkdir -p build-termux
 cd build-termux
 ../configure --default-pic --enable-synth --with-llvm-config=llvm-config --prefix="$PREFIX"


BenjaminRenz:
Thanks @umarcor for the fast reply...

I have tested it on fresh installation but had to change the file to:
```
pkg install libgnat-10 gnat-9 llvm make clang
setupgcc-9
```
unfortunatelly after it compiles for a minute I get linker errors with this version, see attached log file, which was generated with ./termux.sh 2> log.txt
[log.txt](https://github.com/ghdl/ghdl/files/5190070/log.txt)

Could this be caused by version missmatches between gnat-9 and libgnat-10?
Not really sure what I'm doing wrong...



With the current version in your merge I get the output:
```
gcc now points to gcc-10
Sorry, you need GNAT to build GHDL. See README
(gnatmake executable is: gnatmake)
```

umarcor:
> 
> 
> Thanks @umarcor for the fast reply...
> 
> I have tested it on fresh installation but had to change the file to:
> 
> ```
> pkg install libgnat-10 gnat-9 llvm make clang
> setupgcc-9
> ```

My suggestion is NOT to mix versions of GNAT. That's likely to fail. `gnat` is the ADA frontend for GCC. `libgnat` are the runtime libraries. Hence, if you build a tool with `gnat-9`, it is expected to use `libgnat-9`.

> unfortunatelly after it compiles for a minute I get linker errors with this version, see attached log file, which was generated with ./termux.sh 2> log.txt
> [log.txt](https://github.com/ghdl/ghdl/files/5190070/log.txt)
> 
> Could this be caused by version missmatches between gnat-9 and libgnat-10?

Yes.

> With the current version in your merge I get the output:
> 
> ```
> gcc now points to gcc-10
> Sorry, you need GNAT to build GHDL. See README
> (gnatmake executable is: gnatmake)
> ```

I was confused and I changed `gnat-9` to `libgnat-10`. It needs to be `gnat-10`.

The reason it is failing is likely a regression in its-pointless' repo. Precisely, `gnatmake` does NOT exist. It is named `gnatmake-10`. Hence, a symbolic link needs to be created:

```sh
ln -s $PREFIX/bin/gnatmake-10 $PREFIX/bin/gnatmake
```

Ref its-pointless/gcc_termux#84 #1173

BenjaminRenz:
Oh, I see...
For it to work starting from a blank termux installation I also had to add:
```
ln -s $PREFIX/bin/gnatbind-10 $PREFIX/bin/gnatbind
ln -s $PREFIX/bin/gnatlink-10 $PREFIX/bin/gnatlink
```
which were not present.

But now I'm getting errors in ioctl.h.
I've once again attached the log
[log(1).txt](https://github.com/ghdl/ghdl/files/5199172/log.1.txt)
please ignore the first three lines, which occur because I've run the script twice and the links are already in place.

There seem to be two definitions, one signed and one unsigned inside '/data/data/com.termux/files/usr/include/bits/ioctl.h' which according to the comment next to the code get selected by BIONIC_IOCTL_NO_SIGNEDNESS_OVERLOAD
(see https://android.googlesource.com/platform/bionic/+/master/libc/include/bits/ioctl.h).

so would it be a solution to pass this define as an compiler option?
like they seem to do here:
https://github.com/mono/mono/pull/15650

Since this is probably not direcly related to ghdl I would consider to raise an issue at the repository causing this, but I'm not sure if its caused by the it's pointless repo or ndk or something else.
Thank you for your patience 😃

tgingold:
> The reason it is failing is likely a regression in its-pointless' repo. Precisely, gnatmake does NOT exist. It is named gnatmake-10. Hence, a symbolic link needs to be created:
> 
> ln -s $PREFIX/bin/gnatmake-10 $PREFIX/bin/gnatmake
> Ref its-pointless/gcc_termux#84 #1173

You can also use: `make GNATMAKE=gnatmake-10`

umarcor:
> For it to work starting from a blank termux installation I also had to add:
> 
> ```
> ln -s $PREFIX/bin/gnatbind-10 $PREFIX/bin/gnatbind
> ln -s $PREFIX/bin/gnatlink-10 $PREFIX/bin/gnatlink
> ```
> 
> which were not present.

Would you mind mentioning it in its-pointless/gcc_termux#100?

> There seem to be two definitions, one signed and one unsigned inside '/data/data/com.termux/files/usr/include/bits/ioctl.h' which according to the comment next to the code get selected by BIONIC_IOCTL_NO_SIGNEDNESS_OVERLOAD
> (see https://android.googlesource.com/platform/bionic/+/master/libc/include/bits/ioctl.h).
> 
> so would it be a solution to pass this define as an compiler option?
> like they seem to do here:
> [mono/mono#15650](https://github.com/mono/mono/pull/15650)

Honestly, I have no guess about this one. You might want to try with `gnat-9` and `setupgcc-9`. If that works (which should), you will be able to use GHDL, while we fix the issue with `gnat-10`.

> Since this is probably not direcly related to ghdl I would consider to raise an issue at the repository causing this, but I'm not sure if its caused by the it's pointless repo or ndk or something else.

I would first ask it's pointless.

BenjaminRenz:
Hi,

> Honestly, I have no guess about this one. You might want to try with gnat-9 and setupgcc-9. If that works (which should), you will be able to use GHDL, while we fix the issue with gnat-10.

unfortunatelly I already tried that (with gnat-9 and libgnat-9 installed):

I get the error 'gnat not found, you need gnat to build ghdl'.
If I call gnat-9 directly it errors out and prints:
`CANNOT LINK EXECUTABLE "gnat-9", library "libgnat-10.so not found"`
which is why my original idea was to use libgnag-10 with gnat-9.😕 

> but it works in my existing termux installation.

Can you check if you have libgnat-10 installed?
For
`pkg list-installed | grep gnat*` 
I get gnat-9 and libgnat-9 only.

BenjaminRenz:
Based on answers from it's pointless I now arrived at a working termux,sh file:
[termux.txt](https://github.com/ghdl/ghdl/files/5213696/termux.txt)
the missing softlinks seem to be fixed in the it's pointless repository.
My only remaining problem is that I run out of RAM (2GB) when ghdl builds it's standard library, but I guess there is not much I can do about it, because my kernel does not support swap memory.

umarcor:
> unfortunatelly I already tried that (with gnat-9 and libgnat-9 installed):
> 
> I get the error 'gnat not found, you need gnat to build ghdl'.
> If I call gnat-9 directly it errors out and prints:
> `CANNOT LINK EXECUTABLE "gnat-9", library "libgnat-10.so not found"`
> which is why my original idea was to use libgnag-10 with gnat-9.😕

That seems like some packages where mixed...  I got different errors, but not that one...

I can now reproduce the issue with `ioctl.h` in a clean environment with gnat-10.

For building with gnat-9 in a clean environment, I had to do `ln -s $PREFIX/lib/libgnat-9.so $PREFIX/lib/libgnat.so`. Hence, that seems a regression of its-pointless/gcc_termux#84. Anyway, then it fails with:

```sh
/data/data/com.termux/files/usr/include/sys/cdefs.h:332:18 error: missing binary operator before token "("
 332 | #if __has_builtin(__builtin_umul_overflow) || __GNUC__ >= 5
        |
make: *** :/src/grt/Makefile.inc:183: grt-cstdio.o] Error 1
```

However, on a different device, where I had already built it before:

> > but it works in my existing termux installation.
> 
> Can you check if you have libgnat-10 installed?
> For
> `pkg list-installed | grep gnat*`
> I get gnat-9 and libgnat-9 only.

Both of them are installed. Interestingly:

```
ls -l $PREFIX/lib/libgnat*
/data/data/com.termux/files/usr/lib/libgnat-10.so -> /data/data/com.termux/files/usr/lib/gcc/aarch64-linux-android/10.2.0/adalib/libgnat-10.so
/data/data/com.termux/files/usr/lib/libgnat-9.so -> /data/data/com.termux/files/usr/lib/gcc/aarch64-linux-android/9.3.0/adalib/libgnat-9.so
/data/data/com.termux/files/usr/lib/libgnat.so -> /data/data/com.termux/files/usr/lib/gcc/aarch64-linux-android/10.2.0/adalib/libgnat.so
```

and 

```
ls -l $PREFIX/lib/lib/gcc/aarch64-linux-android/10.2.0/adalib/libgnat*
lignat-so -> libgnat-10.so
```

Nevertheless, it now fails because of `ioctl.h`. So, at least it's consistent.

> Based on answers from it's pointless I now arrived at a working termux.sh file:
> [termux.txt](https://github.com/ghdl/ghdl/files/5213696/termux.txt)
> the missing softlinks seem to be fixed in the it's pointless repository.

Thanks! I had found that `clang` needs to be installed explicitly now. However, I was not aware of `setup-pathforgcc`. That fixed the issue, both on the existing and on the clean devices!

> My only remaining problem is that I run out of RAM (2GB) when ghdl builds it's standard library, but I guess there is not much I can do about it, because my kernel does not support swap memory

That's unfortunate... I only tried it on relatively powerful devices (Redmi 5 and 7).

BenjaminRenz:
> Nevertheless, it now fails because of `ioctl.h`. So, at least it's consistent.

Yeah it's pointless answer implies that "setup-pathforgcc" also patches the ioctl.h header.
https://github.com/its-pointless/gcc_termux/issues/102

> > My only remaining problem is that I run out of RAM (2GB) when ghdl builds it's standard library, but I guess there is not much I can do about it, because my kernel does not support swap memory
> 
> That's unfortunate... I only tried it on relatively powerful devices (Redmi 5 and 7).

I also tested if on an Samsung A70, which unfortunatelly is not mine :smile:, with 6GB of ram and it seems to work there. Maybe I can copy the  standard library files from there over to my device. But on the other hand it could also then be insufficient ram for operating ghdl normally, I will need to test that, but my plan was to use if for simple simulation only.

umarcor:
@BenjaminRenz I would not expect memory issues. GHDL is mostly CPU bound. See https://dbhi.github.io/pdf/FPGA2020_poster.pdf. The three graphs in the first row are execution time, the second row is RAM usage and each column is a different hardware design. Look at green and black bars only. As you can see, memory usage is quite stable regardless of the device.

In fact, the last column corresponds to a VexRiscv SoC generated from SpinalHDL. There is a significant runtime difference dure to maximum CPU frequency of each "computer", but memory usage varies 5MB at most.

The design in column 2 is a filter for hyperspectral image cubes, which are huge. That's why it takes more than 400MB of RAM. Still, that should work flawlessly on your 2GB device.

Last, but not least, the testsuite in this repo should work. It makes more than 40min on my Redmi 5. But it is an interesting benchmark to see how usable (responsive) it is. I'd say that includes the synthesis testsuite too (which is much slower).

Issue #1204, opened 2020-04-09
Revisit CLI syntax


eine:
Currently, the architecture of the CLI is based on (sub)commands and options. Commands are expected to be provided as the first argument, and do effectively decide which feature is to be used. OTOH, options provide parameters to the commands. However, there is no syntactical difference, as both commands and options start with `--` or `-i`. As a result, we rely on properly formating `--help` and on the documentation, to ensure that users understand the usage.

I think it might be easier to explain and to understand if (sub)commands were supported without `--` or `-`. In order to avoid a breaking change, I think that this should be added while the current behaviour is kept. Optionally, single letters can be capitalised.

At the same time, executing `ghdl` only produces `ghdl:error: missing command, try ghdl --help`. It might be useful if it called a subcommand by defaut instead (either `--help` or `-a`).

tgingold:
I am OK with allowing commands without - or -- like many tools.

However, if there is no command, the default can only be --help.

I don't think we should allow one letter commands.  So `-a` could be an alias of `analyze`.

tgingold:
So how to make progress on this issue ?  Do you want me to add all the commands ?

eine:
If it is not much effort, please do. I started having a look at it, but I'm finding it difficult to understand where are commands defined. For example https://github.com/ghdl/ghdl/blob/master/src/ghdldrv/ghdlcomp.adb#L127 and https://github.com/ghdl/ghdl/blob/master/src/ghdldrv/ghdldrv.adb#L1244 feel like a duplication, but `ghdl --help` shows the second one only.

tgingold:
ghdlcomp is for mcode while ghdldrv is for gcc/llvm; hence the overlap.

eine:
Coming from [#1362](https://github.com/ghdl/ghdl/issues/1362#issuecomment-641423542), perhaps we should consider adding `--sim` or `--simulate` as the "long form" of `-c`. That would make it more explicit that both are the "higher-level" (single command) approaches, while other commands are for more granular control. We might update or extend the Quick Start guide accordingly.

/cc @ozbenh

tgingold:
I am not sure the long form of "-c" is "--simulate".  Because there are two forms of "-c":
"-c .. -e" compile and elaborate
"-c .. -r" compilte, elaborate and runs.

Apparently the second one is not available with gcc/llvm backends.

eine:
[...]

PR #1467, opened 2020-09-17
Revisit CLI syntax

umarcor:
This PR implements the initial proposal in #1204. Commands without `--` or `-` are supported, and others are defined as aliases. The help is modified accordingly.

I modified the testsuite for printing the help at the end. As a result, the proposed new layout can be seen in the tests below.

There are two sets of commands which are defined differently, and I didn't know how define aliases for them, so I left them untouched (the layout changed only). Those are `--libghdl-*` and `--vpi-*` commands, defined/registered in `ghdlvpi.adb` and `ghdlsynth.adb`, respectively.

Apart from that, there are some commands which have the same prefix. For example `disp-config` and `disp-std`, or `gen-makefile` and `gen-depends`. The last two are close, but `disp-*` are far from each other. I tried to understand how to change the order, but it seems not straightforward because commands are shown in the same order as registered. @tgingold what do you think about displaying commands alphabetically? Alternatively, I thought about making subcommands (`ghdl disp config`, `ghdl disp std`, etc.), but I think it is not worth (not yet, at least).

### DIFF
--- a/src/ghdldrv/ghdlcomp.adb
+++ b/src/ghdldrv/ghdlcomp.adb
@@ -116,14 +116,20 @@ package body Ghdlcomp is
    is
       pragma Unreferenced (Cmd);
    begin
-      return Name = "-r" or Name = "--elab-run";
+      return
+        Name = "elab-run" or else
+        Name = "--elab-run" or else
+        Name = "-r" or else
+        Name = "run";
    end Decode_Command;
 
    function Get_Short_Help (Cmd : Command_Run) return String
    is
       pragma Unreferenced (Cmd);
    begin
-      return "-r,--elab-run [OPTS] UNIT [ARCH] [RUNOPTS]  Run UNIT";
+      return "elab-run [OPTS] UNIT [ARCH] [RUNOPTS]"
+        & ASCII.LF & "  Elaborate and run design UNIT"
+        & ASCII.LF & "  aliases: --elab-run, -r, run";
    end Get_Short_Help;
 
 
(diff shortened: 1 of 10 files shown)


tgingold:
Don't you need to also change the documentation ?

I would also not add "dispconfig" as the name is too weird.  Likewise for "bootstrap-std" and "bug-box" which is are not user options.

For "disp", maybe we should create a sub-list.
Likewise for "help": I think that the current commands (--options-help, --run-help) are confusing and should be replaced by "help options" and "help run".  (Humm, wouldn't "help run" be ambiguous ?)

umarcor:
> Don't you need to also change the documentation ?

Not strictly. This PR does not introduce any backwards incompatible change. All previous command names can still be used. I want to update the docs, but I want to do it after we discuss the details (below). I have planned it in four steps:

1. This PR.
2. Make "no command" default to "help".
3. Provide `ghdl sim|synth FILEs -e UNIT ...`  as discussed in #1204.
4. Update the docs. I think that these changes are specially relevant for the User Guide.

> I would also not add "dispconfig" as the name is too weird. Likewise for "bootstrap-std" and "bug-box" which is are not user options.

I did it for consistency. My proposal is to transition towards not having any command starting with `-` or `--`. I believe that commands/subcommands should be plain words, while options should have preceding hyphens. That allows users to know the nature of the keywords, by just looking at them.

I understand that dispconfig, bootstrap-std and bug-box might feel weird, but I think that's because those are mostly internal commands, not because of the naming. Nevertheless, if the long hypehanted forms exist, the short forms should exist too, for consistency. What I would propose is adding a different mark to tell these internal commands apart:

- `--disp-config`, `--dispconfig`, `_disp-config` and `_dispconfig`.
- `--bootstrap-standard` and `_bootstrap-std`.
- `--bug-box` and `_bug-box`.

> For "disp", maybe we should create a sub-list.

I believe it would be nice to do it for `disp` and for `gen` too. However, I'm far from being able to implement it myself. Since the alphabetic ordering might partially mitigate the issue, that's what I commented above.

> Likewise for "help": I think that the current commands (--options-help, --run-help) are confusing and should be replaced by "help options" and "help run". (Humm, wouldn't "help run" be ambiguous ?)

I am familiar with spf13/cobra, where any command has a help subcommand. Hence, in this case it would be `ghdl analyse help`, `ghdl elaborate help`, `ghdl run help`, etc. The issue is that it would prevent e.g. using design units named precisely `help`.

As a result, I agree with you, and I think in GHDL we should use `help` as a submenu, and put everything in there:

- `ghdl help analyse` (would include opts)
- `ghdl help elaborate` (would include opts)
- `ghdl help run` (would include runopts)
- `ghdl help opts`
- `ghdl help runopts`
- `ghdl help sim` (would include opts and runopts)
- `ghdl help synth`

My question is: do we want to preserve `--options-help` and `--run-help` or do we want to deprecate those?

tgingold:
Thanks.  I will create the 'disp' sub command.

Right now, you can already do: `ghdl help -a` or now even `ghdl help analyze`.

PR #1487, opened 2020-10-06
PSL doc: only clocked directives are supported

tmeissner:
Small update to PSL documentation adding a hint that all directives has to be clocked. Not only `assert` as  before.


eine:
Thanks!

Issue #1414, opened 2020-07-28
PSL assert statement sometimes misparsed as cover statement


rlee287:
**Description**
PSL assert statements sometimes misparse as a cover statement.

**Expected behaviour**
No cover statements end up in the synthesis output.

**How to reproduce?**
Cover statements end up in the synthesis output despite there being no PSL cover statements in the source:

```vhd :file: cover_nonoverlap_misparse.vhd
library IEEE;
use IEEE.STD_LOGIC_1164.ALL;
use IEEE.NUMERIC_STD.ALL;

entity cover_nonoverlap_misparse is
    Port (
        clk: in STD_LOGIC;
        counter_val: out STD_LOGIC_VECTOR(3 downto 0)
    );
end entity;

architecture Behavioral of cover_nonoverlap_misparse is
    signal ctr_internal: UNSIGNED(3 downto 0) := x"0";
begin
    process(clk) is
    begin
        ctr_internal <= ctr_internal + 1;
    end process;
    counter_val <= STD_LOGIC_VECTOR(ctr_internal);

    -- psl default clock is rising_edge(clk);
    --psl assert always ctr_internal = 0 |=> ctr_internal = 1;
end Behavioral;
```
```vhd :file: cover_overlap_misparse.vhd
library IEEE;
use IEEE.STD_LOGIC_1164.ALL;
use IEEE.NUMERIC_STD.ALL;

entity cover_overlap_misparse is
    Port (
        clk: in STD_LOGIC;
        counter_val: out STD_LOGIC_VECTOR(3 downto 0)
    );
end entity;

architecture Behavioral of cover_overlap_misparse is
    signal ctr_internal: UNSIGNED(3 downto 0) := x"0";
begin
    process(clk) is
    begin
        ctr_internal <= ctr_internal + 1;
    end process;
    counter_val <= STD_LOGIC_VECTOR(ctr_internal);

    -- psl default clock is rising_edge(clk);
    --psl assert always ctr_internal = 0 |-> ctr_internal = 0;
end Behavioral;
```
```sh :image: ghdl/ghdl:buster-mcode
ghdl --synth -fpsl cover_nonoverlap_misparse.vhd -e
ghdl --synth -fpsl cover_overlap_misparse.vhd -e
```

Synthesis results for `cover_nonoverlap_misparse.vhd` (has an `assert_cover` comment):

```vhdl
library IEEE;
use IEEE.std_logic_1164.all;
use IEEE.numeric_std.all;
[...]

PR #1488, opened 2020-10-06
Synth doc: add infos about synth specific options

tmeissner:
Adding documentation of synthesis specific options in the Synthesis document. I don't have checked how it looks in the RTD, as I don't remember how I could.

### DIFF
--- a/doc/using/Synthesis.rst
+++ b/doc/using/Synthesis.rst
@@ -58,6 +58,50 @@ See :ghdlsharp:`1174` for on-going discussion about other output formats.
 .. TIP::
    Files can be provided in any order.
 
+.. _synthesis_options:
+
+Synthesis options
+==================
+
+In addition to analyze options, there are some synthesis specific options.
+
+.. option:: -gNAME=VALUE
+
+  Override top unit generic `NAME` with value `VALUE`. Similar to the run-time option :option:`-gGENERIC=VALUE`.
+
+  Example::
+
+    $ ghdl --synth --std=08 -gDEPTH=12 my_unit
+
+.. option:: --vendor-library=NAME
+
+  Any unit from library NAME is a black box.
+
+  Example::
+
+    $ ghdl --synth --std=08 --vendor-library=vendorlib my_unit
+
+.. option:: --no-formal
+
+  Neither synthesize assert nor PSL.
+
+  Example::
+
+    $ ghdl --synth --std=08 --no-formal my_unit
+
+.. option:: --no-assert-cover
+
+  Disable automatic cover PSL assertion activation. If this option isn't used, GHDL generates
+  `cover` directives for each `assert` directive automatically during synthesis.
+
+  Example::
+
+    $ ghdl --synth --std=08 --no-assert-cover my_unit
+
+.. TIP::
+  Furthermore there are lot of debug options available. Beware: These debug options should only used
+  for debugging purposes as they aren't guaranteed to be stable during development of GHDL's synthesis feature. You can find them in the file `src/ghdldrv/ghdlsynth.adb <https://github.com/ghdl/ghdl/blob/master/src/ghdldrv/ghdlsynth.adb>`_ in the procedure `Decode_Option()`.
+
 .. _Synth:plugin:
 
 Yosys plugin


doc/using/Synthesis.rst
```diff
@@ -58,6 +58,37 @@ See :ghdlsharp:`1174` for on-going discussion about other output formats.
...
    Files can be provided in any order.
 
+.. _synthesis_options:
+
+Synthesis options
+==================
+
+In addition to analyze options, you can use following synthesis specific options:
+
+.. option:: -gNAME=VALUE
```
eine, line 68: This is a run option too: https://ghdl.github.io/ghdl/using/Simulation.html#cmdoption-ghdl-gGENERIC

On the one hand, I don't know if the prototype should be the same ( `-gGENERIC=VALUE` vs `-gNAME=VALUE`). It might produce a collision.

On the other hand, I don't know if the comments/notes about this option changing to an analysis option in the future applies to synthesis too.

In any case, I believe a link/ref should exist from this section to the other one. The types and syntax which are supported are likely the same in both cases.

doc/using/Synthesis.rst
```diff
@@ -58,6 +58,37 @@ See :ghdlsharp:`1174` for on-going discussion about other output formats.
 .. TIP::
    Files can be provided in any order.
 
+.. _synthesis_options:
+
+Synthesis options
+==================
+
+In addition to analyze options, you can use following synthesis specific options:
```
eine, line 66: Since "additional" synthesis options are explained, might be worth adding a reference to "debugging options for synthesis": https://github.com/ghdl/ghdl/blob/8d5ae7bf63c616925306b42d42d01d1cebf3ffc7/src/ghdldrv/ghdlsynth.adb#L197-L214

Instead of documenting each of them, I'd just suggest users to find them in `src/ghdldrv/ghdlsynth.adb`.

doc/using/Synthesis.rst
```diff
@@ -58,6 +58,37 @@ See :ghdlsharp:`1174` for on-going discussion about other output formats.
...
+    $ ghdl --synth --std=08 --vendor-library=vendorlib my_unit
+
+.. option:: --no-formal
+
+  Neither synthesize assert nor PSL.
+
+  Example::
+
+    $ ghdl --synth --std=08 --no-formal my_unit
+
```
eine, line 91: Would `--no-assert-cover` fit here?

https://github.com/ghdl/ghdl/blob/267ceb3b009f4fa2ccecbfd7ce0f822daa2e2541/src/ghdldrv/ghdlsynth.adb#L129-L130

eine (commented):
Thanks for adding these! Let me know what you think about debugging options and `no-assert-cover`. I can take care of adding them if you wish, but I'm not familiar.

tmeissner:
Yeah, I only took the options that was added in the commit this PR is related. I will add the other ones also.

eine:
I overlooked that. This is the other commit: https://github.com/ghdl/ghdl/commit/9efb27060dc5ca23493fb3d3aab6d455049a5790. Anyway, let's wait for @tgingold, because I'm not sure about the description of `no-assert-cover`. It feels to me that it should say "deactivation" instead of "activation".

tmeissner:
Yep. I'm pretty sure that `--no-assert-cover` deactivates the automatic generation of `cover`directives for each `assert` directive, as the commit fixed issue #1414 which mentioned that.

doc/using/Synthesis.rst
tmeissner: I added a link to the run option.

doc/using/Synthesis.rst
tmeissner: I added a hint with some explanation and a link to the ghdlsynth.adb file with name of the relevant procedure.

doc/using/Synthesis.rst
tmeissner: I added documentation of these option as I understand it.

doc/using/Synthesis.rst
tmeissner: Maybe you could check the link, as it isn't generating a link but only text?

tgingold:
The description is correct, so I have merged it.

Thank you for updating the doc.

doc/using/Synthesis.rst
eine: I fixed it: https://github.com/ghdl/ghdl/commit/1b23af06d9806867887aebad36b71c26e10584c1. The key needs to be `-gGENERIC`.

doc/using/Synthesis.rst
eine: There is a shortcut for referencing sources in the repo. See https://github.com/ghdl/ghdl/commit/1b23af06d9806867887aebad36b71c26e10584c1.

doc/using/Synthesis.rst
eine: Thanks!

doc/using/Synthesis.rst
eine: Ups... https://github.com/ghdl/ghdl/commit/856c45e7f2374de4015c26cab5d02d092d53b267

PR #1489, opened 2020-10-07
vendor-libs: add machxo3d support

fayalalebrun:
I added a MachXO3D FPGA option to the precompilation script, since this is the FPGA I am currently working on and it was simple enough to add. I also updated the docs indicating that it is supported.


tgingold:
Thanks!

PR #1510, opened 2020-11-17
Skip some tests if gcc is not installed

LukasVik:
**Background**
Related to ghdl/docker#34.

We want to have a docker image with ghdl that is minimal in size. That implies not including a gcc installation in the image. We do however want to run the testsuite to make sure the docker image "works".

**Description**
This PR makes it possible to run the testsuite on a system that does not have gcc installed. Tests that require gcc will be skipped (reported as "ok").

### DIFF
--- a/testsuite/gna/bug097/testsuite.sh
+++ b/testsuite/gna/bug097/testsuite.sh
@@ -2,17 +2,19 @@
 
 . ../../testenv.sh
 
-if [ -z $CC ]; then
-  CC="gcc"
+if c_compiler_is_available; then
+  if [ -z $CC ]; then
+    CC="gcc"
+  fi
+
+  $CC -c -fPIC getrand.c
+  $CC -o getrand.so --shared getrand.o
+
+  analyze tb.vhdl
+  elab_simulate tb
+
+  rm -f getrand.o getrand.so
 fi
-
-$CC -c -fPIC getrand.c
-$CC -o getrand.so --shared getrand.o
-
-analyze tb.vhdl
-elab_simulate tb
-
 clean
-rm -f getrand.o getrand.so
 
 echo "Test successful"
--- a/testsuite/gna/issue1326/testsuite.sh
+++ b/testsuite/gna/issue1326/testsuite.sh
@@ -7,10 +7,13 @@ elab mytestbench
 
 simulate mytestbench --wave=dump.ghw | tee mytestbench.out
 
-gcc ../../../src/grt/ghwdump.c ../../../src/grt/ghwlib.c -I../../../src/grt/ -o ghwdump
+if c_compiler_is_available; then
 
-# We're just checking that ghwdump doesn't crash on a zero length signal.
-./ghwdump -ths dump.ghw > dump.txt
+  gcc ../../../src/grt/ghwdump.c ../../../src/grt/ghwlib.c -I../../../src/grt/ -o ghwdump
+
+  # We're just checking that ghwdump doesn't crash on a zero length signal.
+  ./ghwdump -ths dump.ghw > dump.txt
+fi
 
 rm -f mytestbench.out ghwdump dump.txt dump.ghw
 clean
--- a/testsuite/testenv.sh
+++ b/testsuite/testenv.sh
@@ -148,6 +148,15 @@ synth_tb()
   clean
 }
 
+# Check if a C compiler is installed on this system
+c_compiler_is_available ()
+{
+  if [ -z $CC ]; then
+    CC="gcc"
+  fi
+  which $CC
+}
+
 # Check if a feature is present
 ghdl_has_feature ()
 {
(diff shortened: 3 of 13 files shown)


testsuite/vpi/vpi001/testsuite.sh
```diff
@@ -5,7 +5,7 @@
 analyze mydesign.vhdl
 elab myentity
 
-if ghdl_has_feature myentity vpi; then
+if c_compiler_is_available && ghdl_has_feature myentity vpi; then
```
eine, line 8: Instead of skipping each vpi test, I would skip the whole (sub)testsuite at the top of https://github.com/LukasVik/ghdl/blob/testsuite_gcc/testsuite/vpi/testsuite.sh

testsuite/gna/bug097/testsuite.sh
```diff
@@ -2,17 +2,19 @@
 
 . ../../testenv.sh
 
-if [ -z $CC ]; then
-  CC="gcc"
-fi
+if c_compiler_is_available; then
```
eine, line 5: Please, echo a message and `exit 0`, instead of wrapping all the commands. See https://github.com/ghdl/ghdl/blob/master/testsuite/gna/bug032/testsuite.sh#L3-L7

testsuite/gna/issue1326/testsuite.sh
```diff
@@ -7,10 +7,13 @@ elab mytestbench
 
 simulate mytestbench --wave=dump.ghw | tee mytestbench.out
 
-gcc ../../../src/grt/ghwdump.c ../../../src/grt/ghwlib.c -I../../../src/grt/ -o ghwdump
+if c_compiler_is_available; then
 
-# We're just checking that ghwdump doesn't crash on a zero length signal.
-./ghwdump -ths dump.ghw > dump.txt
+  gcc ../../../src/grt/ghwdump.c ../../../src/grt/ghwlib.c -I../../../src/grt/ -o ghwdump
```
eine, line 12: BTW, maybe use `$CC` instead of `gcc`, in coherence with other tests.

testsuite/testenv.sh
```diff
@@ -148,6 +148,15 @@ synth_tb()
...
+c_compiler_is_available ()
+{
+  if [ -z $CC ]; then
+    CC="gcc"
+  fi
+  which $CC
+}
+
 # Check if a feature is present
 ghdl_has_feature ()
```
eine, line 161: Might be worth calling `c_compiler_is_available` from inside `ghdl_has_feature`. That will avoid most of the modifications to individual tests (except bug097 and issue1326).

tgingold:
I am not against that PR, but I'd like to suggest something simpler.

If you aim at reducing the size of the docker image, just remove the testsuite (or maybe just keep the sanity/ part).
You can run the testsuite with the image if you want to check the build.

eine:
> If you aim at reducing the size of the docker image, just remove the testsuite (or maybe just keep the sanity/ part).
> You can run the testsuite with the image if you want to check the build.

That is already the case. The testsuite is not part of the image. The procedure is as follows:

- [pre-built] Create ghdl/build image with all build dependencies (zlib-dev, gnat, gcc, llvm...).
- [pre-built] Create a ghdl/run image with runtime dependencies (zlib, libgnat, libllvm...).
- Build GHDL in a ghdl/build container, run `make DESTDIR=<TMP_DIR>` and generate a tarball from `<TMP_DIR>`.
- Install the tarball in a ghdl/run container and generate a ghdl/ghdl image.
- Execute a ghdl/ghdl image by binding the testsuite from the host.

So, the 'production' image (ghdl/ghdl) currently contains the following only:

- libgnat
- make
- zlib1g-dev
- gcc
- libc6-dev
- [GCC backend only] lcov
- [LLVM backend only] libllvm
- GHDL tarball extracted to `/`

Lukas' proposal is the following:

- libgnat
- make
- zlib1g-dev
- **[GCC backend only]** gcc
- **[GCC backend only]** libc6-dev
- [GCC backend only] lcov
- [LLVM backend only] libllvm
- GHDL tarball extracted to `/`

I am not sure about `gcc` and `libc6-dev` being necessary for LLVM backend. The point is that those are not required in the mcode images.

tgingold:
But when is the testsuite executed ?

You can run the whole testsuite after the build, and simply the sanity check later.

eine:
> But when is the testsuite executed ?
> 
> You can run the whole testsuite after the build, and simply the sanity check later.

The point is to run the testsuite in a "clean" environment, not the one with all the build dependencies. So, the testsuite is executed in a container with `libgnat` and `zlib1g-dev` only.

We do it like that because we want to replicate the experience of a user which installs a pre-built package and attempts to run the full testsuite (or any specific test).

Precisely, if we run the tests in the build environment and only the sanity check "outside", then we would not detect the difference of GCC being available on the run environment.

tgingold:
Thanks!

PR #1497, opened 2020-10-23
Support building GHDL with LLVM backend using clang on MSYS2/mingw64

umarcor:
Currently, building GHDL with LLVM backend using clang on MSYS2/mingw64 fails because the output of `clang -dumpmachine` is not supported.

On GNU/Linux:

```
# gcc -dumpmachine
x86_64-linux-gnu

# clang -dumpmachine
x86_64-pc-linux-gnu
```

On MINGW64:

```
# gcc -dumpmachine
x86_64-w64-mingw32

# clang -dumpmachine
x86_64-w64-windows-gnu
```

This PR fixes `configure` and `src/grt/Makefile.inc` for detecting `x86_64-w64-windows-gnu` properly. At the same time, the PKGBUILD recipe is modified to use `CC=clang` and `CXX=clang++`.

The build is successful and several tests do pass. However, one of the tests in gna fails:

```
  gna issue635: failed
  analyze tb.vhdl ram.vhdl fsm.vhdl
  ram.vhdl:29:15:warning: declaration of "ram" hides entity "ram" [-Whide]
  elaborate and simulate (failure expected) testbench
  ../../src/ieee2008/numeric_std-body.vhdl:3036:7:@100ns:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
  ../../src/ieee2008/numeric_std-body.vhdl:3036:7:@100ns:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
  ../../src/ieee2008/numeric_std-body.vhdl:3036:7:@100ns:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
  ../../src/ieee2008/numeric_std-body.vhdl:3036:7:@100ns:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
  ../../src/ieee2008/numeric_std-body.vhdl:3036:7:@200ns:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
  ../../src/ieee2008/numeric_std-body.vhdl:3036:7:@200ns:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
  ../../src/ieee2008/numeric_std-body.vhdl:3036:7:@200ns:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
[...]

PR #1516, opened 2020-11-19
win32: add fprintf to grt_save_backtrace

eine:
Honestly, I don't know how to explain this. Since MSYS2 recipes were modifidied for using clang when building with LLVM backend (#1497), one test has been failing in CI: issue635. It is difficult to debug, because I cannot reproduce locally.

The issue is that the exit code is `3`, when it should be `0`. I run the test on GDB in CI, and I could confirm that: https://github.com/eine/ghdl/runs/1422980823?check_suite_focus=true#step:8:64. Then, Tristan told me that the issue might be in `src/grt/config/win32.c`. Hence, I added some `printf` statements, to see which functions are used. Surprisingly, that fixed the issue. I removed the print statements one by one, until I came to this minimal "fix".

So, adding an empty printf statement to `grt_save_backtrace` fixes the issue, but I cannot explain why.

### DIFF
--- a/src/grt/config/win32.c
+++ b/src/grt/config/win32.c
@@ -198,6 +198,14 @@ __ghdl_run_through_longjump (int (*func)(void))
 void
 grt_save_backtrace (struct backtrace_addrs *bt, int skip)
 {
+  /* FIXME
+  testsuite/gna/issue635 fails on GitHub Actions when executed with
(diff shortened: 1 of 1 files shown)


tgingold:
You should really add a comment to explain why there is a dummy printf.

PR #1525, opened 2020-11-26
Fix 'ghdl build produces an executable with RWX sections'

### LINKED ISSUE #1524: ghdl-0.37 build produces an executable with RWX sections
**Description**

While trying to create a custom ebuild (package build specification for the portage package manager), I ran across the following portage error:

> QA Notice: The following files contain writable and executable sections
> Files with such sections will not work properly (or at all!) on some
> architectures/operating systems.  A bug should be filed at
> https://bugs.gentoo.org/ to make sure the issue is fixed.
> For more information, see:
>
>   https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart
>
> Please include the following list of files in your report:
> Note: Bugs should be filed for the respective maintainers
> of the package in question and not hardened@gentoo.org.
>  `RWX --- --- usr/bin/ghdl`

<details>

**Expected behaviour**
I didn't expect this QA error to be shown. It indicates that the executable contains sections (I suppose, often: an WX stack) which are marked both executable and writable, which is discouraged.

**How to reproduce?**

```sh
# enable my gentoo portage overlay
layman -a zscheile

# install ghdl
emerge -v1 --autounmask=y =sci-electronics/ghdl-0.37
# (it may be needed to apply package.mask changes using
#  'dispatch-conf' and rerun the above 'emerge' command)
```
The 'emerge' command currently prints an QA message.
Alternatively, after an successful emerge, `scanelf -lpqe` should print at least the following:

`RWX --- ---  /usr/bin/ghdl`

**Context**

- OS: Gentoo Linux with `sys-libs/glibc-2.32-r2, dev-lang/gnat-gpl-2019-r2 (based upon GCC 8.3.1)`
- Origin:
  - [X] Package manager: portage, [custom ebuild](https://github.com/zserik/portage-zscheile/blob/c41ef0f353bfd8dfe7ac88fa465354c33f131aa3/sci-electronics/ghdl/ghdl-0.37.ebuild)
  - [X] Built from sources: `27093a5b476335b1e6b98cc7d20e34516afde196` (tag `v0.37`)

</details>

**Additional context**
I scanned the work tree (e.g. source tree + compilation output):
```
[...] /tmp/portage/sci-electronics/ghdl-0.37/work/ghdl-0.37 # scanelf -qeR .
  RWX --- ---  ./ghdl_mcode
  !WX --- ---  ./chkstk.o
```
 and found that `chkstk.o` seems to be the culprit, which is the compilate of **`src/grt/config/chkstk.S`**.

This is probably fixable by following the tips in https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart#How_to_fix_the_stack_.28in_practice.29 (section "How to fix the stack (in practice)")

zserik:
This PR adds an `.note.GNU-stack` section to `chkstk.o`, which fixes #1524.
Additionally, it updates `.gitignore` to ignore a few additional generated files on linux. I sorted the "generated files list" to be easier comparable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,26 +16,31 @@ testsuite/get_entities
 /dist/msys2-mingw/**/logpipe.*
 /dist/msys2-mingw/**/*.log
 
-# Generated directories on Linux
-lib/
-
 # Generated files on Linux
-/Makefile
-libghdlvpi.so
-version.ads
-version.tmp
 config.status
 default_paths.ads
+doc/_build/
 elf_arch.ads
-ghdl.gpr
 ghdl1-*
+ghdl.gpr
 ghdl_llvm
+ghdl_mcode
+ghdlsynth_maybe.ads
 grt-backtraces-impl.ads
 grt-files
 grt-files.in
 grt.lst
+lib/
+libghdl-*.so
+libghdl.a
+libghdl.bind
+libghdl.link
+libghdlvpi.so
 libgrt.a
+/Makefile
+ortho_code-x86-flags.ads
 run-bind.adb
 run-bind.ads
 src/version.ads
-doc/_build/
+version.ads
+version.tmp
--- a/src/grt/config/chkstk.S
+++ b/src/grt/config/chkstk.S
@@ -50,3 +50,7 @@ __chkstk:
 #endif
 
 	.ident	"Written by T.Gingold"
+
+#if defined(__linux__) && defined(__ELF__)
+	.section .note.GNU-stack,"",%progbits
+#endif


zserik:
About some of the CI failures: The code seems to fail in `src/ghdldrv/ghdlvpi.adb:151` with `access check failed`, which, according to [SO](https://stackoverflow.com/questions/55987833/how-to-prevent-an-access-check-failed-when-using-all-on-records-pointers), might be caused by the fact that `Program_Name` might be `null`, in which case the `.all` method (?) call fails with an access check error, because it would try to dereference a null ptr, I suppose.

tgingold:
I fear that the CI failure is a consequence of your patch.

While your patch is correct, I fail to see how it would make the whole executable !X for the stack.  As far as I know, part of the runtime still uses trampolines.

Does the test pass on your system before and after your change ?

zserik:
As far as I'm concerned, `scanelf -qeR .`, when run on an build tree of this patched source doesn't return any hits, so that would mean that no other trampolines using WX are used in the code afaik. I ran the testsuite on my system, once before the change, and once after the change, and both times they were successful.

zserik:
I'm unsure how to reproduce the failing tests, as I'm not sure what the `mcode` backend from `ci-run.sh` exactly is.

eine:
I fear that CI issues might have been my fault, as some WIP images were unwantedly upstreamed. It should be fixed now. I restarted CI workflows and all looks good.

tgingold:
Great!

tgingold:
All good.  Sorry for the incorrect diagnostic and thank you for the patch.

PR #1556, opened 2020-12-28
Cleanup and Restructuring of pyGHDL

Paebbels:
# Description
In cooperation with @umarcor, we did some brainstorming and cleanup to `/python` so the *thin* API as well as other higher layer APIs can grow on top of GHDLs `libghdl`. We bundled a first big set of refactoring in this PR and would like to discuss our goals and ideas for GHDL's Python interface(s).

**Changes in summary:**
1. rename `/python` to `/pyGHDL`
2. move `setup.py` from `/python` to `/`
3. reorganize Python packages:
   * `pyGHDL`
     * `cli` - command line interfaces (e.g. language server entrypoint)
     * `dom` - Document Object Model based on [pyVHDLModel](https://github.com/VHDL/pyVHDLModel)
     * `libghdl` - low-level API for GHDL (general); mainly auto generated from Ada sources
       * `vhdl` -  low-level API for VHDL; mainly auto generated from Ada sources
     * `lsp` - Language Server Protocol implementation (libghdl adapter, protocol, server)
     * `xtools` - ???
4. This also merges hierarchy level `thin` with `libghdl`.
5. Testcase/example `python/units01` is translated to a Python unittest in `pyunit/libghdl/Initialize.py`
6. Enhancements to `.gitignore`

## Naming clarifications
Currently, there is a mix of: `libghdl`, `pyghdl`, `ghdl-ls` and `python`. I like to unify it all under `pyGHDL` and use subpackages to distinguish between low-level (partially autogenerated) APIs (`libghdl` - refering to the matching *.so/*.dll), the language server implementation `lsp` and other higher-level APIs like `dom`. As libghdl and dom are mainly libraries itself for use by other applications or users, the command line interfaces and entrypoints become secondary. They are gathers in `cli` either as a real entry point like for `ghdl-ls` or as an example / test.

## Renaming the `python` directory
Python package projects need to have (everything else gets ugly and complicated) the package name as a directory in the root of a VCS repository. That why `/python` gets renamed to `pyGHDL`. In addition, `setup.py` is moved to the root and the collection of packages is automated with `packages=find_packages()`, so no hand curated list is needed.

## `pyGHDL.cli`
This will contain command line interfaces for various APIs or services. Currently we have two:
* `ghdl-ls` - GHDLs language server protocol implementation for VHDL
* Another CLI is planned for testing the DOM. Maybe as it is for testing, it should be part of `testsuite` or get another package namespace like `pyGHDL.test` or `pyGHDL.example`.

## `pyGHDL.dom`
With this PR comes a proof of concept implementation of a document object model (DOM) based on [pyVHDLModel](https://github.com/VHDL/pyVHDLModel).

A simple prettyPrint output looks like this:
image

## `pyGHDL.libghdl` / `pyGHDL.libghdl.vhdl`
In here, I would like to see all parts of the low-level interface to `libghdl....so` including some iterators and generators for easier handling of linked lists.

## `pyGHDL.lsp`
Here we can put all code for the language server protocol implementation using `pyGHDL.libghdl`.

## `pyGHDL.xtools`
These files are currently unclear.

## Python `unittest`
The testcase in `python/units01` has been transformed to `pyunit/libghdl/Initialize.py` and can be called like this:
```bash
GHDL_PREFIX=/opt/ghdl/0.37-dev/lib/ghdl python -m unittest testsuite.pyunit.libghdl.Initialize
```

In a later step, the whole testsuite can be called like this:
```bash
GHDL_PREFIX=/opt/ghdl/0.37-dev/lib/ghdl python -m unittest testsuite.pyunit
```

## Changes to `.gitignore`
Added Python specific settings. Added some comments and structure.

# What has been tested so far after restructuring?
* [x] DOM example is OK
* [x] `pyunit.libghdl.Initialize` testcase is OK
* [x] LSP &rArr; I have no environment to test the language server

# NOT covered by this PR
* Final or stable implementation of a DOM
* Documentation updates especially for `pyGHDL`

-------------------
/cc @eine


tgingold:
Ok but for xtools.  These are scripts used to generate some sources files and they are referenced by src/*/Makefile
I don't think they are part of pyGHDL.  Maybe they could be moved to a different directory.  But the Makefile must be adjusted.

umarcor:
A probably minor detail, but which must be explained: `pyutils.py` was moved from being an script in `libghdl.thin.vhdl` to a submodule: `pyGHDL.libghdl.utils` (so, to `pyGHDL/libghdl/utils/__init__.py`).

The motivation is that all `pyGHDL/libghdl/*.py` and `pyGHDL/libghdl/vhdl/*.py` sources are auto-generated, but `pyGHDL/libghdl/utils/*.py` contains manually written iterators and helpers on top of those. @tgingold, can you please confirm this? 

> Ok but for xtools.  These are scripts used to generate some sources files and they are referenced by src/*/Makefile
I don't think they are part of pyGHDL.  Maybe they could be moved to a different directory.  But the Makefile must be adjusted.

I believe we do also need to modifiy the target directories where python scripts are autogenerated (because we moved `libghdl/thin` to `libghdl/` and `libghdl/thin/vhdl` to `libghdl/vhdl`. Is that done in xtools only or also passed as args from the Makefiles?

Paebbels:
**Update:**  
I have now an environment to test the `ghdl-ls` (which is broken) and I'm fixing it now. No changes to the logic, just to import statements and related effects of restructuring.

@umarcor takes care of xtools...

tgingold:
Why not.

Not all py files were automatically generated.  Some of them were manually written.  See src/vhdl/Makefile.
But pyutils.py is not a binding but an additional layer.  So it might be put in a different place.

> I believe we do also need to modifiy the target directories where python scripts are autogenerated (because we moved libghdl/thin to libghdl/ and libghdl/thin/vhdl to libghdl/vhdl. Is that done in xtools only or also passed as args from the Makefiles?

The xtools write on standard output and the Makefile redirect the output to the right place.

Paebbels:
**Update:**
* `ghdl-ls` starts on my side without errors. I can't test the client side, because I have no environment for it.
* Many updates where made to CI and testsuite scripting.
  * `testsuite/pyunit` is now included in regression testing
  * It offers 1 initialization test (formerly `units01`) and 4 tests for a simple DOM implementation
* `pyGHDL.libghdl` needs an update: tests in pyunit refer to non-existing functions
  * [ ] Activate `analyze_init_status` check again.

Paebbels:
@umarcor can you please confirm if LSP client was now satisfied with our rework?

Unfortunately a PR message got lost... (I must have closed the browser tab without saving it.)
@tgingold please run xtools to sync Ada code and auto generated parts of libghdl. Some new functions you call in Python code are not yet part of libghdl bindings. This was unnoticed, because Python tests where not enabled in CI yet. This was down now and failed as expected.

When libghdl (Ada and Python) are in sync again, I'll activated 4 more Python lines and this PR should be complete :).

tgingold:
For xtools, you just run `make` in `src/vhdl`.

For LSP, you can checkout ghdl-language-server, there are tests inside (not very sure about the state of them, maybe they should be moved to the ghdl repo instead).

Paebbels:
@umarcor explain, LSP was merged (copied) into GHDL some time ago and the language server repo is "outdated".
So the matching tests where not copied over?

**Update for xtools:**
The `pnodes.py` was pure Python 2 code, I used `2to3.py` (which is part of every Python 3 installation) to upgrade that file to Python 3 syntax. I check the results and it looks good. I also called it and found one missing import in a nested error case (maybe never needed and thus not visible as an error before). The update was needed so `pnodespy.py` can be called in a pure Python 3 environment.

tgingold:
No, only the langserver was moved to GHDL.  But the repo contains integrations with IDE (like vs code).

Ok for xtools!

Paebbels:
@umarcor we need to fix xtools to generate the appropriate import statements. See https://github.com/ghdl/ghdl/pull/1556/commits/7e787e9ba5b4308ec1975760636c4f23adb1e762

image

Paebbels:
@umarcor Urgh. Our dear friend `raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : libraries.adb:639` is back...

@tgingold 1 out of 5 pyunit tests is passing. The 4 remaining tests are one group in `testsuite/pyunit/dom/SimpleEntity.py`.  
We get an error printout from libghdl, but it's hard to figure out what's the exact problem.

You can run it from project root directory:
```bash
cd <ghdl-root>
python -m unittest testsuite.pyunit                    # run all tests at once
python -m unittest testsuite.pyunit.libghdl.Initialize # run the low-level test (formerly 'units01')
python -m unittest testsuite.pyunit.dom.SimpleEntity   # run problematic test
```

tgingold:
> @umarcor Urgh. Our dear friend raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : libraries.adb:639 is back...

This is not the same issue.  You are just initializing libghdl twice.

Paebbels:
@tgingold How can you tell from `SYSTEM.ASSERTIONS.ASSERT_FAILURE` it's initialized twice?  
It looks like a very generic error message.

> You are just initializing libghdl twice.

Hmmm ok. Any idea how to solve this in regression testing?

Possibilities I see:
1. Make `libghdl` robust to double inits and/or allow to reset it (might be very complicated.)  
   I assume we would need to unload and reinit the *.dll/*.so file for this &rArr; not good...
2. Change the test, to only initialize once. Needs a DOM, that allows deletion of nodes.
3. Split the Testcase class into multiple classes.

Any other ideas?

Paebbels:
> 3. Split the Testcase class into multiple classes.

Hmmm this doesn't work, because it's still executed in the same Python process with a once initialized shared library.

tgingold:
> @tgingold How can you tell from SYSTEM.ASSERTIONS.ASSERT_FAILURE it's initialized twice?

I can guess it from 'libraries.adb:639' and look at the source.  There is a comment on the assertion.

It is possible to de-initialize but the API is not fully written.  Maybe for the next step.
The simplest is to change the test.

Paebbels:
I would say it's ready for a final review. CI is running and expected to pass.

tgingold:
Squash or no squash ?

Is it possible to build pyGHDL without the dom part (example: if you just need to build ghdl-ls) ?

Paebbels:
No squash, just merge :).

For partial Python packages. It's possible to split pyGHDL into multiple packages, so every directory in pyGHDL becomes one package:
* dom
* lsp
* libghdl

But this is some research and investigation work, because my last 2 tries where not successful, but I know projects that use it and it's also officially documented by Python.org.

umarcor:
> @umarcor can you please confirm if LSP client was now satisfied with our rework?

I could not test it yet... I could successfully start ghdl-ls on MSYS2. However:

- VSCode does not find resources in MSYS2, by default.
- The VSCode client is broken at the moment: ghdl/ghdl-language-server#70

I downloaded the last successful build of the client from two months ago: https://github.com/ghdl/ghdl-language-server/actions/runs/314253457. Then, inside a MINGW64 shell, I set PYTHONPATH and I started VSCode from there (with the extension installed beforehand). That worked partially. The language server was started and it crashed because it could not find `vhdl_langserver`. So, I'm confused because `vhdl_langserver` does not exist neither in https://github.com/ghdl/ghdl-language-server/search?q=vhdl_langserver nor in https://github.com/ghdl/ghdl/search?q=vhdl_langserver. Therefore, I don't know where is the call/import coming from...

> When libghdl (Ada and Python) are in sync again, I'll activated 4 more Python lines and this PR should be complete :).

I seems that the PR was merged before enabling those 4 Python lines again?

> For xtools, you just run `make` in `src/vhdl`.

@tgingold, while fixing the calls to xtools, I found that it is used in other makefiles too (`src/edif/Makefile` and `src/psl/Makefile`). Should we run make in those subdirs too?

> @umarcor explained: LSP was merged (copied) into GHDL some time ago and the language server repo is "outdated".

Small correction: the language server repo is used for clients only, but it's not outdated. Well, the VSCode client is outdated at the moment, but that's not on purpose.

> we need to fix xtools to generate the appropriate import statements.

Ack.

> I assume we would need to unload and reinit the .dll/.so file for this ⇒ not good...

AFAIAA, this is the only possible procedure ATM. GHDL does not allow reset or restarting an "instance". This applies to either libghdl or loading simulation models as shared libraries. Hence, the DLL needs to be unloaded and loaded again, should you want to run different cases.

There is a tip about it at the bottom of https://ghdl.github.io/ghdl-cosim/vhpidirect/dynamic.html#loading-a-simulation. This is a C example: https://github.com/ghdl/ghdl-cosim/blob/master/vhpidirect/shared/shghdl/main.c#L37. And this is a Python example: https://github.com/ghdl/ghdl-cosim/blob/master/vhpidirect/shared/pycb/run.py#L35-L36.

I agree that it's not ideal, and there are a couple of issues about the topic in ghdl-cosim (ghdl/ghdl-cosim#14 and ghdl/ghdl-cosim#15). However, until we can fix that in GHDL, unloading/reloading is the way to go.

> It is possible to de-initialize but the API is not fully written.  Maybe for the next step.
The simplest is to change the test.

Hence, I agree with Tristan here. We should change the test, and focus on de-initialisation API as the next important requirement before releasing 1.0.

> Is it possible to build pyGHDL without the dom part (example: if you just need to build ghdl-ls) ?

> For partial Python packages. It's possible to split pyGHDL into multiple packages, so every directory in pyGHDL becomes one package

Currently, setting `PYTHONPATH=$(pwd)/pyGHDL` will make `import libghdl`, `import lsp` or `import dom` possible. However, there is no separated `pip install` for each of them.

I'm not sure we should support installing sub-packages through `pip`, because of a semantic issue. We reorganised/renamed the packages considering that `pyGHDL` is what gives a sense/meaning to the subpackages. That is `pyGHDL.lsp` or `pyGHDL.dom` do make sense. But users should not `import lsp`, because that's misleading. If we wanted to support that, we should add `*ghdl*` or `*vhdl*` to the names of all the subpackages.

Paebbels:
> It seems that the PR was merged before enabling those 4 Python lines again?

The current test code is as good as before. Meanwhile we can investigate how to load/unload the shared object or how to run `pyunit` with a parallel test runner (using multiple processes).

> I'm not sure we should support installing sub-packages through pip, because of a semantic issue. We reorganised/renamed the packages considering that pyGHDL is what gives a sense/meaning to the subpackages. That is pyGHDL.lsp or pyGHDL.dom do make sense. But users should not import lsp, because that's misleading. If we wanted to support that, we should add *ghdl* or *vhdl* to the names of all the subpackages.

When splitting a package into a multi-package setup construct, the packagename on PyPI and on disc stay as e.g. `pyGHDL.dom`. When you install one package or more packages, it only describes that another package can write into an existing package directory in you Python installation and not overwriting code.

umarcor:
Agree to the first point, and I was not aware of the second one. All good.

Checklist for further enhancements:

- [ ] imports generated by pnodes
- [ ] auto-generate from `src/edif/Makefile` and/or `src/psl/Makefile`?
- [ ] test ghdl-ls with a client
- [ ] update docs
- [ ] GHDL vs PREFIX vs GHDL_PREFIX
- [ ] run pyunit from *any* location, run individual tests, run multiple, list tests... i.e. get the previous functionality back (https://github.com/ghdl/ghdl/blob/340fc792bba2ffdb4f930bc427a39ea3a1b659b2/testsuite/python/testsuite.sh#L14-L28).

Issue #1561, opened 2020-12-30
Coverage Collection for Python APIs


Paebbels:
As we now have test cases in unit test format and pytest to run them, we can also collect line and branch coverage from Python APIs.

It requires the following Python packages (for testing/CI only):
* `coverage` and
* `pytest-cov`

Currently we achieve 50% coverage. This will increate when LSP tests are also executed.

![image](https://user-images.githubusercontent.com/956109/103344290-742f4b00-4a8e-11eb-81a4-8d33c9a00461.png)

The results can be exported as:
* command line output / ascii table
* HTML report
* XML report file ([cobertura format](https://cobertura.github.io/cobertura/))

I would like to visualize coverage results via [Codecov.io](https://codecov.io/).
We get these benefits compared to uploaded a raw HTML uploads to GH-Pages:
* Coverage badge [![codecov](https://codecov.io/gh/Paebbels/pyVHDLParser/branch/master/graph/badge.svg?token=zaxMBYSvEN)](https://codecov.io/gh/Paebbels/pyVHDLParser)
* Coverage history/graph over time  
  ![image](https://user-images.githubusercontent.com/956109/103344901-3f23f800-4a90-11eb-8ef9-d5cd93fe8f97.png)
* Coverage diffs (green = new code, but red parts, not yet covered by tests)  
 ![image](https://user-images.githubusercontent.com/956109/103345046-9de97180-4a90-11eb-9812-000f774cad37.png)
 

Here is an example file visualized from pyVHDLParser: [pyVHDLParser/Blocks/Generic.py:344](https://codecov.io/gh/Paebbels/pyVHDLParser/src/master/pyVHDLParser/Blocks/Generic.py#L244)  
![image](https://user-images.githubusercontent.com/956109/103344481-059ebd00-4a8f-11eb-8b93-fa3f77d6cd7d.png)

umarcor:
Let's go for it.

umarcor:
Ref #1567.

Paebbels:
@tgingold can you register GHDL at these services?
* [ ] https://www.codecov.io/
* [ ] https://www.codacy.com/

tgingold:
Done for codacy.

tgingold:
But no success for codecov.  I don't know how to add an organization.

Paebbels:
# For codacy:

Please create a secret for GHDL, so a GH Action can use the token to push coverage results.
* Token name: `CODACY_PROJECT_TOKEN`
[...]

PR #1562, opened 2020-12-30
Improvements to pyGHDL

Paebbels:
# Changes
* Added missing file headers incl. license text for Python files (non auto generated files)
* Improved documentation in `pyGHDL.libghdl` for "iterators" - actually its generators.
  * Added missing docstrings, added rst roles to reference parameters in text.
  * Added type annotations
* Removed `load_tests` protocol of Pythons unittests, because it's not supported by pytest and cause exceptions

# Ongoing Tasks
* [x] Merge with `master` and
  * [x] Check new init/finalze feature
  * [x] If OK, remove `pytest-forked`
* [x] Finalize code coverage checks (local in CI)
  * [x] Move coverage execution like test execution into directory `/testsuite`
* [ ] Wait on discussion #1561 
  * [ ] Activate sending coverage data to codecov.io via CI workflow.


Paebbels:
**Update:**
Python tests can now be executed from within PyCharm.

To get it working in general:
1. Change the used test framework / default test runner from Pythons `unittest` to `pytest`.  
   image
2. Change the pytest template to add a environment variable by default:  
   image

Paebbels:
**Update:**
Coverage collection is now working: https://github.com/Paebbels/ghdl/runs/1628245154?check_suite_focus=true#step:3:1776

@tgingold What about the service(s) mentioned in #1561 ?

tgingold:
I am OK with adding coverage.  I am not sure about how to use it but we will see.

Concerning this PR, I think it would be better if some trivial parts (like copyright headers) were in a separate PR as there is no real discussion about them.

Paebbels:
> I am OK with adding coverage. I am not sure about how to use it but we will see.

I use this as an indication how much I have tested. Do my tests cover most of the provided API. I enabled branch coverage, because you get it for free in Python. And services like CodeCov can display it. There is also Codacy (I recommend to enable it too). This con only display line coverage (from same data source), but it brings other checks like for Documentation, Config files, ...

Actually, in my public Python projects I enabled both services.

> Concerning this PR, I think it would be better if some trivial parts (like copyright headers) were in a separate PR as there is no real discussion about them.

I made semantically coherent commits. You could review this e.g. commit by commit.  
Shall I split this PR? What parts do you prefer?
* Code headers
* pyunit testcase update and coverage
* Changes to pyGHDL.libghdl code
  * Type annotations and docstring
  * Changed os.path to pathlib and exception handling

We can activate sending data to codecov/codacy later in another PR.

Paebbels:
Btw. splitting a branch (and thus the PR) on top of a master with squashed commits is not a simple task and can lead to data loss.

I'll try if I can cherry pick my changes to other branches...

tgingold:
> I use this as an indication how much I have tested.

Sure, but...

> Do my tests cover most of the provided API.

What is the reference ?  Sources or specifications ?  What is an acceptable coverage level ?  What do you do when you cannot achieve 100% coverage ? Do you write unit tests or integration tests ? How do you handle changes or refactoring ?

To be short, there are a lot of testing strategies.  For GHDL, I have always preferred integration tests (ie simple designs) because that's what the user does and it as independent as possible from the sources.  But the strategy for libghdl might be different.

> I made semantically coherent commits. You could review this e.g. commit by commit.

Ah OK.  But I generally look at the end result as it is easier to have a better view of the PR and to spot issues.

> Shall I split this PR? What parts do you prefer?

At least code headers.  That's a no brainer.  Just create a PR with that, I look quickly on it and merge it.  If code headers are with other changes, they will be delayed until the whole PR is accepted and in general it makes the PR much larger.

That's just the way I proceed!

tgingold:
Just to clarify: I am trying to find the best work flow for you and me.  We all have slightly different way to work so we need to find the optimal way.  Feel free to suggest variations!

(and yes, I am very happy with your contributions.  But I also want to understand it so that I could work with them or work on them)

Paebbels:
> What is the reference ?  Sources or specifications ?  What is an acceptable coverage level ?  What do you do when you cannot achieve 100% coverage ? Do you write unit tests or integration tests ? How do you handle changes or refactoring ?

For my public Python libraries I try to reach an 80% level. That's also what some tools have as a minimum coverage limit by default. This gives you room to not cover all special cases like "file not found exception".

I believe testing against specification is to much for this project or what I do in my public repos. My goal is to get a feeling where I miss testing for code I've written or to identify unneeded code. So its not a safety critical measure nor a scientific approach.

It also offers users of GHDL or pyGHDL an inside how well a project is tested. This can also be achieved with use cases or example based testing, but it shows user if they'll use feature XYZ, that it was probably tested before, but not every special cases.

----------
Writing tests for a parser is too complicated (as figured out my self with pyVHDLParser). So I also do here example based testing with golden references.

pyGHDL is already on a higher layer. So when we give a file, we can test an entity was found and has correct name in the matching objects, so we didn't mess up iterating nodes, right?

-----------
I'll split the PR. Pushes coming soon.

P.S. we can also have a teams or zoom call and I can explain quicker what I did or had in mind.

Paebbels:
OK, PR is split into 4 parts.
1. headers
2. pyGHDL.libghdl.utils
3. pyGHDL.libghdl
4. remaining stuff.

Will get 2 times a conflict for this section:  
image

First about the header, second about the import.

I virtually merged everything locally and merged the old branch, which show no modifications in the last merge. So I assume (besides possible side effects in Git for partial merges) that everything was split and no data should be lost.  
image

tgingold:
That's fine with me, I see your point.

Issue #1557, opened 2020-12-29
pyGHDL enhancements


umarcor:
Coming from #1556. 

Checklist for further enhancements:

- [ ] imports generated by pnodes
- [ ] auto-generate from `src/edif/Makefile` and/or `src/psl/Makefile`?
- [ ] test ghdl-ls with a client
- [ ] update docs
- [ ] GHDL vs PREFIX vs GHDL_PREFIX
- [ ] run pyunit from *any* location, run individual tests, run multiple, list tests... i.e. get the previous functionality back: https://github.com/ghdl/ghdl/blob/340fc792bba2ffdb4f930bc427a39ea3a1b659b2/testsuite/python/testsuite.sh#L14-L28

Paebbels:
**Update for `pyunit`:**
* Parallel execution with `pytest` and its plugin `pytest-forked` is now possible.  
  This works around the problem for independent unit testing that `libghdl` can only be initialized once.
* `pytest`s strange/strict naming conventions were disabled via `pytest.ini` in `testsuite/pyunit` to comply with normal unittest naming rules.
* `testsuite.sh` was adjusted to run from within `testsuite` directory.
* All 5 testcases are now found and executed.

See PR #1558

umarcor:
After the latest changes, running the test suite in-place fails on MSYS2 because it cannot find the shared library:

```
  make install.vpi.local
  make CC=gcc test
```

Paebbels:
I added another task to our list: integrate language server tests into pyunit. This is handled in PR #1585.

Paebbels:
outdated.

PR #1564, opened 2020-12-31
Added file headers and license texts.

Paebbels:
Splitting #1562:
* File headers and license texts.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,7 +1,8 @@
 -r ../pyGHDL/requirements.txt
-sphinx>=3.0.0
-recommonmark
-python-dateutil
+
+sphinx>=3.4.1
+recommonmark>=0.7.1
+python-dateutil>=2.8.1
 # sphinxcontrib-textstyle>=0.2.1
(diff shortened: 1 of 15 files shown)


Paebbels:
OK, this is complete for review. First CI jobs are good.

tgingold:
Thanks!

Paebbels:
P.S. should the auto generated files write a tiny header like:
```
# This file was auto generated by `pnodepy.py` on DD.MM.YYYY.
# Run 'make' in '/src/vhdl' to regenerate.
```
I assume a GPL header is not required for it.

tgingold:
Yes, we can add such an header but without the date so that if the output is the same, there is no need to commit.

Paebbels:
Oh, you're right, because it's pushed to Git :).

umarcor:
Is adding that header to auto-generated files done or tracked (#1557)?

Paebbels:
This PR is about fixed files.

The auto generated headers are here: #1568.

It's not tracked in #1557.

PR #1565, opened 2020-12-31
Modifications to pyGHDL.libghdl.utils

Paebbels:
This is a split from #1562 and fixes or adds:
* Type annotations
* Fixed imports (unused imports)
* Object references
* Changes assert /which raises an AssertionException) to a regular Exceptions.
* Modified docstrings to use restructured text roles for processing by Sphinx.


Paebbels:
This split from #1562 should be ready for review.

tgingold:
Thanks!

PR #1566, opened 2020-12-31
Modifications to pyGHDL.libghdl

Paebbels:
This is split from #1562 and brings:
* Updated docstrings
* Using [pathlib](https://docs.python.org/3/library/pathlib.html) instead of os.path.  
  e.g. `absolute(join(dir, "..", "..", "lib"))` becomes `(dir / "../../lib").resolve()`
* Add exception handling instead of None checks.


Paebbels:
This is also ready for review.

tgingold:
Thanks!

PR #1567, opened 2020-12-31
pyGHDL: Remaining changes

Paebbels:
This is split from #1562:
* activate coverage collection
* new SimplePackage test case

### DIFF
--- a/testsuite/testsuite.sh
+++ b/testsuite/testsuite.sh
@@ -141,7 +141,7 @@ do_gna () {
 # The Python Unit testsuite: regression testsuite for Python bindings to libghdl
 do_pyunit () {
   gstart "[GHDL - test] pyunit"
-  PYTHONPATH=$(pwd)/.. python3 -m pytest -rA $PYTEST_ARGS pyunit
+  PYTHONPATH=$(pwd)/.. python3 -m pytest -rA pyunit
   gend
 }
 
--- a/testsuite/testsuite.sh
+++ b/testsuite/testsuite.sh
@@ -141,7 +141,7 @@ do_gna () {
 # The Python Unit testsuite: regression testsuite for Python bindings to libghdl
 do_pyunit () {
   gstart "[GHDL - test] pyunit"
-  PYTHONPATH=$(pwd)/.. python3 -m pytest pyunit
+  PYTHONPATH=$(pwd)/.. python3 -m pytest -rA --cov=.. --cov-config=.coveragerc pyunit
   gend
 }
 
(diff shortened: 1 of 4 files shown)


Paebbels:
So this should be now the last part of #1562.

testsuite/testsuite.sh
```diff
@@ -141,7 +141,7 @@ do_gna () {
 # The Python Unit testsuite: regression testsuite for Python bindings to libghdl
 do_pyunit () {
   gstart "[GHDL - test] pyunit"
-  PYTHONPATH=$(pwd)/.. python3 -m pytest pyunit
+  PYTHONPATH=$(pwd)/.. python3 -m pytest -rA --cov=.. --cov-config=.coveragerc pyunit
```
umarcor, line 144: Do we want to enable coverage collection always? We run this test suite in 8-10 different jobs at the same time. We need to either merge all of them at the end (if possible), or discard all but one.

If we enable some external service, can it handle several submits for the same CI run?

Paebbels: I think one of them can handle a "label" like nightly, regression, etc. so we could distinguish results.

In my Python projects I run only on Python 3.8 (will change to 3.9 soon), but don't run on 3.6, 3.7, ...
Uploading is currently not enabled and we have just 7 tests, so this is harmless, but yes, we should think about load reduction when the upload to the server works.

Paebbels: `-rA` report all results
`--cov=.. --cov-confi=.coveragerc` is for coverage

I move this from `pytest.ini`, because paths are relative to working dir not the `pytest.ini`. I wanted to avoid confusion if parameters are in the ini, but need a call from a specific directory.

umarcor: I propose the following:

```
PYTHONPATH=$(pwd)/.. python3 -m pytest -rA $PYTEST_ARGS pyunit
```

Then, in just one of the jobs, set:

```yml
env:
  PYTEST_ARGS: '--cov=.. --cov-config=.coveragerc'
```

umarcor: See branch [ci/coverage](https://github.com/umarcor/ghdl/commits/ci/coverage) in my fork. Feel free to pick the first commit only, or both of them. See CI run: https://github.com/umarcor/ghdl/actions/runs/455551490

tgingold: Yes, I agree with @umarcor : just enable coverage on one test.

Paebbels: @umarcor I integrated your changes via FF-merge.

@tgingold this PR is then ready for merge. CI results are running for this branch, but it was already proven on Unai's fork.

tgingold:
Thanks!

PR #1569, opened 2020-12-31
scripts: add 'update_py_bindings.sh'

umarcor:
This PR adds an script for regenerating all python bindings in pyGHDL. At the same time, a test is aded to CI. It will produce a failure if a commit is pushed without updating pyGHDL sources (if required).


tgingold:
Thanks!

PR #1571, opened 2020-12-31
synth: Add option to treat PSL asserts as PSL assumes and vice-versa

tmeissner:
This adds the ability to GHDL synthesis to treat PSL assertions as PSL assumes and vice-versa.

It adds two new options to the `--synth` command:

##### `--assert-assumes`
Treat all PSL assertions as PSL assumes

##### `--assume-asserts`
Treat all PSL assumes as PSL assertions

The options names mimic their counter parts of the Yosys [`read_verilog`](http://www.clifford.at/yosys/cmd_read_verilog.html) command. If we want other names (`assert` instead of `asserts`  for example), I am happy to discuss this.

Furthermore the PSL implementation documentation is enhanced with these two options.

This is a work-in-process. Test cases will be added before the PR is finished.

### DIFF
--- a/doc/using/Synthesis.rst
+++ b/doc/using/Synthesis.rst
@@ -92,11 +92,38 @@ simulation back-ends. Hence, available options for synthesis are the same as for
 .. option:: --no-assert-cover
 
   Disable automatic cover PSL assertion activation. If this option isn't used, GHDL generates
-  `cover` directives for each `assert` directive automatically during synthesis.
+  `cover` directives for each `assert` directive (with an implication operator) automatically during synthesis.
 
   Example::
 
     $ ghdl --synth --std=08 --no-assert-cover my_unit
+
+.. option:: --assert-assumes
+
+  Treat all PSL asserts like PSL assumes. If this option is used, GHDL generates an `assume` directive
+  for each `assert` directive during synthesis. This is similar to the `-assert-assumes`
+  option of Yosys' `read_verilog <http://www.clifford.at/yosys/cmd_read_verilog.html>`_ command.
+
+  Example::
+
+    $ ghdl --synth --std=08 --assert-assumes my_unit
+
+  As all PSL asserts are treated like PSL assumes, no `cover` directives are automatically generated for them,
+  regardless of using the :option:`--no-assert-cover` or not.
+
+
+.. option:: --assume-asserts
+
+  Treat all PSL assumes like PSL asserts. If this option is used, GHDL generates an `assert` directive
+  for each `assume` directive during synthesis. This is similar to the `-assume-asserts`
+  option of Yosys' `read_verilog <http://www.clifford.at/yosys/cmd_read_verilog.html>`_ command.
+
+  Example::
+
+    $ ghdl --synth --std=08 --assume-asserts my_unit
+
+  `cover` directives are automatically generated for the resulting asserts (with an implication operator)
+  if :option:`--no-assert-cover` isn't used.
 
 .. TIP::
   Furthermore there are lot of debug options available. Beware: these debug options should only used
--- a/src/ghdldrv/ghdlsynth.adb
+++ b/src/ghdldrv/ghdlsynth.adb
@@ -128,6 +128,10 @@ package body Ghdlsynth is
       P ("    Neither synthesize assert nor PSL");
       P ("  --no-assert-cover");
       P ("    Cover PSL assertion activation");
+      P ("  --assert-assumes");
+      P ("    Treat all PSL asserts like PSL assumes");
+      P ("  --assume-asserts");
+      P ("    Treat all PSL assumes like PSL asserts");
    end Disp_Long_Help;
 
    procedure Decode_Option (Cmd : in out Command_Synth;
@@ -152,6 +156,10 @@ package body Ghdlsynth is
          Synth.Flags.Flag_Assert_Cover := False;
       elsif Option = "--assert-cover" then
          Synth.Flags.Flag_Assert_Cover := True;
+      elsif Option = "--assert-assumes" then
+         Synth.Flags.Flag_Assert_As_Assume := True;
+      elsif Option = "--assume-asserts" then
+         Synth.Flags.Flag_Assume_As_Assert := True;
       elsif Option = "--top-name=hash" then
          Cmd.Top_Encoding := Name_Hash;
       elsif Option = "--top-name=asis" then
--- a/src/synth/netlists-disp_vhdl.adb
+++ b/src/synth/netlists-disp_vhdl.adb
@@ -1343,7 +1343,8 @@ package body Netlists.Disp_Vhdl is
             Put_Line (";");
          when Id_Assert =>
             Disp_Template
-              ("  \l0: postponed assert \i0 = '1' severity error;" & NL, Inst);
+              ("  \l0: postponed assert \i0 = '1' severity error; --  assert"
+               & NL, Inst);
          when Id_Assume =>
             Disp_Template
               ("  \l0: assert \i0 = '1' severity warning; --  assume" & NL,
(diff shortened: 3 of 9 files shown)


tmeissner:
@eine I added hints in the docs explaining that the feature matches the one in Yosys. And fixed the options name to be similar to the `read_verilog` ones :wink:

tmeissner:
I'm not sure about useful test cases.

In the GHDL repo I could synthesize the design and check the resulting VHDL netlist for `assume` or `assert`. They look the same in netlist, but with an additional comment showing an `assume`. For example:

##### assert
```vhdl
-- psl_always.vhd:41:3
with_always_a: postponed assert n14_o = '1' severity error;
```

##### assume (with `--assert-assumes`, same PSL code)
```vhdl
-- psl_always.vhd:41:3
with_always_a: assert n14_o = '1' severity warning; --  assume
```

Other possibility would be to add an test case to the ghdl-yosys-plugin project. There I could check the correct behaviour. But actually the feature has nothing to do with the ghdl-yosys-plugin.

tgingold:
I would just generate the netlist and use grep to check the result (presence of the comment).

doc/using/Synthesis.rst
```diff
@@ -98,6 +98,26 @@ simulation back-ends. Hence, available options for synthesis are the same as for
 
     $ ghdl --synth --std=08 --no-assert-cover my_unit
 
+.. option:: --assert-assumes
+
+  Treat all PSL assertions like PSL assumes. If this option is used, GHDL generates `assume` directives
+  instead each `assert` directive during synthesis. This is similar to the `-assert-assumes`
+  option of Yosys `read_verilog`<http://www.clifford.at/yosys/cmd_read_verilog.html>`_ command.
```
eine, line 105: I think it should be:

```
`read_verilog <http://www.clifford.at/yosys/cmd_read_verilog.html>`_ 
```

Note the space after `read_verilog`. Same below.

tmeissner: Strange syntax. I will fix it.

tgingold:
Note about Yosys: you can also use `chformal -assert2assume` and `chformal -assume2assert`

eine:
@tgingold, does that mean that this feature is mostly useful for `ghdl --synth` because it could be done as a command in yosys (between `ghdl` and `write_*`)?

tmeissner:
Honestly, the yosys documentation is huge and I try to avoid to do more than simple things. Most of the time you have to do every single step explicitly when you want to change some command. Also I don't know how I can use such commands when I simply want use SymbiYosys.

Xiretza:
> Also I don't know how I can use such commands when I simply want use SymbiYosys.

You simply put it in the `[script]` section of your sby file.

eine:
@tmeissner, I think it is good to have three different places: GHDL frontend, Verilog frontend or pipeline command. Moreover, the options names for the frontends match, which is nice. I was asking Tristan just for better understading.

tmeissner:
> > Also I don't know how I can use such commands when I simply want use SymbiYosys.
> 
> You simply put it in the `[script]` section of your sby file.

Thanks. I added it in a test design and it seems to work. I also can select the unit on which `chformal` should work.

tmeissner:
I agree to @eine to have these option also in the VHDL frontend (GHDL synthesis) would be good for consistency. So I will add the missing test cases and finish the PR.

tgingold:
Yes, seconded.

tmeissner:
@tgingold Question: I have two test files which contain exactly one PSL `assume` directive .

If I generate the VHDL netlist with the `--synth` command and without the new option, the resulting file contains two `assume` directives.

If I use the new options, only one of the two `assume` is converted in an assert.

For example:

Input file:
```vhdl
library ieee;
use ieee.std_logic_1164.all;

entity assume0 is
  port (
    clk : std_logic;
    i   : integer
  );
end assume0;

architecture behav of assume0 is

begin

  default clock is rising_edge(clk);

  psl_a : assume always i = 1;

end behav;
```

```shell
ghdl --synth --std=08 assume0.vhdl > test0.vhdl
ghdl --synth --std=08 --assume-asserts assume0.vhdl > test1.vhdl
```

```shell
$ grep assert test0.vhdl 
  psl_a: assert n13_o = '1' severity warning; --  assume
  psl_a: assert n27_o = '1' severity warning; --  assume
$ grep assert test1.vhdl 
  psl_a: postponed assert n13_o = '1' severity error; --  assert
  psl_a: assert n27_o = '1' severity warning; --  assume
```

If the input netlist only contains an `assert` directive, only one `assert` directive is in the output netlist. Which is converted to exact one `assume` directive when the `--assert-assumes` option is used.  

Is this correct and desired behavior?

doc/using/Synthesis.rst
```diff
@@ -98,6 +98,26 @@ simulation back-ends. Hence, available options for synthesis are the same as for
 
     $ ghdl --synth --std=08 --no-assert-cover my_unit
 
+.. option:: --assert-assumes
+
+  Treat all PSL assertions like PSL assumes. If this option is used, GHDL generates `assume` directives
+  instead each `assert` directive during synthesis. This is similar to the `-assert-assumes`
+  option of Yosys `read_verilog <http://www.clifford.at/yosys/cmd_read_verilog.html>`_ command.
```
cmarqu, line 105:
```suggestion
  Treat all PSL asserts like PSL assumes. If this option is used, GHDL generates an `assume` directive
  for each `assert` directive during synthesis. This is similar to the `-assert-assumes`
  option of Yosys' `read_verilog <http://www.clifford.at/yosys/cmd_read_verilog.html>`_ command.
```

doc/using/Synthesis.rst
```diff
@@ -98,6 +98,26 @@ simulation back-ends. Hence, available options for synthesis are the same as for
...
+
+  Example::
+
+    $ ghdl --synth --std=08 --assert-assumes my_unit
+
+.. option:: --assume-asserts
+
+  Treat all PSL assumes like PSL asserts. If this option is used, GHDL generates `assert` directives
+  instead each `assume` directive during synthesis. This is similar to the `-assume-asserts`
+  option of Yosys `read_verilog <http://www.clifford.at/yosys/cmd_read_verilog.html>`_ command.
```
cmarqu, line 115:
```suggestion
  Treat all PSL assumes like PSL asserts. If this option is used, GHDL generates an `assert` directive
  for each `assume` directive during synthesis. This is similar to the `-assume-asserts`
  option of Yosys' `read_verilog <http://www.clifford.at/yosys/cmd_read_verilog.html>`_ command.
```

src/ghdldrv/ghdlsynth.adb
```diff
@@ -128,6 +128,10 @@ package body Ghdlsynth is
       P ("    Neither synthesize assert nor PSL");
       P ("  --no-assert-cover");
       P ("    Cover PSL assertion activation");
+      P ("  --assert-assumes");
+      P ("    Treat all PSL assertions like PSL assumes");
+      P ("  --assume-asserts");
+      P ("    Treat all PSL assumes like PSL assertions");
```
cmarqu, line 134:
```suggestion
      P ("    Treat all PSL asserts like PSL assumes");
      P ("  --assume-asserts");
      P ("    Treat all PSL assumes like PSL asserts");
```
(otherwise it would have to be "assumptions" too)

doc/using/Synthesis.rst
tmeissner: Changed as suggested.

src/ghdldrv/ghdlsynth.adb
tmeissner: Changed as suggested.

doc/using/Synthesis.rst
tmeissner: Changed as suggested.

tmeissner:
@tgingold @eine I think, the PR is ready to merge now.

tgingold:
Thanks!

PR #1578, opened 2021-01-03
Print more information on non fatal warnings

antonblanchard:
Make non fatal warnings print the same one line of context
that fatal ones do, eg:

```
../../src/ieee2008/numeric_std-body.vhdl:3036:7:@1102465ns:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
in process .core_tb(behave).soc0@soc(behaviour).processor@core(behave).icache_0@icache(rtl).itlb_update
```

### DIFF
--- a/src/grt/grt-lib.adb
+++ b/src/grt/grt-lib.adb
@@ -30,6 +30,11 @@ with Grt.Severity;
 with Grt.Options; use Grt.Options;
 with Grt.Fcvt;
 with Grt.Backtraces;
+with Grt.Signals;
+with Grt.Rtis_Addr;
+with Grt.Threads;
+with Grt.Processes;
+with Grt.Rtis_Utils;
 
 package body Grt.Lib is
    --procedure Memcpy (Dst : Address; Src : Address; Size : Size_T);
@@ -52,8 +57,16 @@ package body Grt.Lib is
                         Loc : Ghdl_Location_Ptr)
    is
       use Grt.Severity;
+      use Grt.Signals;
+      use Grt.Rtis_Addr;
+      use Grt.Threads;
+      use Grt.Processes;
+      use Grt.Rtis_Utils;
+
       Level : constant Integer := Severity mod 256;
       Bt : Backtrace_Addrs;
+      Proc : Process_Acc;
+      Proc_Rti : Rti_Context;
    begin
       Report_S;
       Diag_C (Loc.Filename);
@@ -92,6 +105,15 @@ package body Grt.Lib is
          Error_E_Call_Stack (Bt);
       elsif Level >= Grt.Options.Backtrace_Severity then
          Save_Backtrace (Bt, 2);
+         Proc := Get_Current_Process;
+         if Proc /= null then
+            Proc_Rti := Get_Rti_Context (Proc);
+            if Proc_Rti /= Null_Context then
+               Diag_C ("in process ");
+               Put (Get_Error_Stream, Proc_Rti);
+               Newline_Err;
+            end if;
+         end if;
          Grt.Backtraces.Put_Err_Backtrace (Bt);
       end if;
    end Do_Report;


antonblanchard:
I basically copied what the fatal error path does, I guess it should go into a common function.

tgingold:
Isn't similar to the existing --backtrace-severity=LEVEL option ?  Maybe it should be more advertized!

antonblanchard:
I did stumble over that option, but it gives me no output. I wonder if the backtracing isn't working:

```
# ./core_tb --backtrace-severity=warning | head -30
...
../../src/ieee2008/numeric_std-body.vhdl:3036:7:@0ms:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
../../src/ieee2008/numeric_std-body.vhdl:3036:7:@0ms:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
../../src/ieee2008/numeric_std-body.vhdl:3036:7:@0ms:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
```

but the assert path gives me a line of useful info:

```
./core_tb --assert-level=warning |head -20
...
../../src/ieee2008/numeric_std-body.vhdl:3036:7:@0ms:(assertion warning): NUMERIC_STD.TO_INTEGER: metavalue detected, returning 0
./core_tb:error: assertion failed
in process .core_tb(behave).soc0@soc(behaviour).processor@core(behave).icache_0@icache(rtl).itlb_lookup
```

I've tried the Fedora 33 version of ghdl as well as the `ghdl/ghdl:buster-llvm-7 ghdl` image. I also tried building my own with libbacktrace.a.

eine:
@antonblanchard would you mind providing a minimal reproducer that illustrates the feature? Then, I'll test it on multiple distros (Debian, Ubuntu and Fedora).

antonblanchard:
@eine this shows the issue I'm hitting: [backtrace-test.tar.gz](https://github.com/ghdl/ghdl/files/5774208/backtrace-test.tar.gz)

I try to use `backtrace-severity=warning`, but get no backtrace (I presume it's meant to):

```
ghdl -c --std=08 test_tb.vhdl test.vhdl -e test_tb
./test_tb --backtrace-severity=warning
../../src/ieee2008/numeric_std-body.vhdl:2025:7:@109345ns:(assertion warning): NUMERIC_STD."/=": metavalue detected, returning TRUE
../../src/ieee2008/numeric_std-body.vhdl:2025:7:@109355ns:(assertion warning): NUMERIC_STD."/=": metavalue detected, returning TRUE
...
```

My patch in this bug gives us one line of context, enough in this case to sort out what is going on:

```
../../src/ieee2008/numeric_std-body.vhdl:2025:7:@106715ns:(assertion warning): NUMERIC_STD."/=": metavalue detected, returning TRUE
in process .test_tb(behave).check_process
```

eine:
@antonblanchard, I bet you did all your tests with LLVM backend. Running the tests with the three backends (Debian and Fedora) shows that `--backtrace-severity=warning` produces a different output for mcode, or for LLVM/GCC. See: https://github.com/eine/ghdl/runs/1678810852?check_suite_focus=true.

According to `configure`, `--with-backtrace-lib` needs to be used for LLVM, which we don't do when building the containers, and I don't think it's done in Fedora packages (/cc @sharkcz). However, for GCC it should be used automatically. Furthermore, if you built it explicitly...

tmeissner:
> According to `configure`, `--with-backtrace-lib` needs to be used for LLVM, which we don't do when building the containers, and I don't think it's done in Fedora packages (/cc @sharkcz). However, for GCC it should be used automatically. Furthermore, if you built it explicitly...

Are there any comprehensible reasons to not use ˋ--with-backtraceˋ in container builds. Would be useful also ther, I assume.

eine:
> Are there any comprehensible reasons to not use ˋ--with-backtraceˋ in container builds? Would be useful also ther, I think.

Honestly, until today I was not aware that we were missing that feature. I knew libbacktrace was not used, but I thought it was also useful for debugging crashes. So, I agree, it'd be nice to add it. We need to find the package name and the location in each container, though.

PR #1583, opened 2021-01-05
mv dist/* scripts/

umarcor:
This PR includes some fixes related to recent merges:

- `graphviz` is added to the container where the docs are built.
- ~~https://github.com/ghdl/ghdl/commit/156c2fe69ef9a08084196721cc13df8ea2a694e6 is partially reverted for CI to pass.~~ The CI issue when autodocumenting pynodespy is fixed.
- `recommonmark` is removed from doc requirements, because it is not used at the moment.
- A text header is added to the index of the website.
- MSYS2 recipes are modified to avoid collision when installing a nightly package on top of a system package. This allows users to pick bleeding-edge and matching GHDL, libghdl and pyGHDL.
- `dist` was added to `.gitignore` because it is used by Python packaging. However, that subdir did already exist and contained most of the scripts used in CI. The content of `dist` is now moved to `scripts` for avoiding conflicts.


tgingold:
Thanks!

PR #1587, opened 2021-01-07
Add docstrings and @export to pyGHDL.libghdl interface functions.

Paebbels:
# Changes
* handle `str` to `c_char_p` conversion in wrapper functions of `pyGHDL.libghdl`.
  * simplified code in `pyGHDL.lsp`
* Move CDLL init code into function `def _get_libghdl_path():`
* Added decorator `EnumLookupTable` to precalculate enumeration literal name tables
* Added type variables (`TypeVar`) to mirror types from with Ada code. e.g. `NameId` is internally an integer
* Added export decorator to list functions and classes in `__all__`. These members get auto documented if docstrings are available. Variables are added manually, because a decorator can't be used.
* transformed callable objects stored in variables (function pointers) to wrapper functions. This allows to add docstrings and an export decorator.
* added docstrings for handwritten Python adapter code.  
  auto generated code will be covered in a separate PR.
* let `pnodespy.py` generate `@export` in front of classes
* typo fixes in 3 ADS files

# For separate Pull Requests
* cache modifications for callables from ctypes
  * implement a closure or decorator (proof-of-concept is done and tested)
* let `pnodespy.py` generate documented code in auto generated files from Ada sources
* Other simplifications for `pyGHDL.libghdl`


tgingold:
I'm globally ok.

But for the ctypes interface, I think we should cache the object when there is argstypes and/or restype, like:

    func = libghdl.errorout__memory__get_error_record
    func.argstypes = [c_int32]
    func.restype = Error_Message

Maybe create a variable with `__proto_` prefix. WDYT ?

For bytes to/from str, I am not fully convinced if it applies to a frequently used function.

Paebbels:
In a first round, I'm trying to understand the API. So I'm writing docstrings and add types to the interface. I needed this branch and PR to get documentation built. It's also for you and @umarcor to follow the current state.

I'm fully happy with my work (hence the missing PR description), so I'll do multiple iterations.
For example accessing variables in libghdl is to complicated...

**For `ctypes` and setting parameters**  
I'll investigate how to improve:
1. option is a closure
2. option is to convert the API to a class  
   &rArr; this could also solve the variable problem, because I can abstract the variable with a property.

**For `str` <-> `bytes`:**
Do you have an example function for me to check?

umarcor:
> **For `str` <-> `bytes`:**
> Do you have an example function for me to check?

Not sure if this is relevant: https://github.com/ghdl/ghdl-cosim/blob/master/vhpidirect/shared/pycb/utils.py#L37-L45

tgingold:
The most relevant function is Get_Identifier.  I am a bit worried as the function does more than the Ada function.  I am not opposed to define a wrapper like Get_Identifier_str.

In the language server, I would have hoped to avoid conversions.  The IO streams are bytes, which is then converted to unicode (by JSON), and then converted back to bytes for ghdl...  A lot a conversions.

Paebbels:
> I am a bit worried as the function does more than the Ada function.

I understand this concern, but for now I have only seen Python working with `str` and libghdl (C interface) using of cause `char*` with is then `bytes` in Python. So conversion `str` to `char*` must be done and can't be avoided at either libghdl or another layer.

As I currently haven't found these patterns:
1. `str` &lArr; `char*` &larr; libghdl &larr; `char*` &lArr; `str`, nor
2. while:
   &emsp; `char*` := libghdl::func( `char*` )

I think libghdl can then offer a Python friendly interface and hide conversions and encodings internally. I also think we're doing here a job that ctype could have done for us. Or can it do the job for us?

tgingold:
Ok, we will revisite the issue if there are performance problems.

Paebbels:
@tgingold please review these changes. I'm splitting it here so the changes is not growing to much.

Paebbels:
Addendum: the PR has now a summary description :)

tgingold:
I am still worried about str/bytes conversion.
The problem is that utf-8 may not be correct.  VHDL sources used ISO 8859-1.  So I fear we will get an exception if we try to use pyGHDL on a source file with 8859-1 identifiers.

We will see.

Paebbels:
yes, lets see. Anyway now it's a central place where to change it :).

Xiretza:
Is there a good reason why this PR changed all indentation in python files from 4 spaces (which is the standard according to [PEP 8](https://www.python.org/dev/peps/pep-0008/#indentation)) to tabs?

tgingold:
I don't think so!

umarcor:
I think those were Patrick's editor settings, because auto-generated files do have spaces.

umarcor:
Since this is automatically modified by black, I'll wait for the outcome of #1643 before doing any manual fix.

Paebbels:
We should add an `.editorconfig` file, so my editor would follow different rules for the GHDL project.

PR #1585, opened 2021-01-06
Integrating Language Server tests into pynit

Paebbels:
This PR integrates the language server tests from https://github.com/ghdl/ghdl-language-server into the unittest infrastructure in `testsuite/pyunit` under a new namespace `lsp`.

# Changes
* Moved test files from ghdl-language-server repository to this repository under `testsuite/pyunit/lsp`
* Implemented a Testcase class derived from `unittest.TestCase`
* Fixed possible bugs in LS:
  * [ ] Fixed workspace: A function is called, so must be `Get_Date_State`. See 5eb1878
  * [ ] Fixed references: `Get_Is_Ref` has only 1 parameter. See 0c90f42

# Identified Problems / Anomalies
* [ ] Test 005 apears 2x, but there is no 006, which one to rename from 005 to 006?
* [ ] How to handle hardcodes paths like `file:///home/tgingold/work/vhdl-language-server/tests/files/adder.vhdl`?

# Running Tests with pytest
**Note:** I can currently run these tests from within PyCharm, but I have problems in PowerShell. Running pytest in PowerShell goes into a busy loop or similar. But pytest can still be aborted wih a Ctrl+C. Next I'll test from CMD and MinGW64.

**Running single test in a testcase:**
```bash
cd testsuite
export GHDL_PREFIX=.....
export GHDLLS=.....
python -m pytest -rA pyunit/lsp/LanguageServer.py::Test002_Coverage::test_Request_Response
```

**Running single testcase:**
```bash
cd testsuite
export GHDL_PREFIX=.....
export GHDLLS=.....
python -m pytest -rA pyunit/lsp/LanguageServer.py::Test002_Coverage
```

**Running all tests:**
```bash
cd testsuite
export GHDL_PREFIX=.....
export GHDLLS=.....
python -m pytest -rA pyunit
```

PR #1590, opened 2021-01-10
Bind Python to libghdl via decorator

Paebbels:
This is a follow up on #1587.

# Changes
* A new decorator to create binding functions from Python to libghdl using annotations (type hints), ctypes and dynamic function compilation at startup.

# Dependency
For testing, a working lsp testsuite in `pyunit` would be needed.
Work started in #1585.

### DIFF
--- a/pyGHDL/dom/Common.py
+++ b/pyGHDL/dom/Common.py
@@ -37,8 +37,8 @@
 """
 from pydecor import export
 
-from pyGHDL                import GHDLBaseException
-from pyGHDL.libghdl        import LibGHDLException, errorout_memory
+from pyGHDL import GHDLBaseException
+from pyGHDL.libghdl import LibGHDLException, errorout_memory
 
 __all__ = []
 
@@ -61,5 +61,6 @@ class GHDLMixin:
             for i in range(errorCount):
                 print(errorout_memory.Get_Error_Message(i + 1))
 
-            raise DOMException("Error in libghdl.") \
-                from LibGHDLException("libghdl: Internal error 2.")
+            raise DOMException("Error in libghdl.") from LibGHDLException(
+                "libghdl: Internal error 2."
+            )
--- a/pyGHDL/dom/DesignUnit.py
+++ b/pyGHDL/dom/DesignUnit.py
@@ -41,19 +41,24 @@ This module contains all DOM classes for VHDL's design units (:class:`context <E
 """
 from pydecor import export
 
-from pyVHDLModel.VHDLModel import Entity        as VHDLModel_Entity
-from pyVHDLModel.VHDLModel import Architecture  as VHDLModel_Architecture
-from pyVHDLModel.VHDLModel import Package       as VHDLModel_Package
-from pyVHDLModel.VHDLModel import PackageBody   as VHDLModel_PackageBody
-from pyVHDLModel.VHDLModel import Context       as VHDLModel_Context
+from pyVHDLModel.VHDLModel import Entity as VHDLModel_Entity
+from pyVHDLModel.VHDLModel import Architecture as VHDLModel_Architecture
+from pyVHDLModel.VHDLModel import Package as VHDLModel_Package
+from pyVHDLModel.VHDLModel import PackageBody as VHDLModel_PackageBody
+from pyVHDLModel.VHDLModel import Context as VHDLModel_Context
 from pyVHDLModel.VHDLModel import Configuration as VHDLModel_Configuration
 
 from pyGHDL.libghdl import utils
 from pyGHDL.libghdl.vhdl import nodes
 
 from pyGHDL.dom._Utils import NodeToName, GetIirKindOfNode
-from pyGHDL.dom._Translate import GetExpressionFromNode, GetSubtypeIndicationFromNode, GetGenericsFromChainedNodes, GetPortsFromChainedNodes, \
-    GetDeclaredItemsFromChainedNodes
+from pyGHDL.dom._Translate import (
+    GetExpressionFromNode,
+    GetSubtypeIndicationFromNode,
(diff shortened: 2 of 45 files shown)


Paebbels:
@umarcor and @tgingold here is the decorator code for creating binding function at startup mostly based on type hints specified on a Python function.

Usage looks like this:
```py
@BindToLibGHDL("name_table__get_identifier_with_len", context=dict(NameId=NameId))
def Get_Identifier(string: str) -> NameId:
  """
  Get or create an entry in the name table.

  :param string: String to create or lookup.
  :return:       Id in name table.
  """
```

I want to change this:
* Reduce the usage of `context=`.  
  `BindToLibGHDL` will know all `TypeVars` for types in libghdl as defined in `pyGHDL.libghdl._types`.
* Allow chaining of decorators e.g. with `export`.

tgingold:
This is too complex I fear.
Either because the implementation can be simplified or either because there is no other way to implement.

If the implementation cannot be simplified, I need to understand what does it provide.
Do a type checker need to execute python (and therefore decorator) to check the types ?
If we just want to dynamically check the number of arguments, an assertion would be enough.

tgingold:
I am not sure this is the best example.  Python function Get_Identifier has to call get_identifier_with_len.  So there is an extra parameter that is not easy to guess.

Paebbels:
> This is too complex I fear.
> Either because the implementation can be simplified or either because there is no other way to implement.

Yes I admit, it has a little of complexity if one is not used to write decorators or similar.

> If the implementation cannot be simplified, I need to understand what does it provide.

Besides the obvious (creating an adapter) it does this:
* all is executed just once per subprogram you decorate from Python to libghdl.
* it does automatically, what you would write per hand. The function above creates this intermediate code:
  ```py
  def Binder(string:str) -> NameId:
    string = string.encode("utf-8")
    return libghdl.name_table__get_identifier_with_len(c_char_p(string), len(string))
  ```
* It's taking the type information and how to bind from Python's function "prototype".
* The order of parameters must match to libghdls paramaters
* If return type is None, a procedure is assumed and no `return` statement is generated
* If a parameter is a string (`str`), it does 3 things:
  1. add a conversion line
  2. insert an addition parameter as position n+1 for the length. This assumes, all subprograms in libghdl have the form `..., char*, int, ...`
  3. it calls 2 convert functions for parameter n and n+1: `c_char_p(...)` and `len(...)`
* If a function has a return type of `str`, then `restype` of ctypes is set to `c_char_p` and the result is encoded to (currently) utf8
* If a type is a TypeVar, it's reading the "inner type" which is mostly int for libghdl.

Besides `c_int32` and `c_char_p` does libghdl need other interface types?

> I am not sure this is the best example.  Python function Get_Identifier has to call get_identifier_with_len.  So there is an extra parameter that is not easy to guess.

As shown above, I can handle all this :). Of cause I assume some argument order and style, but I think no C-like interface programmer puts a length for an arrays not directly behind the array itself, right?

> Do a type checker need to execute python (and therefore decorator) to check the types ?

Types are not checked at runtime. Types in Python are just hints to:
* run a type checker like a unit test
* implement code that checks types, e.g. by decorators
* give IDEs hints for auto completion
* In case of GHDL and Sphinx, it's also used to write types into the documentation
* ...

I use this information (accessible via `obj.__annotations__` to read what and how I can bind to libghdl.

> If we just want to dynamically check the number of arguments, an assertion would be enough.

tgingold:
I am not against a decorator to create the adapter, I am not favorable to a complex decorator.  The first version of the decorator was much simpler!

I am not sure that handling bytes/str conversion should be automatic.  It hides the complexity.  The conversion is not something obvious so it must clearly appear. I prefer to handle that manually as there aren't many functions about them.

> Types are not checked at runtime. Types in Python are just hints to:
> 
> 	• run a type checker like a unit test
> 	• implement code that checks types, e.g. by decorators
> 	• give IDEs hints for auto completion
> 	• In case of GHDL and Sphinx, it's also used to write types into the documentation
> 	• ...
> I use this information (accessible via obj.__annotations__ to read what and how I can bind to libghdl.

Which of those features need execution ?  Is it only the first one ?

To sum up, I am not sure about the advantage of your second version of the decorator.  For IDE and Sphinx, I think it brings no new features.  For execution, it checks the number of arguments but that could also be checked by a linter.

Paebbels:
I've now pushed the reworked decorator. The other one is saved on my disk.

The solution does this:
* use a decorator and an inner function to bind to `libghdl`
  * uses `*args` to call the callable from ctypes
* get type information from `__annotations__` to set configuration options to ctype callable for parameter types and return types.
* no type conversion
* docstrings come from original function prototype
* typehins are also used for auto generated documentation
* should be usagable in `pnodespy.py`

Paebbels:
@tgingold I cleaned this working branch, so only changes for `pyGHDL.libghdl` are left (including some pnodespy changes).

I'm using this approach since a while in my tests and it works very well.
What is missing from my side to get this idea and enhancement merged?

The original (complex) decorator was changed to a simpler variant as discussed above.

---------
This PR does not cover:
* Handing of float values as discussed in Gitter for `Get_Fp_Value`.
* Improvements to `pyGHDL.libghdl` initialization
* Improvements to `pyGHDL.libghdl.utils`

tgingold:
I think I understand the code, so that's ok with me.
BUT: some tests are failing!

Paebbels:
I'll check the tests, maybe I broke things as I split the original branch into smaller pieces for review.

Paebbels:
The current pipeline hast 18 successful jobs, 4 are still running.

I fixed this:
* Reformatted Python code with black
* Added a missing Exception class - has been lost while splitting the old PR into multiple parts
* Removed some VHDL code lines from simple package test file, because types and subtypes are not yet supported.

PR #1595, opened 2021-01-13
update license headers

umarcor:
It turned out that the license header fix applied in https://github.com/ghdl/ghdl/commit/6496bc9045f9c07c71ca92dbe9d7afbc107c7d7d was also needed in almost all the sources of the repo.

- I'm not sure about pushing this without confirmation. @tgingold, does any source need to be "GCC"?
- I'm neither sure about "*If not, write to the Free Software Foundation, 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA*". Should it be replaced with "*If not see <http://www.gnu.org/licenses/>*"?


tgingold:
Yes, we also need to change the text.  Let's use the neutral text from https://www.gnu.org/licenses/gpl-3.0.html (but still using v2):

```
    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 2 of the License, or
    (at your option) any later version.

    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.

    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <https://www.gnu.org/licenses/>.
```


Sorry, that's a lot of files to change again.

umarcor:
> Sorry, that's a lot of files to change again.

No worries. I ain't crazy enough for doing it manually. I used sed and I will use some script now :wink:

umarcor:
@tgingold, it's done now.

tgingold:
Thanks!  Maybe the largest commit!

PR #1602, opened 2021-01-16
 libraries: unuse 'openieee', rename enable_openieee to enable_gplcompat

### LINKED ISSUE #1060: Replace IEEE libraries with VHDL-2019 packages (Apache license)
> [December 30, 2019 10:32 PM](https://gitter.im/ghdl1/Lobby?at=5e0a6cdd0fd3413f4c7df244)
>
> On 20.12.2019, [IEEE Std. 1076-2019](https://standards.ieee.org/standard/1076-2019.html) was officially released.
> In parallel, all [VHDL packages](https://opensource.ieee.org/vasg/Packages) (std.standard, std.env, ..., ieee.std_logic_1164, ...) have been release as **OPEN SOURCE** under Apache License, 2.0.

> [December 30, 2019 10:50 PM](https://gitter.im/ghdl1/Lobby?at=5e0a71282d574c7d1ee3c7a6)
>
> The good with VHDL-2019 packages and Apache license is as follows:
> 1. GHDL project creates a clone of the original sources and keeps the master branch.
> 2. GHDl implements [LCS-2016-061](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_061) to support conditional analysis
> 3. GHDl creates a modification branch, which adds conditional analysis statements into the sources.  
>    Thus, we can use VHDL-2019 sources for 1987, 1993, 2002, 2008, 2019

This issue is to track the migration of current non free IEEE sources, and openieee, to the open source [VHDL packages](https://opensource.ieee.org/vasg/Packages).

I think that the natural procedure would be to fork/clone the repo to this organization and add it as a git submodule in subdir `libraries`. Then, fix the makefiles accordingly.

However, there have been previous reluctances to using submodules in this org. @tgingold, what do you think?

IANAL, but this should allow us to get rid of the `--gpl` build option, and the distinction when packaging for Debian.

umarcor:
Close #1060.
Ref: [December 10, 2020 6:06 AM](https://gitter.im/ghdl1/Lobby?at=5fd1acbf0697c1210db1246f)

This PR removes `openieee` (except for `upf` packages), since current ieee libraries are GPL compatible. At the same time, configure option `enable-openieee` is renamed to `enable-gplcompat`. Now, that option decides whether VITAL libraries (95 and 00) are included or not. The configure script, makefiles and CI scripts are updated accordingly.

@tgingold, did you actually expect to remove the sources from `libraries/openieee`, or do you want to keep them in the repo, although unused?


tgingold:
I am not sure what to do with openieee.  math_real should be much faster than the official package, because it uses the C functions.
The numeric_std/std_logic_1164 should also be faster (there are much less calls).

So at least I would keep them.

umarcor:
Maybe we can build openieee by default and let users include it in their designs through `library openieee;`, if they want to?

tgingold:
Please no.  No need to create incompatibilities!

umarcor:
Then, can we build both of them and provide a mechanism for the user to choose which one to use? That is, make `--openieee` an option of GHDL (the CLI or libghdl)?

Paebbels:
Isn't `openieee` a IEEE flavor like we have `synopsys`, `mentor`, ...? So it can be selected with `-ieee=openieee`.

umarcor:
@Paebbels, I think that mentor is deprecated, only synopsys is supported starting with the next release:

- https://github.com/ghdl/ghdl/blob/master/NEWS.md
- https://ghdl.github.io/ghdl/using/InvokingGHDL.html#cmdoption-ghdl-fsynopsys
- https://ghdl.github.io/ghdl/using/InvokingGHDL.html#cmdoption-ghdl-ieee

Apart from that `--ieee=` modifies some files only, not the IEEE library. `openieee` is a replacement for the wole IEEE lib. So, `synopsys` can be used together with `numeric_std` and `math_real`, and those can be from either `ieee` or `openieee`.

However, since `--ieee` is now deprecated for that usage, we can now repurpose it. `--ieee=open`?

tgingold:
That could be an option, but I am not sure it is worth.

umarcor:
Then, we can have this merged and think about that in a future PR. Adding it as an option to `--ieee` does not conflict with these changes; it's done on top of it.

PR #1614, opened 2021-01-17
Improvements to the UVVM pre-compile scripts.

Paebbels:
**Changes:**
* Fixed `ccv` vs. `vvc` typo.
* Read `component_list.txt` also in Bash-based script.
* `--halt-on-error` to work for Bash-based script.
* Handle `\n` / `\r\n` problem.

It was tested with uvvm/uvvm@51f9287.
A problem was reported to UVVM as uvvm/uvvm#130.

Thanks to @tpcorrea.


Paebbels:
This is ready for merge.

PR #1615, opened 2021-01-18
testsuite: add suite_driver.sh

### LINKED ISSUE #723: Top-most testsuite does not pass options to individual testsuits
**Description**
Test suits like `sanity` support parameters like `-k` for *keep going*.  
This option cannot be set from top-most test suite file.

**Expected behavior**
Allow setting of global (useful) options to all test suits from top-most test suite.

------------
Is this code incomplete?

```bash
for opt; do
  case "$opt" in
      [a-z]*) tests="$tests $opt" ;;
      *) echo "$0: unknown option $opt"; exit 2 ;;
  esac
done
```

... because, `$opt` is not declared, thus it's empty.

umarcor:
Close #723

`testsuite/gna/testsuite.sh`, `testsuite/sanity/testsuite.sh`, `testsuite/synth/testsuite.sh` and `testsuite/vpi/testsuite.sh` are almost identical. In this PR, `testsuite/suite_driver.sh` is created, which takes the target suite name as the first argument. Then, those four testsuites are rewritten for calling the suite_driver. At the same time `testsuite/testsuite.sh` is updated for calling the suite_driver directly. BTW, positional arguments to be passed to the suite_driver are supported:

```sh
# All test suites, no opts
./testsuite.sh

# Test suite gna only, no opts
./testsuite.sh gna

# Test suite gna only, opt keep-going
./testsuite.sh gna -- -k

# All the test suites, opt keep-going
./testsuite.sh -- -k
```

### DIFF
--- a/testsuite/gna/testsuite.sh
+++ b/testsuite/gna/testsuite.sh
@@ -1,53 +1,5 @@
 #! /bin/sh
 
-# Driver for a testsuite.
-
 set -e
 
-# This is the only place where test dirs are specified.  Do not duplicate this
-# line
-dirs="*[0-9]"
-
-failures=""
-full=n
-
-for opt; do
-  case "$opt" in
-  -k | --keep-going)  full=y ;;
-  --dir=*) dirs=`echo $opt | sed -e 's/--dir=//'` ;;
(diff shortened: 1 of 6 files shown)


tgingold:
Thanks!

Issue #1115, opened 2020-01-25
macOS binary is unusable under the current version of the operating system


misterakko:
**Description**
The downloadable binary (pre-built release) at https://github.com/ghdl/ghdl/releases cannot be used on updated Mac computers.

**Expected behaviour**
./bin/ghdl should be an executable. But since it's not signed, the OS refuses to consider it so.
Issue can be easily fixed by distributing as a DMG file instead of a TGZ file, and digitally signing/notarizing  the file with the free tool inside the Xcode IDE. It's not necessary to register with Apple, nor pay any fee.

**How to reproduce?**
E.g. if you try
./ghdl  -a hello.vhdl
You will get a high level alert from the OS saying that the binary is unsigned and unnotarized, and could contain malware. The OS then offers to throw the GDHL executable in the trash.
Execution is stupid with the error message
./ghdl:error: exec error

**Context**
macOS 10.15 "Catalina"
See also https://stackoverflow.com/questions/39811791/mac-os-gatekeeper-blocking-signed-command-line-tool

tgingold:
Humm, is it just a matter of using hdiutil ?

eine:
FTR, currently, in the CI jobs macOS builds use the same script as GNU/Linux builds. There are some differences, but the prodecure is to configure (mcode), make, make install, generate a tarball and run the test suite. No specific macOS tool is used at all. Note that mcode backend is tested only.

Moreover, the macOS tarballs (mcode and llvm) available at [releases](https://github.com/ghdl/ghdl/releases) are contributed manually by @diogratia; unlike GNU/Linux and Windows tarballs, which are uploaded automatically from the CI jobs.

As a result, in the current status, we need @diogratia to voluntarily decide that he/she wants to investigate how to produce DMG files instead of tarballs.

Alternatively, the current CI jobs might be fixed and extended so that DMG files are generated and uploaded automatically, instead of requiring manual builds. See:

- #867: about enabling libghdl on macOS jobs. Users have reported that it works for them, so it must be some issue with the CI setup.
- https://github.com/eine/ghdl/commits/osx-hdiutil: an attempt to use hdiutil to extract the installer of GNAT, instead of using 7z installed through brew (which we are doing ATM).

Resources to reproduce/investigate:

- This is the CI job in GitHub Actions: https://github.com/ghdl/ghdl/blob/master/.github/workflows/push.yml#L29-L42.
[...]

PR #1630, opened 2021-02-04
ci: run macOS 10.15 and 11.0

eine:
As commented in https://github.com/ghdl/ghdl/issues/1115#issuecomment-772951318, currently macos-11.0 is in "private preview" mode. We'll keep this as a draft until we can have it tested.


eine:
Since August, *"macOS-11 is generally available for all customers"* (actions/virtual-environments#4060).

PR #1635, opened 2021-02-07
doc: update instructions about GCC and GNAT Community installation

rodrigomelo9:
I updated instructions about GNAT Community Edition installation.

### DIFF
--- a/doc/development/building/GCC.rst
+++ b/doc/development/building/GCC.rst
@@ -9,7 +9,7 @@ GCC backend
 
 * GCC (Gnu Compiler Collection)
 * GNAT (Ada compiler for GCC)
-* GCC source files. Download and untar the sources of version 4.9.x, 5.x, 6.x, 7.x, 8.x, 9.x or 10.x.
+* GCC source files. Download and untar the sources of version 4.9.x, 5.x, 6.x, 7.x, 8.x, 9.x or 10.x (`GCC mirror sites <https://gcc.gnu.org/mirrors.html`__).
 
 .. HINT :: There are some dependencies for building GCC (``gmp``, ``mpfr`` and ``mpc``). If you have not installed them on your system, you can either build them manually or use the ``download_prerequisites`` script provided in the GCC source tree (recommended): ``cd /path/to/gcc/source/dir && ./contrib/download_prerequisites``.
 
--- a/doc/development/building/index.rst
+++ b/doc/development/building/index.rst
@@ -17,9 +17,14 @@ further details.
 
 .. IMPORTANT::
    Since GHDL is written in `Ada`, independently of the code generator you use, a compiler is required. Most GNU/Linux package
-   managers provide a package named ``gcc-ada`` or ``gcc-gnat``. Alternatively, `GNU Ada compiler`, `GNAT GPL`, can be downloaded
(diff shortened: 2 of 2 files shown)


rodrigomelo9:
I was helping with a Centos7 compilation, using GCC as backend, and I would like to perform some more improvements to the instructions:
* Add a comment after *`gcc-ada` or `gcc-gnat`*: *(which could be outdated)*
* *can be downloaded anonymously from* what about to remove anonymously? It sounds **illegal** when not.

And what about to provides an example link to the GCC sources into `development/building/GCC.html#build-gcc`? I used https://codeload.github.com/gcc-mirror/gcc/tar.gz/releases/gcc-8.3.0.

rodrigomelo9:
I realized that only `x86 GNU Linux (64 bits)` is available at https://www.adacore.com/download, so also the comment *(2014, or later; for x86, 32 or 64 bits).* must be updated. You can't choose a version, I mean, only 2020 is available there.

eine:
> I realized that only `x86 GNU Linux (64 bits)` is available at [adacore.com/download](https://www.adacore.com/download), so also the comment _(2014, or later; for x86, 32 or 64 bits)._ must be updated. You can't choose a version, I mean, only 2020 is available there.

There is a link at the botton to other versions: https://www.adacore.com/download/more. There, you can see that packages did need to be extracted until 2017. Then, executable installers are distributed.

Therefore, I think that this PR should account for both. That is, instead of modifying the text, add a clarification/note. 

> Add a comment after gcc-ada or gcc-gnat: (which could be outdated)

Agree. We should clarify why would users want to use AdaCore's GNAT instead of gcc-ada or gcc-gnat.

> can be downloaded anonymously from what about to remove anonymously? It sounds illegal when not.

Anonymously means that users don't need to register in order to get the link, which is still an annoying practice in many companies. I think it is a common term for services/permissions. Anyway, feel free to reword it.

> And what about to provides an example link to the GCC sources into development/building/GCC.html#build-gcc? I used codeload.github.com/gcc-mirror/gcc/tar.gz/releases/gcc-8.3.0.

That's just a mirror. I don't think it's necessary to explain users how to get GCC sources. If someone is building GHDL with GCC backend, they must already be experienced enough so that finding the sources is negligible. We do already provide a link to GCC site in https://ghdl.github.io/ghdl/development/building/index.html.

rodrigomelo9:
> Therefore, I think that this PR should account for both. That is, instead of modifying the text, add a clarification/note.

I add a HINT clarifying the situation. Let me know if HINT is ok for that or what to use.

> Anonymously means ...

I know, but I prefer to be clear (changed by *without registration*) and avoid to sound illegal.

> That's just a mirror. I don't think it's necessary to explain users how to get GCC sources.

This part (GCC installation) was one of the main problems of the user that I helped. I had no present how to get the GCC sources, so I checked the CI of the project to get a link (which was no straight forward, due that the version in the link is a variable in `ci-run.sh`). What about at least a link to the GCC mirrors (https://gcc.gnu.org/mirrors.html)? I only want to improve the user experience when GCC backend is needed. Let me know what you think about and in case of agree, if you prefer a separate PR (this change is in another file, doc/development/building/GCC.rst).

rodrigomelo9:
Hi @eine, I added a link to the *GCC mirror sites* into doc/development/building/GCC.rst.

eine:
Thanks!

PR #1558, opened 2020-12-29
Executing Python unit tests in Parallel Processes

Paebbels:
This PR contains:
* Parallel execution with `pytest` and its plugin `pytest-forked` is now possible.  
  This works around the problem for independent unit testing that `libghdl` can only be initialized once.
* `pytest`s strange/strict naming conventions were disabled via `pytest.ini` in `testsuite/pyunit` to comply with normal unittest naming rules.
* `testsuite.sh` was adjusted to run from within `testsuite` directory.
* All 5 testcases are now found and executed.

PR #1643, opened 2021-02-10
pyGHDL: auto-format python code

Xiretza:
**Description**

This was already introduced in #1439, but disabled again in #1558 for reasons not completely obvious to me. One issue that was brought up was the formatting of the generated bindings, but simply running black after generating the file should fix that.

cc @eine @umarcor @Paebbels

### DIFF
--- a/src/vhdl/Makefile
+++ b/src/vhdl/Makefile
@@ -79,31 +79,37 @@ vhdl-elocations_meta.adb: vhdl-elocations_meta.adb.in vhdl-elocations.ads $(DEPS
 ../../pyGHDL/libghdl/vhdl/nodes.py: $(DEPS) $(PNODESPY)
 	$(RM) $@
 	$(PNODESPY) $(NODES_FLAGS) libghdl-nodes > $@
+	python -m black $@
 	chmod -w $@
 
 ../../pyGHDL/libghdl/vhdl/nodes_meta.py: $(DEPS) $(PNODESPY)
 	$(RM) $@
 	$(PNODESPY) $(NODES_FLAGS) libghdl-meta > $@
+	python -m black $@
 	chmod -w $@
 
 ../../pyGHDL/libghdl/std_names.py: $(PNODESPY) ../std_names.ads
 	$(RM) $@
 	$(PNODESPY) $(NODES_FLAGS) libghdl-names > $@
+	python -m black $@
 	chmod -w $@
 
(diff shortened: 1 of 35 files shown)


Xiretza:
This codacy thing should probably be tamed a little, some of its complaints are downright ridiculous. Multi-line docstrings have to start in both the first and the second line at the same time, class declarations and their docstrings have to be separated by a blank line (never seen that in any codebase), and somehow it even managed to invent some whitespace to complain about.

src/vhdl/Makefile
```diff
@@ -79,31 +79,37 @@ vhdl-elocations_meta.adb: vhdl-elocations_meta.adb.in vhdl-elocations.ads $(DEPS
 ../../pyGHDL/libghdl/vhdl/nodes.py: $(DEPS) $(PNODESPY)
 	$(RM) $@
 	$(PNODESPY) $(NODES_FLAGS) libghdl-nodes > $@
+	python -m black $@
```
umarcor, line 82: I don't think this is desirable. Running `python -m black` or `black` will work. We can optionally provide a list of subdirs to include/exclude, as a target in the main makefile.

Xiretza: The problem is that the generated files are set `chmod -w` in the next line, so they can't be formatted after being generated. Or are you suggesting not formatting these at all?

umarcor:
The motivation for disabling it was it didn't match Patrick's preferences for `pyGHDL.dom`, and Tristan didn't have a strong feeling about it. But it was not removed because we might want to discuss (as we will do here) where to enforce it.

I don't mind using it everywhere, in some places only, or nowhere.

> This codacy thing should probably be tamed a little, some of its complaints are downright ridiculous.

Yes, it needs to be tweaked, but there are hundreds of options so it's not easy/fast. As a rule of thumb, just ignore those errors unless you read something which is obviously meaningful for you.

Xiretza:
> The motivation for disabling it was it didn't match Patrick's preferences for `pyGHDL.dom`, and Tristan didn't have a strong feeling about it.

Currently `pyGHDL.dom` contains a bunch of tabs instead of spaces, so while there might be some legitimate differences in coding style preferences, this is an example of an obvious mistake that would've been caught if black had been in place.

Personally I don't care which particular style is used as long as it's used and enforced consistently.

tgingold:
Thanks!

PR #1644, opened 2021-02-11
Add gitignore files for testsuite

Xiretza:
**Description**

After running the testsuite, a bunch of generated files clutter up the `git status` output. Adding only a few lines of `.gitignore` hides them all.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,7 +52,19 @@ version.tmp
 e~*.lst
 
 # GHDLs testsuite
-testsuite/get_entities
+/testsuite/get_entities
+/testsuite/test_ok
+/testsuite/gna/**/*.out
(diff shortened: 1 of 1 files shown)


Paebbels:
Some tools have problems in applying ignore rules from subdirectories.

I recommend to list these ignore rules centrally in the root directories `.gitignore`.

Xiretza:
Hm, sounds messy... I've never heard of that, which tools are those?

tmeissner:
I would also prefer to have one "global" .gitignore file in the root directory of the project.

tgingold:
Thanks!

For individual tests, I prefer to clean in the test rather than having .gitignore entries.

PR #1651, opened 2021-02-17
update for LLVM 11.1

sharkcz:
We need this in Fedora, because we already have the LLVM 11.1 release candidates available in the distro. A similar update for LLVM 12 will follow soon.


eine:
Thanks!

PR #1653, opened 2021-02-17
vendors: support deprecated param names

abyszuk:
**Description**
Commit c0f0539c introduced a change in two command line parameters
present in all scripts used to compile vendor libraries.
That change breaks any automated infrastructure that compiles vendor libraries.
I can't just update it because my users have different versions of GHDL and I have no control over that.

This PR brings back the old param names to maintain backwards compatibility.


Paebbels:
I would be ok with maintaining old parameter names until the next GHDL release, but deprecated options shouldn't be listed in the help output.

abyszuk:
OK, I'll fix that in a moment!

eine:
Thanks!

PR #1656, opened 2021-02-19
CI: print diff of necessary python formatting

Xiretza:
While looking at the CI output of 58c808dd, I noticed that black only complains about formatting being necessary, but doesn't show the necessary changes - the `--diff` option does just that.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -99,7 +99,7 @@ jobs:
         git diff --exit-code
 
     - name: Check if python follows code formatting standards
-      run: python -m black --check pyGHDL
+      run: python -m black --check --diff pyGHDL
 
 #
 # GPL


eine:
It was done on purpose, because the list of files it's complaining about is listed: https://github.com/ghdl/ghdl/runs/1933667638?check_suite_focus=true#step:7:9. Users/contributors should run black locally anyway, instead of "fixing" style issues manually. Therefore, I'm not sure about the usefulness of adding `--diff`. Yet, it is harnless. I'll wait for you ack to this comment and I'll merge if you still find it useful.

Xiretza:
Alright, that makes sense. Unrelated: is it possible that the macOS tests are a little flaky? While CI is green in this PR, the CI run for the branch in my repo failed on `synth12`: https://github.com/Xiretza/ghdl/actions/runs/581288861

eine:
TBH, the LLVM build on macOS was added very few days ago. We didn't have time yet to know whether it's flaky or not. Hence, it might be a legit issue, but I cannot confirm.

PR #1660, opened 2021-02-22
revert "log stderr of make to a file and use tail to show last 1000 lines"

Xiretza:
This reverts commit 30ee7a32b64c94f1ac0c8e3b5068894b6ea8b556 (#601), cc @eine .

This workaround masks the return value of `make`, causing runs to appear to fail during install, even though build failed already. Since travis is no longer used for CI, it's probably no longer necessary anyway.

### DIFF
--- a/scripts/ci-run.sh
+++ b/scripts/ci-run.sh
@@ -308,10 +308,7 @@ build () {
   #--- make
 
   gstart "[GHDL - build] Make"
-  set +e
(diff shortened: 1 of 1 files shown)


eine:
You are correct. I think this was added due to some limitation in Travis CI. We can try removing it. I did also find the weird behaviour you explained, but I did not dive to find the root.

PR #1666, opened 2021-03-01
VPI: support loading multiple libraries

marlonjames:
**Description**
Store library information for all `--vpi=` options and load each in the order they are passed in.
Library file names are stored in a simple singly-linked list.


umarcor (approved):

tgingold:
Thanks!
Bonus point if you add a tail pointer, so that append is O(1).

Issue #1648, opened 2021-02-15
Adding VHPI support


marlonjames:
As far as I am aware, there are no open-source VHDL simulators that have more than rudimentary VHPI support.
I think that the VHDL community would benefit from having a mature implementation of VHPI in an open-source project, and GHDL is the obvious target.

Is the GHDL project receptive to contributions towards that goal?

VHPI is divided into the different capability sets, which helpfully provide subsets of the full functionality, and can be worked in turn. My initial goal would be loading a library and calling the registration function(s), and the minimal implementation as noted in the LRM:

> A minimal implementation of the VHPI need only provide the function interface described in Clause 23 and
Annex B, with none of the capability sets described in this subclause being implemented by the tool. In such a minimal
implementation, calls to functions would, in most cases, raise a VHPI error condition.

To be followed by some of the capability sets, in whatever order seems most useful/requested, with an eye towards cocotb support (e.g., `vhpiProvidesDebugRuntime`, `vhpiProvidesHierarchy`).

A related issue: I've started working on a set of VHPI compliance tests. The plan is to have a freely available testsuite that simulators can run to test their VHPI implementation. I'm not sure how that should integrate with testing of a GHDL implementation.

tgingold:
Of course, contributions for VHPI are welcome!

There is a base to implement it: Avhpi, which mimic the VHPI interface but with much less dynamic allocation.

umarcor:
> A related issue: I've started working on a set of VHPI compliance tests. The plan is to have a freely available testsuite that simulators can run to test their VHPI implementation. I'm not sure how that should integrate with testing of a GHDL implementation.

See https://github.com/VHDL/Compliance-Tests.

umarcor:
> There is a base to implement it: Avhpi, which mimic the VHPI interface but with much less dynamic allocation.

Ref: https://ghdl.github.io/ghdl-cosim/vhpidirect/grt.html

marlonjames:
What's the preferred path for the VHPI thunk?

1. separate `libghdlvhpi`
2. combine VPI/VHPI to `libghdlpli` similar to what other sims do, and hard link from `libghdlvpi` for backwards compatibilty?

tgingold:
Yes, combine in the same library.  I am not sure we even need to rename it.

PR #1670, opened 2021-03-04
VHPI: minimal implementation

marlonjames:
**Description**
Adds a very minimal implementation of VHPI as a base to build from.

This provides VHPI library loading and VHPI error support.
Calling most vhpi_* functions will raise a VHPI error condition.

**Further comments**

Supports loading multiple libraries, and supports optional user-specified entry point.
If no entry point is supplied, it will look for `vhpi_startup_routines[]` array (This is what Aldec simulators do).
For now I didn't implement support for tabular registry files (LRM 20.2.2) for foreign models/applications.

`vhpi_check_error` is implemented to inform VHPI applications of the unimplemented status of the other commands.

I tried to strip this down to the minimum for the initial PR, but it's still quite large.

Any and all feedback is appreciated!

xref #1648


tgingold:
Thanks!  Eager to see more...

eine (commented):
@marlonjames congratulations for this so nice PR. You not only added this very wanted feature, but you also updated all the different build/test scripts in the repo. That's awesome!

PR #1674, opened 2021-03-05
Include directory structure proposal.

m-kru:
I would propose following policy for storing and installing include header files.

1. They are always in separate `include` directory, for example in `src/synth/include`.
2. In terms of the installation they are put under `ghdl/`. For exampple, 
```
ghdl/
├── synth
│   ├── gates.h
│   └── synth.h
├── vhpi_user.h
└── vpi_user.h
```
The include would be `#include "ghdl/synth/synth.h"`.

The Makefile probably can be improved, but this is just proposal.


tgingold:
Ok, but do you have the corresponding change for the yosys plugin ?

m-kru:
@tgingold no I don't. Should it be applied only here https://github.com/ghdl/ghdl-yosys-plugin?

tgingold:
Yes.

m-kru:
Ok, I think it would also make sense to do the same for `vhpi` and `vpi`if there are going to be more header files related with them.

umarcor (commented):
IMHO the `synth` subdir adds unnecessary complexity. There are two header files and four total. Even if we add a couple more, having additional subdirs is overkill. Instead, I would merge `synth.h` and `gates.h` in a single file.

I think that renaming ghdlsynth to synth and ghdlsynth_gates to gates is sensible.

m-kru:
@umarcor 
I agree, after some thoughts I think there is no gain in having separate directories within `ghdl` directory. It adds extra complexity to the Makefile and in C symbols are global anyway.  

>  I would merge synth.h and gates.h in a single file.

I guess they are separate because one is auto generated. 
Please note that gates.h is included in synth.h.
I would not merge them.

umarcor:
> I guess they are separate because one is auto generated.
> Please note that gates.h is included in synth.h.
> I would not merge them.

I was thinking about con`cat`ing them during install, in case you wanted to have a single unit (subdir or file) for synthesis. However, after your clarification, I agree with not merging.

m-kru:
How are the tests for vpi organized? If I replace `#include <vpi_user.h>` with `#include <ghdl/vpi_user.h>` they fail. If I do not replace they also fail.

tgingold:
It must be `#include <vpi_user.h>`.  But the path returned by `--vpi-include` should append the `ghdl/` part.

m-kru:
If I understand correctly, another way is to add `ghdl/` to install destination paths in
```Makefile
install.vpi.local: all.vpi
	$(MKDIR) -p include lib
	$(INSTALL_DATA) -p $(GRTSRCDIR)/vpi_user.h include/
	$(INSTALL_DATA) -p $(GRTSRCDIR)/vhpi_user.h include/
```
and then use `#include <ghdl/vpi_user.h>` in tests, the same way it is supposed to be used by the user.

m-kru:
But wait ..., if I was correct then the tests should pass right now.

tgingold:
I think it is important to keep `#include <vpi_user.h>` so that the code is portable among simulators.

m-kru:
Ok, it is all getting a little bit fuzzy, so lets clear it.
There are 2 Makefile targets `install.vpi` and `install.vpi.local`. I am not sure what is the purpose of the second one?
I also do not understand what is the of point `--vpi-include`?

> I think it is important to keep `#include <vpi_user.h>` so that the code is portable among simulators.

Isn't it related with what people have in `PATH`?
I really do not like installing headers to some common include paths like `/usr/local/include`.
It is like delay-action bomb.
Some simulators will have to update `#include <vpi_user.h>` to `#include <ghdl/vpi_user.h>`, but this is very small change.

tgingold:
`install.vpi` is used during `make install`.  `install.vpi.local` is used to locally copy the vpi include files in the object directory so that it is possible run the testsuite without installing.

The option `--vpi-include-dir` (sorry I forget the `-dir`) is used while compiling vpi plugin to get the include path.

If it contains the `/ghdl` subdir at the end, users will be able to use `#include <vpi_user.h>` while the `vpi_user.h` file be in a `/ghdl` subdir (like `/usr/local/include/ghdl`).

m-kru:
Ok, I I understand what you have written, but I do not understand why is this complexity with `--vpi-include-dir` needed.

In `*nix` there are some paths in the include `PATH` by default, for example `/usr/local/include`.
If some tool `ABC` wants to install headers in one of the the default paths there are 2 possible ways.
1. If this is single header, then name it `ABC.h` and install it directly in the `/usr/local/include`.
2. If these are multiple header files, then create directory `ABC` within `/usr/local/include` and put headers into that directory. `/usr/local/include/ABC/foo.h`, `/usr/local/include/ABC/bar.h`.

If user wants some custom installation path then it is specified in the Makefile with variable or during configure step.
However, GHDL (the tool) is aware of the vpi include directory, and user can "ask" it to get that directory path.
I am sure it serves some purpose, however at first glance it looks very suspicious, like something was structured in a intricate way.

tgingold:
You can consider it is useful at least for the testsuite.  The vpi tests need to have `vpi_user.h` in the include path and the corresponding library. Instead of configuring the testsuite, a test can invoke ghdl to get the paths and the libraries.

m-kru:
Ok, the tests pass. The only change that would need to be applied to the ghdl-yosys-plugin is probably:
```diff
diff --git a/src/ghdl.cc b/src/ghdl.cc
index 413d371..de06f1f 100644
--- a/src/ghdl.cc
+++ b/src/ghdl.cc
@@ -27,7 +27,7 @@ USING_YOSYS_NAMESPACE

 #ifdef YOSYS_ENABLE_GHDL

-#include "ghdlsynth.h"
+#include "ghdl/synth.h"

 using namespace GhdlSynth;
```

@tgingold you can check and merge if you like, or give feedback what you do not like.

Xiretza:
I think `install.vpi.local` needs to be adjusted (or the include path setup for tests should be reworked completely), currently tests fail because they can't find the header if ghdl is not installed.

m-kru:
@Xiretza which tests?

Xiretza:
`gna issue1226` is the first one that fails, but all VPI tests should be affected. `ghdl --vpi-compile` adds `-I/path_to_ghdl_repo/include/ghdl` to the compiler arguments, which doesn't exist.

m-kru:
@Xiretza where are these tests? Why they are not catched by the CI?

Xiretza:
This specific one is in `testsuite/gna/issue1226/`, and it's not caught because the CI runs the tests in a docker image where ghdl is already installed, which is honestly pretty unconventional. Maybe that should be changed, but that's mostly an unrelated issue. For now, `install.vpi.local` should probably just install the headers to `include/ghdl/` instead of `include/`.

m-kru:
Ok this is all getting messy. I would opt for one uniform way. Locally or globally headers are always placed in `.../include/ghdl/` directory. However, the include path returned by GHDL or added with `--vpi-compile` should be `../include`, and within C sources there should always be `#include <ghdl/foo.h>`.

Xiretza:
I think `#include <vpi_user.h>` is expected to work as-is, without a tool prefix - so either `--vpi-compile` has to include the `ghdl/` directory in the include path, or the header needs to be placed in `/usr/include`, outside the new directory structure.

m-kru:
Why `ghdl --vpi-compile` serves as an wrapper for vpi compilation at all?

If it is supposed that user will always compile vpi via `ghdl --vpi-compile` and never directly via gcc or clang, then why is it installed in one of the system default include paths at all?
If it is intended to be used anyhow user wants, then it should be placed in `include/ghdl`. I know it breaks backward compatibility, but it will be done how it should be done from the beginning. By the way, project of any size can be easily adjusted to this change within few seconds:
```shell
find . -name *.c -exec sed -i 's/#include <vpi_user.h>/#include <ghdl\/vpi_user.h>/' {} \;
```
.

Xiretza:
`vpi_user.h` is a standard header and not specific to GHDL, so it doesn't really make much sense to install it under a ghdl prefix and force people to either add it to their include path or specialize all the source files to GHDL.

tgingold:
Thanks.  I will fix the fallouts.

tgingold:
vpi_user.h is almost standard, but it may differ from simulator to simulator, and every simulator provides its version.

So I think it makes sense to install it in a ghdl specific subdirectory.  Just to avoid a clash.

tgingold:
> Why ghdl --vpi-compile serves as an wrapper for vpi compilation at all?

At least for the testsuite.

> By the way, project of any size can be easily adjusted to this change within few seconds:
> 
> find . -name *.c -exec sed -i 's/#include <vpi_user.h>/#include <ghdl\/vpi_user.h>/' {} \;

It's indeed not difficult, but better to avoid it!

eine:
> This specific one is in `testsuite/gna/issue1226/`, and it's not caught because the CI runs the tests in a docker image where ghdl is already installed, which is honestly pretty unconventional.

GHDL's testsuite was initially designed for being executed after `make install`. Installing GHDL in the same environment where it was built did lead to dependency issues. This is because build dependencies are typically a superset of runtime dependencies. Therefore, in CI, GHDL is built on one environment and it is tested on a different environment. On Linux, containers are used. On Windows, different jobs are used. IMHO, what we do unconventionally is using tarballs instead of DEB/RPM packages on Linux. Using containers is the only solution for testing multiple distributions (Debian, Ubuntu, Fedora and Arch).

Later, some features where added to the makefiles because some packagers wanted to run the testsuite in-place and before `make install`. I consider that to be a partial/smoke testing, not a robust execution of the testsuites.

src/ghdldrv/ghdlvpi.adb
```diff
@@ -37,7 +37,7 @@ package body Ghdlvpi is
       --  Compute install path
       Ghdllocal.Set_Exec_Prefix_From_Program_Name;
 
-      return Ghdllocal.Exec_Prefix.all & Directory_Separator & "include";
+      return Ghdllocal.Exec_Prefix.all & Directory_Separator & "include/ghdl";
```
umarcor, line 40: `/` might be problematic on Windows. However, we currently test on MSYS2 only, so it should work.

Xiretza: Tristan already fixed it on master: 2036d713a73a91a8014d0d63201d795807e0c924

PR #1683, opened 2021-03-12
scripts/vendors: add missing workdir argument

corna:
ghdl invocation in scripts/vendors was missing the `workdir` argument, causing the output files to be in the wrong directory.
Tested with compile-xilinx-vivado.sh.


Paebbels:
OMG. I did so many tests when updating it and I missed this....

Paebbels:
@tgingold I reviewed it, you can merge it.

tgingold:
Thanks

Issue #572, opened 2018-05-19
Documentation broken reference


gritbub:
I'm cleaning up the documentation at my [fork][1]. Besides editing simple typos and grammatical nuances, I also came across a broken reference **_:ref:'GHDL_options'_** that is used 3 times in [Invoking GHDL][2]. Look at [readthedocs][3] to see that the link is broken.

I can't figure out where this is supposed to refer back to. As a side note, **_GHDL_options_** (without the reference tag **_:ref:_**) is found 3 times in [ghdl.texi][4].

I am new to git. I'm also new to these file types (**_.md_**, **_.rst_**, and also **_readthedocs_** stuff), so I apologize in advance if I'm doing something wrong or just plain missing something obvious.

[1]: https://github.com/gritbub/ghdl
[2]: https://github.com/ghdl/ghdl/blob/master/doc/using/InvokingGHDL.rst
[3]: http://ghdl.readthedocs.io/en/latest/using/InvokingGHDL.html#analysis-a
[4]: https://github.com/ghdl/ghdl/blob/master/doc/ghdl.texi

Paebbels:
`*.md` are MarkDown files. A simple markup language with simple formatting: headlines, bold/italic font face, identifier markup, embedded source code, tables, images and links. It's used by e.g. StackOverflow. 

`*.rst` are ReStructured Text files. A translation tool like Sphinx can translate it into an intermediate format and from there it can produce HTML, PDF, ePub reader and other output formats. It's more powerful than MarkDown. ReadTheDocs is a cloud service that offers Sphinx compile time and a webhosting service `*.readthedocs.org`. More over it can host different versions (and languages) of a documentation.

@tgingold was `:ref:'GHDL_options'` intended to reference to `ghdl --options-help` ?

```
C:\> ghdl.exe --options-help
Main options:
  --work=LIB         use LIB as work library
  --workdir=DIR      use DIR for the file library
  -PPATH             add PATH in the library path list
  --std=87/93/00/02/08  select vhdl 87/93/00/02/08 standard
  --std=93c          select vhdl 93 standard and allow 87 syntax
  --[no-]vital-checks  do [not] check VITAL restrictions
Warnings:
  -Wbinding          warns for component not bound
  -Wreserved         warns use of 93 reserved words in vhdl87
  -Wlibrary          warns for redefinition of a design unit
  -Wvital-generic    warns of non-vital generic names
  -Wdelayed-checks   warns for checks performed at elaboration
[...]

PR #1694, opened 2021-03-20
Psl warn uncovered

### LINKED ISSUE #1690: PSL cover - failure
**Description**
When a PSL cover point is not covered during simulation, at the end, error (failure) is reported.

**Expected behaviour**
Uncovered PSL cover points shall not report error (failure), this is also how commercial simulators behave.
The rationale behind is:
 - One test-bench with many cover points scattered through TB and RTL.
 - Many tests. Each test is separate run of simulator (different procedure), but the same test-bench. E.g. as in Vunit.
 - Each test indends to verify (and cover) different scenario, situation.
 - After whole regression, functional coverage from all tests is merged together to find out if there were some cover points
   not hit by any test (e.g. in some HTML report)..

GHDL behavior in this situation makes all test fail (unless there is a test which covers all cover points, which is unrealistic),
making the PSL implementation completely unusable for any decently complex testbench.

**How to reproduce?**

```vhd :file: top.vhd
Library ieee;
use ieee.std_logic_1164.all;

entity tb_top is
end entity;


architecture tb of tb_top is

    signal a,b,c : std_logic;
    signal clk_sys : std_logic;

begin
	
    gen_clock_proc : process
    begin
        clk_sys <= '1';
        wait for 5 ns;
        clk_sys <= '0';
        wait for 5 ns;
    end process;
	
	test_proc : process
	begin
	    a <= '0';
	    b <= '0';
	    c <= '0';
	    wait until rising_edge(clk_sys);
	    a <= '1';
	    wait until rising_edge(clk_sys);
	    a <= '0';
	    b <= '1';
	    wait until rising_edge(clk_sys);
	    b <= '0';
	    c <= '1';
	    wait until rising_edge(clk_sys);
	    
        std.env.finish;
	end process;


    --  psl default clock is rising_edge(clk_sys);

    -- psl cov_simult_a_b_c : cover {a = '1' and b = '1' and c = '1'};

end architecture tb;
```

```sh :image: ghdl/ghdl:buster-gcc-8.3.0
ghdl -a --workdir=build --std=08 -fpsl top.vhd
ghdl -e --workdir=build --std=08 tb_top
ghdl -r tb_top
```

> NOTE: `:file:` and `:image:` identifiers are specific to [issue-runner](https://github.com/1138-4EB/issue-runner). We suggest to use these, since it allows continuous integration workflows to automatically test the MWE. Using `ghdl/ghdl:*` docker images to run the MWEs ensures that the latest available GHDL is used.

> NOTE: Large files can be uploaded one-by-one or in a tarball/zipfile.

**Context**

- OS: Ubuntu
- Origin:
  - [x] Built from sources: 8eb6eb35ae475be271cef614af0256282286606b

**Additional context**

Simulation log:
simulation finished @40ns
top.vhd:45:12:@40ns:(psl cover failure error): sequence not covered

Blebowski:
Resolves failing psl cover points which were not covered during simulation. Added Wpsl-uncovered option.

- Issue fixed:
- [Close #1690](https://github.com/ghdl/ghdl/issues/1690)
- 

- [ ] CONSIDER modifying the docs, at least in the TODO, if your contribution is relevant to any of the content.

- [ ] DO make sure that the build is successful. See [ghdl/ghdl#572 (issuecomment-390466024)](https://github.com/ghdl/ghdl/issues/572#issuecomment-390466024).

**Further comments**

:heart: Thank you!


eine:
Please, replace `Issue URL` with `Close #1690`. That will allow the other issue automatically when this is merged.

src/options.adb
```diff
@@ -289,17 +295,20 @@ package body Options is
...
 --    P ("           simulation.  LEVEL is note, warning, error,");
 --    P ("           failure or none");
       P ("Extensions:");
-      P ("  -fexplicit         give priority to explicitly declared operator");
-      P ("  -frelaxed-rules    relax some LRM rules");
-      P ("  -C  --mb-comments  allow multi-bytes chars in a comment");
-      P ("  --bootstrap        allow --work=std");
-      P ("  --syn-binding      use synthesis default binding rule");
-      P ("  -fpsl              parse psl in comments");
+      P ("  -fexplicit        give priority to explicitly declared operator");
```
eine, line 306: Please, keep the alignment or add one space, instead of removing it.

src/flags.ads
```diff
@@ -158,6 +158,10 @@ package Flags is
    --  If True, disp original source line and a caret indicating the column.
    Flag_Caret_Diagnostics : Boolean := False;
 
+   -- If True, warning is printed for each uncovered PSL cover points at the
+   -- end of simulation
+   Flag_Psl_Warn_Uncovered : Boolean := False;
```
eine, line 163: It seems that this flag decides whether to report or not, instead of selecting the severity level (warning or other). Therefore, I would suggest `Flag_Psl_Report_Uncovered`.

src/options.adb
```diff
@@ -289,17 +295,20 @@ package body Options is
       P ("  -Wspecs            warns if a all/others spec does not apply");
       P ("  -Wunused           warns if a subprogram is never used");
       P ("  -Werror            turns warnings into errors");
+      P ("  -Wpsl-uncovered    Prints warning after simulation for each");
+      P ("                     uncovered PSL cover point.");
+      P ("                     Only valid with '-fpsl'");
```
tmeissner, line 300: What's with inline PSL? I think, the option is valid for all kinds of PSL, isn't it?

eine:
> CONSIDER adding a unit test if your PR resolves an issue.

You can just copy the testbench from https://github.com/ghdl/ghdl/issues/1690. See, for instance, https://github.com/ghdl/ghdl/tree/master/testsuite/gna/issue1371 or https://github.com/ghdl/ghdl/tree/master/testsuite/gna/ticket19.

tmeissner (changes requested):

Blebowski:
I have processed all the comments (stil not pushed), and I am adding a test for the feature.

Btw. how do I run single test? I am trying: `./testsuite.sh gna` in "testsuite" folder. However, it runs all gna tests.
How do I specify single test? (To debug it locally before pushing). I tried `./testsuite.sh gna/issue1690` , however
it does not work.

eine:
Go into the specific subdir, and execute the `./testsuite.sh` in there:

```sh
cd testsuite/gna/issue1690
./testsuite.sh
```

Blebowski:
Thanks, done and pushed.
Btw. how do you make the test fail? Or what is it needed for a test to fail? Exit code 1 of analyze, elaborate ?
I was thinking to do something like grep the simulation log for "sequence uncovered" and making the test fail if it is not present (the feature does not work).

eine:
> Btw. how do you make the test fail? Or what is it needed for a test to fail? Exit code 1 of analyze, elaborate ?
> I was thinking to do something like grep the simulation log for "sequence uncovered" and making the test fail if it is not present (the feature does not work).

Strictly, we don't want the test itself to fail, because we might be executing other tests afterwards, so we want to handle the failures. Nevertheless, you can explicitly check multiple cases:

```sh
#! /bin/sh

. ../../testenv.sh

export GHDL_STD_FLAGS='--std=08'
analyze top.vhd
elab_simulate tb_top
clean

export GHDL_STD_FLAGS='--std=08 --assert-level=warning'
analyze top.vhd
elab_simulate tb_top
clean

export GHDL_STD_FLAGS='-Wpsl-uncovered --std=08 -fpsl --assert-level=warning'
analyze top.vhd
elab_simulate_failure tb_top
clean

export GHDL_STD_FLAGS='-Wpsl-uncovered --std=08 -fpsl'
elab_simulate tb_top > log_file
cat log_file
grep -e "sequence not covered" log_file
clean

echo "Test successful"
```

tgingold:
Two comments:

* It shouldn't be a warning option as this is not for a warning emitted during analysis

* I think it should be a runtime option instead of a compile time option.  Those options are handled in package `grt.options`

I understand my comments will need some rewrite...

Blebowski:
@tgingold Sure, I can change it to run-time option, it makes more sense.

But, if I change it to be run option, I dont have the information available at analysis/elaboration time, therefore I need to adjust psl code generation to check for global flag at run time, instead of pre-generating the PSL code. Is that correct?
Btw, do you know why does the pipeline throws permission denied failure? Did I forget something?

eine:
> Btw, do you know why does the pipeline throws permission denied failure? Did I forget something?

`git update-index --chmod=+x path/to/failing/bash/script`

tgingold:
Yes that's correct: you simply need to modify ghdl_psl_cover_failed (in grt-lib.adb) to report nothing (or just an info).

Blebowski:
All resolved, only docs are not updated. Btw. "run options" are not available in any CLI, only in docs. (e.g. --psl-report). It would be good to add those somewhere.

eine:
> All resolved, only docs are not updated. Btw. "run options" are not available in any CLI, only in docs. (e.g. --psl-report). It would be good to add those somewhere.

Did you check `ghdl --options-help` and/or `ghdl help elaborate` or `ghdl help run`?

`--psl-report` might be missing, indeed; yet there is place in CLIs to show/add it.

tgingold:
It's OK.  You just have to update the doc/.

tgingold:
Thanks!  But please, don't forget the doc!

Blebowski:
Thanks Tristan, I did not update it yet, I had other work. I will update the doc once I manage to build it with Sphinx on my PC. I would keep the related issue open, just as a reminder of the doc.

eine:
> I will update the doc once I manage to build it with Sphinx on my PC

1. Extract this tarball (https://github.com/ghdl/ghdl/blob/master/.btd.yml#L9) to `doc/_theme`.
2. Install the requirements: `pip3 install -r doc/requirements.txt`
3. Execute `make html` in subdir `doc`.

You will get warnings about pyGHDL. Ignore them. You don't care about that part of the docs.

Blebowski:
Updated docs too.

Btw @tgingold , Is it correct that HTML doc is now the main doc source? Or should I maintain also tex documentation?

eine:
@Blebowski,

> Updated docs too.

Please, open a new PR.

> Is it correct that HTML doc is now the main doc source?

The restructuredtext/Sphinx sources in `doc` are the main and only doc source. All outputs (HTML, LaTeX/PDF and man) are generated from those same sources.

> Or should I maintain also tex documentation?

In case you mean https://github.com/ghdl/ghdl/blob/master/doc/ghdl.texi with "tex documentation", that is also an artifact. You can `make ghdl.texi` for having it updated: https://github.com/ghdl/ghdl/blob/master/doc/Makefile#L33-L35. The reason for `ghdl.texi` to be tracked in git, is because it's used when building GHDL with the GCC backend (https://github.com/ghdl/ghdl/blob/2036d713a73a91a8014d0d63201d795807e0c924/Makefile.in#L271), hence, it avoids users needing Sphinx for building GHDL.

PR #1709, opened 2021-04-03
Vhdl parse: resync external name

Blebowski:
Resynchronize on end of External name. Search for colon or semicolon

- [ ] CONSIDER modifying the docs, at least in the TODO, if your contribution is relevant to any of the content.

**Further comments**

:heart: Thank you!

### DIFF
--- a/src/vhdl/vhdl-parse.adb
+++ b/src/vhdl/vhdl-parse.adb
@@ -449,6 +449,26 @@ package body Vhdl.Parse is
       end loop;
    end Resync_To_End_Of_Interface;
 
+   -- Search for next colon (likely before subtype_indication).
+   -- Others detect when we are totally lost. Semi-colon likely
+   -- at the end of line, Double_Greater at the end of external
+   -- name.
+   procedure Resync_To_End_Of_External_Name is
+   begin
+      loop
+         case Current_Token is
+            when Tok_Colon
+               | Tok_Semi_Colon
+               | Tok_Begin
+               | Tok_Eof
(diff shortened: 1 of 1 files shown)


src/vhdl/vhdl-parse.adb
```diff
@@ -449,6 +449,21 @@ package body Vhdl.Parse is
...
+   -- Search for next colon (likely before subtype_indication) or
+   -- semicolon, likely at end of line, if we are totally lost.
+   procedure Resync_To_End_Of_External_Name is
+   begin
+      loop
+         case Current_Token is
+            when Tok_Colon
+               | Tok_Semi_Colon =>
+               exit;
+            when others =>
```
tgingold, line 461: You need to stop the resync at least on `Tok_EOF` otherwise it will abort.
Ideally, you'd better to list the token which much be skipped rather that the token that will stop the resync.  That's safer.

tgingold (changes requested):

src/vhdl/vhdl-parse.adb
Blebowski: True, I will add EOF. OTOH if you specify what to skip, then you need to list everything what possibly user could put there to be ignored. Would it be OK to have Tok_Colon, Tok_Semicolon, Tok_Eof, Tok_Begin, Tok_End, Tok_Double_Greater. These will be likely there at some next place (start of next statement region, end of external name) ?

tgingold:
Thanks!

PR #1716, opened 2021-04-08
Psl wave dump

Blebowski:
This PR prepares for dumping of PSL assertions into GHW file
(so far WIP, AVHPI extension is needed as well as GHW format extension).

Following is done:
  - Separate RTI created for PSL assertions
  - Introduce two counts for each PSL assertion: Start count and Finish count (finish count corresponds to existing count)
    The rationale behind, is to see whether assertion was even started (e.g. if doing assert never on sequence, it is good
    to see, that the sequence even started to run, otherwise, there is bunch of assertions with count = 0 which are passed,
    but in reality we have no clue whether they even started). Further refinement is not done.
    Start count is incremented upon leaving start state by assertion NFA.
  - Introduce "simplified psl state" variable. Simplified state can be one of:
      - Inactive - No NFA state, or only NFA start state is active
      - Running - Any of transient NFA states are active (has priority over Inactive)
      - Failed - For assert/assume statements if final NFA state is active (has priority over Running)
      - Covered - For cover statements if final NFA state is active (has priority over Running)
  - Fix Get_Path_Name when using --trace-processes with PSL assertions. PSL RTI has parent, allowing querying path.
  - Adjust PSL report to dump both counters, as well as using new RTI type.
  - Extend VHPI definitions to match 2008 LRM, updated deprecated suffices.
  - Define custom VHPI kinds for PSL assertions (preparing ground for querying assertions via AVHPI from grt-wave).
  - Enforce encoding of enumerated types in grt-avhpi.ads to match LRM.
    (Here I am not sure whether compiler will then guarantee that enum encoding will then match what is in LRM)...

At the moment, the code builds, and runs regression just fine, OTOH, I am not sure I did not break something.
I will appreciate any comments.
Please note, that doing this change "one thing at a time" is not easy for relatively big change,
therefore I packed multiple things into single PR. Current state seemed good enough that it could be reviewed/merged.

PS: I will consider extending the docs with description of psl-report format.

**Further comments**

:heart: Thank you!


tgingold:
Thanks!

Maybe we will need to refine the state variable (splitting running into started + checking), but that's a detail.

tgingold:
Well, if there is nothing wrong in a patch and no regression, I simply merge it.
Style issues could be fixed later.

> Btw. if I use ADAs "for" + "use" (as in the VHPI) for enum type values, does it mean GNAT
> will satisfy binary compatibility with C implementations of VHPI ?

You also need to add `pragma Convention (C, My_Enum_Type);` so that the layout is the same.

> I saw some issue that someone is working on VHPI interface for GHDL. If I understand it correctly,
> then he can directly use enumerations defined in grt-avhpi in his VHPI implementation when doing
> "case" on values passed from VHPI library? (At least that was the reason I added the "for" "use" clause).

Yes.  I am not fully sure we want a direct interface.  But we will see.

> To the state splitting:

So we agree it would be nice to refine the states.

> I examined NFAs generated by sequence implication like the one above, and I did not find a way
> how to distuiguish between "started" and "checking", since NFAs already loose this information
> upon their construction.

Not completely.  The first checking state has Get_Active_State to True.  That's used by synthesis.

> It seems to me, that anytime NFA reaches final state it is:
> 
> 	• failed - for assert and assumes (both never and always)
> 	• covered - for cover.
> Is that correct?

Yes, that's correct.

> Also, I did not know how to find a state which would be "passed". Maybe, if we know that
> transfering to final state means "failed", then if we dont go to final state, but we are in any state
> which has edge to a final state (in other words, just the condition which would make the assertion
> fail does not hold) we can say it "passed".

Yes there is no passed state.  But usually you don't really care because the assertion should always pass.
You want to know if it fails.  And you'd like to know if it is covered.  For synthesis, a cover is automatically generated for each assertion.
And cover means triggered (ie in the checking state).

> Still, this does not make it possible to distuiguish between "started" and "checking".

If you can see in a waveform that the assertion is in the checking state, that should be enough for the user.


Alternative idea:
as this is an NFA, multiple states are possible: the assertion can check for multiple traces at the same time.
So, instead of defining one state (which has to be maintained), just characterise the NFA: state 1 is the start state, state 2 .. N are the triggering states, states N + 1 .. M are the checking states and state M + 1 the failing state.  At worst, this requires a renumbering of states.  There would be no overhead to maintain the state and at run time, the waveform could be more explicit.
This is just for discussion; I don't want to change your plans now.

PR #1728, opened 2021-04-17
VHPI: add tracing

marlonjames:
**Description**
Add VHPI tracing.

Reference #1648.

### DIFF
--- a/src/grt/grt-vhpi.ads
+++ b/src/grt/grt-vhpi.ads
@@ -43,6 +43,9 @@ package Grt.Vhpi is
    type Vhpi_External_Handle is access Vhpi_Internal_Handle;
    pragma No_Strict_Aliasing (Vhpi_External_Handle);
 
+   --  A null handle.
+   Null_External_Handle : constant Vhpi_External_Handle;
+
    type VhpiFormatT is
      (
       VhpiBinStrVal,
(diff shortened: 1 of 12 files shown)


src/grt/grt-vhpi.ads
```diff
@@ -392,13 +386,15 @@ package Grt.Vhpi is
       VhpiSizeConstraint
      );
    pragma Convention (C, VhpiPutValueModeT);
+   for VhpiPutValueModeT'Size use Integer'Size;
```
tgingold, line 389: You don't need to specify the size of enumeration types.  This could contradict/override the pragma convention just above.

marlonjames: At first I didn't specify the size, but compilation failed with this warning:
```
grt-vhpi.adb:648:07: warning: types for unchecked conversion have different sizes
grt-vhpi.adb:648:07: warning: size of "Vhpiputvaluemodet" is 3, size of "Integer" is 32
grt-vhpi.adb:648:07: warning: source will be extended with 29 high order zero bits
```

Is there a better approach for getting the invalid values for Mode that are passed from C?

tgingold:
I would go for the simplest: pass `Mode` as an integer, check `Mode` is in the range (if needed) and then use `VhpiPutValueModeT'Val(Mode)` to get the corresponding enumeration literal.

In general I really avoid `'Valid` and I prefer not to use `Unchecked_Conversion` between scalar types.

PR #1729, opened 2021-04-18
VHPI: improve C enum interop

marlonjames:
**Description**
Improve interop with C enums, per discussion in #1728.


tgingold:
Thanks.

PR #1730, opened 2021-04-19
Fix xilinx vendorlib compilation

bahm-lgs:
**Description**
Compilation of Xilinx Vivado 2019.2 vendor libraries on Windows fails with current version of the PS scripts. This PR includes all changes I needed to make to the compile-xilinx-vivado.ps1 script to run on my machine with ghdl-1.0.0. For release 0.37 compilation is successful without swapping --std=08 and -frelaxed-rules.

- [X] DO make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [X] DO make sure you are making a pull request against the **master branch** (left side). Also you should start *your branch* off *our master*.
- [ ] DO make sure that GHDL can be successfully built. See [Building GHDL](https://github.com/ghdl/ghdl#building-ghdl).

**Further comments**

I have no build environment for GHDL under Windows, so it would be great if someone else could have a closer look to the changes to make sure they are complete and don't break anything.

### DIFF
--- a/scripts/vendors/compile-altera.ps1
+++ b/scripts/vendors/compile-altera.ps1
@@ -151,6 +151,7 @@ $VHDLVersion,$VHDLStandard,$VHDLFlavor = Get-VHDLVariables $VHDL93 $VHDL2008
 
 # define global GHDL Options
 $Analyze_Parameters = @(
+	"--std=$VHDLStandard",
 	"-fexplicit",
 	"-frelaxed-rules",
 	"--mb-comments",
@@ -171,7 +172,6 @@ if (-not ($EnableVerbose -or $EnableDebug))
 $Analyze_Parameters += @(
 	"--ieee=$VHDLFlavor",
 	"--no-vital-checks",
-	"--std=$VHDLStandard",
 	"-P$DestinationDirectory"
 )
 
--- a/scripts/vendors/compile-intel.ps1
+++ b/scripts/vendors/compile-intel.ps1
@@ -151,6 +151,7 @@ $VHDLVersion,$VHDLStandard,$VHDLFlavor = Get-VHDLVariables $VHDL93 $VHDL2008
 
 # define global GHDL Options
 $Analyze_Parameters = @(
+	"--std=$VHDLStandard",
 	"-fexplicit",
 	"-frelaxed-rules",
 	"--mb-comments",
@@ -171,7 +172,6 @@ if (-not ($EnableVerbose -or $EnableDebug))
 $Analyze_Parameters += @(
 	"--ieee=$VHDLFlavor",
 	"--no-vital-checks",
-	"--std=$VHDLStandard",
 	"-P$DestinationDirectory"
 )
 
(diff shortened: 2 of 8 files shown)


umarcor:
Hi Martin! Correct! `-frelaxed-rules` needs to be after `--std=`. See https://ghdl.github.io/ghdl/using/InvokingGHDL.html#cmdoption-ghdl-std and https://ghdl.github.io/ghdl/using/InvokingGHDL.html#cmdoption-ghdl-frelaxed-rules. The bash version seems to be ok: https://github.com/ghdl/ghdl/blob/b48d15631984d7f82dd4c7e91152915025b2c4f4/scripts/vendors/compile-xilinx-vivado.sh#L262-L263.

However, I'm not sure about the other commit, which is a powershell syntax issue I believe.

umarcor:
> I have no build environment for GHDL under Windows, so it would be great if someone else could have a closer look to the changes to make sure they are complete and don't break anything.

CI takes care of it for you. However, should you want to try it, it's pretty straightforward:

- Download `msys2-base-x86_64-latest.sfx.exe` from https://github.com/msys2/msys2-installer/releases/tag/nightly-x86_64 and extract it with, e.g. 7z.
- Open `mingw64.exe` in there and execute `pacman -Syu` in the terminal. It will update and close.
- Open `mingw64.exe` again and install `pacman -S base-devel git mingw-w64-x86_64-toolchain`.
- Go to the location of your modified sources of GHDL and execute `MINGW_ARCH=mingw64 makepkg-mingw --noconfirm --noprogressbar -sCLf` in subdir `scripts/msys2-llvm`.
- Optionally, install the package `pacman -U *.zst` and run the test suites.

That is exactly the procedure we use in CI: https://github.com/ghdl/ghdl/blob/master/.github/workflows/Test.yml#L196-L314.

bahm-lgs:
> However, I'm not sure about the other commit, which is a powershell syntax issue I believe.

Yes, it is. As I am also not a powershell expert, please have someone else look at it who knows, why I might have a problem with those unnamed arguments since they have been changed from type "bool" to "switch". @Paebbels could you maybe comment on this?

Paebbels:
@bahm-lgs what PowerShell version are you using for testing. The mentioned code line works since years in PowerShell 4, 5, 6. I can currently not understand why it should fail.

What parameters did you use for testing when you called the script. Maybe it's a combination I never tested or tested not so excessively.

------
For moving `--std=...` can you please modify the other vendors scripts accordingly (Intel, Altera, ...)  
They all should have the same issue :).

bahm-lgs:
> @Paebbels I am using the default PowerShell version which comes with current Win10 releases, i.e. PS 5.1
> 
> The issue is not present with the scripts from GHDL 0.37, it seems to be related to your changing the Get-VHDLVariables function's parameter types this January .

I just also had problems reproducing the issue again and discovered that there is some caching mechanism in Powershell for the modules. After undoing my changes I had to restart PS to get the error again. I will update my 2nd commit to simply revert the parameter types of Get-VHDLVariables function, so the function calls won't have to be changed in all scripts.

Can someone else reproduce the error with PowerShell 5.1?

tgingold:
Thanks.

Issue #1731, opened 2021-04-21
Resizing signed produces IIR_PREDEFINED_IEEE_NUMERIC_STD_RESIZE_SGN_SGN when synthesizing


umarcor:
Resizing signed seems not to be supported for synthesis yet:

```vhdl
--:file: axis_conv1d9x1.vhd
library ieee;
context ieee.ieee_std_context;

entity axis_conv1d9x1 is
  port (
    s_axis_clk   : in  std_logic;
    s_axis_rstn  : in  std_logic;
    s_axis_data  : in  std_logic_vector(31 downto 0);
    s_axis_valid : in  std_logic;
    s_axis_last  : in  std_logic
  );
end axis_conv1d9x1;

architecture arch of axis_conv1d9x1 is

  signal d: std_logic_vector(31 downto 0);

  signal acca, accb, acca_n, accb_n : signed(31 downto 0);
  signal rlast : std_logic;

begin

  process(s_axis_clk) begin
    if rising_edge(s_axis_clk) then
      if not s_axis_rstn then
        acca <= (others=>'0');
        accb <= (others=>'0');
      elsif s_axis_valid then
        acca <= acca_n;
        accb <= accb_n;
      end if;
    end if;
  end process;

  acca_n <=
    acca + signed(s_axis_data(15 downto 0)) when rlast='0' else
    resize(signed(s_axis_data(15 downto 0)), acca) when ??(s_axis_valid) else
    (others=>'0');

  accb_n <=
    accb + signed(s_axis_data(31 downto 16)) when rlast='0' else
    resize(signed(s_axis_data(31 downto 16)), acca) when ??(s_axis_valid) else
    (others=>'0');

  process(s_axis_clk) begin
    if rising_edge(s_axis_clk) then
      if not s_axis_rstn then
        rlast <= '0';
      else
        rlast <= s_axis_last;
      end if;
    end if;
  end process;

  d <= std_logic_vector(accb(18 downto 3)) & std_logic_vector(acca(18 downto 3));

end architecture;
```

```shell
# ghdl synth --std=08 --out=raw-vhdl axis_conv1d9x1.vhd -e axis_conv1d9x1
axis_conv1d9x1.vhd:37:5: unhandled function: IIR_PREDEFINED_IEEE_NUMERIC_STD_RESIZE_SGN_SGN
[...]

PR #1732, opened 2021-04-21
Add handling of resize(uns, uns) for synthesis

tmeissner:
This PR adds handling of `resize(uns, uns)` for synthesis. A simple test is included, similar to the one of #1731.


tgingold:
But this is not correct.  We need to extract the length of the second argument.

tmeissner:
Yeah, I've seen that. These functions are very misleading IMHO.

Issue #1718, opened 2021-04-09
CLI support for specifying library along with primary unit


GCHQDeveloper560:
**Is your feature request related to a problem? Please describe.**

Many tools allow running a top level design in a library using a dot notation like ``my_lib.my_top``. GHDL doesn't support this CLI syntax and instead requires something like
```
ghdl -e --work=my_lib --workdir=./my_lib my_top
```

**Describe the solution you'd like**

It would be simpler and more similar to other tools if GHDL supported
```
ghdl -e my_lib.my_top
```

**Additional context**
This was discussed by @m-kru and @tgingold in [an Edalize issue](https://github.com/olofk/edalize/issues/229).

m-kru:
I guess `my_lib` can imply `--work=my_lib`, however the question is whether it can also imply the `--workdir` value.

umarcor:
Overall, I think that the feature request is very nice. I want it! Yet, I'm afraid the proposal is not fully correct. As @m-kru said, see the following example:

```vhdl
entity ent is
end ent;

architecture arch of ent is
begin
process begin report "Hello"; wait; end process;
end;
```

```sh
# ghdl -a --work=my_lib ent.vhd 
# ghdl -e ent
error: cannot find entity or configuration ent
C:\msys64\mingw64\bin\ghdl.exe: compilation error
# ghdl -e --work=my_lib ent
```

That is, sources analyzed to specific libraries do not need to be located in separated directories. Those are different features. You can have one or multiple libraries and one or multiple locations for `*.cf` files (which is what `--workdir` is about).

So, I think it is correct to "*`my_lib` can imply `--work=my_lib`*", and leave the `--workdir` option aside (i.e. use `-P` for that).

PR #1733, opened 2021-04-22
Update docs: [[library.]top_unit [arch]]

umarcor:
This PR includes 3 pending updates to the docs:

- All occurrences of `primary_unit` are replaced with `top_unit`, and `secondary_unit` is replaced with `arch`.
- We explain why `-e` is required in `synth files... -e`, even though specifying the top unit name is optional.

Those two points were discussed in [gitter: September 27, 2020](https://gitter.im/ghdl1/Lobby?at=5f70b0ca5efe9467b11c9ac4).

As requested in #1718, `[library.]top_unit` is supported now (see https://github.com/ghdl/ghdl/commit/849a25e02cfb359e3d9313060156b0643495548b). Documenting this is quite challenging. This is what I did:

Commands affected:

- `-e <[options...] [library.]top_unit [arch]>`
- `-r <[options...] [library.]top_unit [arch] [simulation_options...]>`
- `--elab-run <[options...] [library.]top_unit [arch] [simulation_options...]>`
- `-m <[options...] [library.]top_unit [arch]>`
- `--gen-makefile <[options...] [library.]top_unit [arch]>`
- `--gen-depends <[options...] [library.]top_unit [arch]>`
- ` --synth <[options...] [library.]top_unit [arch]>`

Commands NOT affected:

- ` -a <[options...] file...>`
- ` -s <[options...] file...>`
- ` -c <[options...] file... -<e|r> top_unit [arch]>`
- `-i <[options...] file...>`
- ` --synth <[options...] file... -e [top_unit [arch]]>`

@tgingold, would you please review whether the `[library.]` updates are correct?

---

BTW, I fixed an annoying bug that breaks the search. I also changed the style so that each sentence is started in a new line (in the codebase) and not to have lines longer than 120 chars.


Xiretza:
I remember that discussion, thanks for actually implementing it! Personally, I would prefer reading "architecture" instead of the abbreviated "arch" in long-form documentation, but that's just my own opinion.

doc/requirements.txt
```diff
@@ -1,6 +1,6 @@
 -r ../pyGHDL/requirements.txt
 
-sphinx>=3.4.2
+sphinx==3.3.1
```
Xiretza, line 3: Could you add some more information on this bug (like an upstream issue) either to the commit message or as a comment in the file itself?

umarcor:
@Xiretza we can discuss the specific keywords. The relevant modification in this PR is homogenizing. After Tristan reviews the content, replacing `top_unit` and/or `arch` should be easily doable through text-editor replacement (or sed).

Rather than writing the full name, I was thinking about `TOP_UNIT`, `ARCH` and `LIB`, along with having them explicitly defined as a symbol. `file...`, `options...` and `simulation_options...` are clearly identified because of `...`. However, we might want to use uppercases there too: `FILE...`, `OPTION...`, `SIM_OPTION...`. So:

- Lower or upper?
- All short or all long?

doc/requirements.txt
umarcor: Honestly, I don't know much about it. Lars (from VUnit) found that search was broken with `>=3.3.1` (https://github.com/VUnit/vunit/issues/713). We use the same template in VUnit and here, which is based on sphinx_rtd_theme. Moreover, sphinx_rtd_theme is broken with `docutils>=0.16` (see https://github.com/readthedocs/sphinx_rtd_theme/issues/1115). So, we are temporarily pinning those (see https://github.com/buildthedocs/docker/commit/f7bda0ebb71b51c2b7185cef6b6414efd20ed6cc and https://github.com/buildthedocs/docker/commit/1fd955532d6213bea45c291f44098ea04ac88265) until sphinx_rtd_theme is either fixed or updated (they are expecting to deprecate the current theme and use a new one based on bootstrap, see https://github.com/readthedocs/sphinx_rtd_theme/issues/1086). Then, we will either rebase or update https://github.com/buildthedocs/sphinx.theme for picking the fixes.

Xiretza: I see - incidentally, that comment would make a good commit message :)

umarcor: Touché :wink:

umarcor:
I added a commit for documenting command `elab-order`: https://github.com/ghdl/ghdl/commit/79fc4b05891c91db4cc33f82373b8a0792c41e02. I added it to section "Design rebuilding commands". Furthermore, I added an "attention" admonition explaining that it cannot handle multiple library names yet. @tgingold, is this correct?

umarcor:
> I think '[library.]top_unit' can also be used for sythesis and `-c/-r`.

For `--synth <[options...] file... -e [top_unit [arch]]>` it is said that:

> "*Analyses and elaborates for synthesis the files present on the command line only.*"

Therefore, the library is always `work`, isn't it? If the user provides `--work=something` because files do internally use all an specific lib name (instead of `work`), then `[library.]` would be redundant. So, even if the parser can accept it, I don't see the usefulness of documenting it.

With regard to `-c /-r`, I think that same applies. This is said in the docs:

> "*All the units of the files are put into the work library. But, the work library is neither read from disk nor saved. Therefore, you must give all the files of the work library your design needs.*"

---

> I also think that elab-order can handle multiple library. If not, that would be an issue.

I will test this.

tgingold:
Ok for -c/-r (It might be worth to allow --work= between files).

But for synth, it is possible to synthesis a pre-analyzed unit:
 ghdl --synth my_top
So:
 ghdl --synth my_lib.my_top
should be allowed.

umarcor:
> But for synth, it is possible to synthesis a pre-analyzed unit: ghdl --synth my_top
> So: ghdl --synth my_lib.my_top should be allowed.

That is already covered/documented. See the "Commands affected" list above: `--synth <[options...] [library.]top_unit [arch]>`. So, specifying the library is currently supported when file names are not provided, but not otherwise.

BTW, I checked that elab-order works: `--elab-order <[options...] [libray.]top_unit [arch]>`. But I did not try a design which relies on multiple libs.

> Ok for -c/-r (It might be worth to allow --work= between files).

If this was allowed, I would suggest it to be supported in all the places where `file...` is supported/used. So:

- `-a <[options...] file... [[--work=NAME file...]...]>`
- `-s <[options...] file... [[--work=NAME file...]...]>`
- `-c <[options...] file... [[--work=NAME file...]...] -<e|r> [library.]top_unit [arch]>`
- `-i <[options...] file... [[--work=NAME  file...]...]>`
- `--synth <[options...] file... [[--work=NAME  file...]...] -e [[library.]top_unit [arch]]>`

That would allow to synthesize designs with multiple libraries by specifying a single long synth command. Currently, users need to analyse first. I do find that same feature useful for `-a` and `-i`. I'm not sure about `-s`, but I guess it wouldn't harm.

PR #1738, opened 2021-04-23
Move ghwlib and ghwdump to subdir 'ghw', add function ghw_diff to testenv

### LINKED ISSUE #1323: GHW dumps not well tested
Currently there are no tests that check that the generated ghw wave dumps are good.

I see two options to fix this:
1) Create some golden ghw dumps and confirm that the generated ghw dump matches.
2) Have some logic in the test that checks the content of the ghw file and points out exactly where the problem is.

Option (2) would clearly be more helpful for debugging, but option (1) would be much easier to implement.  I plan on starting off with (1) and only implementing (2) if it seems worth it.

umarcor:
Closes #1323
Closes #1328

In this PR, C sources for reading/dumping GHW waves are moved from `/src/grt/` to `/ghw`. A README.md is added to the subdir, for explaining the purpose of those sources.

Some deps in the main Makefile are moved for slightly reducing the complexity.

The test by @benreynwar from #1328 is added to this PR, so #1323 is solved. A function named `ghw_diff` is added to testenv, so that other golden dumps can be added to other tests.

There are several other issues with TODOs about checking the waves. However, the golden references take +60k lines. I'm not sure about adding all of it to the codebase. See https://github.com/umarcor/ghdl/compare/ghwdump...umarcor:ghwdiff.


tgingold:
Please, move the ghw sources to `src/ghw`, as they are sources.  Unless you have a strong reason not to.

umarcor:
My main thought was that these sources are independent from the source of GHDL. Similarly to pyGHDL, the sources are tightly coupled to GHDL's codebase, because GHW is specified in there; but they are not required for building GHDL. Therefore, someone might want to get the sources of GHDL alone, or the sources of ghw alone.

Moreover, @Xiretza was looking into writing an equivalent to `ghwlib` in Rust, and I'd like to use it from Python. I think that these contributions might better fit in a subdir for "GHW helpers" (or in pyGHDL), apart from GHDL's sources.

PR #1744, opened 2021-04-26
Makefile.in: pass ghwdump location to testsuite

Xiretza:
**Description**
#1738 broke PKGBUILDs because tests are performed before GHDL is installed, so the `ghwdump` binary in the build directory must be used, however the `test` make targets don't set the `$GHWDUMP` variable.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -291,7 +291,7 @@ install.gcc: install.grt
 uninstall.gcc: uninstall.grt
 
 test.gcc:
-	cd $(srcdir)/testsuite; GHDL=$(GHDL_GCC_BIN) ./testsuite.sh
+	cd $(srcdir)/testsuite; GHDL=$(GHDL_GCC_BIN) GHWDUMP=$(CURDIR)/ghwdump$(EXEEXT) ./testsuite.sh
 
 #################### For gcc backend - development only (local build) ####
 
--- a/testsuite/testsuite.sh
+++ b/testsuite/testsuite.sh
@@ -19,6 +19,11 @@ enable_color() {
 disable_color() { unset ENABLECOLOR ANSI_RED ANSI_GREEN ANSI_YELLOW ANSI_BLUE ANSI_MAGENTA ANSI_CYAN ANSI_DARKCYAN ANSI_NOCOLOR; }
 enable_color
 
+die() {
+  printf "${ANSI_RED}%s${ANSI_NOCOLOR}\n" "$@" >&2
+  exit 1
+}
+
 print_start() {
   COL="$ANSI_YELLOW"
   if [ "x$2" != "x" ]; then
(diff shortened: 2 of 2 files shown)


umarcor:
Note that I added GHWDUMP a few days ago, as a workaround for running the tests in CI: https://github.com/ghdl/ghdl/commit/d79fa58f7efc06a048adf9e2172792e64b8abb20. On macOS, the location is added to the PATH instead: https://github.com/ghdl/ghdl/blame/287e808d5dc65165ceda5978976121938336f7c2/scripts/ci-run.sh#L426.

So, instead of adding the definition of GHWDUMP to the makefile, we could add a check to `testsuite/testsuite.sh` or `testsuite/testenv.sh` for retrieving the location relative to the location of GHDL. In a regular build, both binaries are located in the same place. We can do it right after https://github.com/ghdl/ghdl/blob/master/testsuite/testenv.sh#L33-L36.

umarcor:
FTR, I cherry-picked the first commit (black) and pushed it to master.

umarcor:
BTW, in the PKGBUILD recipes for MSYS2, I'm having issues for executing `make test` inside the `check()`. See https://github.com/ghdl/ghdl/pull/1573/files and https://github.com/ghdl/ghdl/actions/runs/786038214. The shared library is not found, so python tests fails. Did you find this problem on Arch?

Xiretza:
`testsuite.sh` now tries to infer the `ghwdump` path automatically from `$GHDL`. I'm not 100% sure it will work on windows due to `EXEEXT`, but iirc `.exe`/`.com` can be left off without a problem.

umarcor:
> I'm not 100% sure it will work on windows due to `EXEEXT`, but iirc `.exe`/`.com` can be left off without a problem.

It will. On Windows, we test it on MSYS2, where the extension is optional.

You left the definition of GHWDUMP in the Makefile for GCC backend. Was that on purpose or an oversight?

Xiretza:
> The shared library is not found, so python tests fails. Did you find this problem on Arch?

That particular log looks like it can't find the VHDL libraries, which isn't a problem with the mcode and llvm AUR packages. As you might know pyGHDL has [a fair bit of logic](https://github.com/ghdl/ghdl/blob/master/pyGHDL/libghdl/__init__.py#L73) to try to find libghdl, I wouldn't be surprised if MSYS needs some extra consideration there.

Xiretza:
> You left the definition of GHWDUMP in the Makefile for GCC backend. Was that on purpose or an oversight?

With the GCC backend, the `ghdl` binary isn't necessarily built in the same directory as `ghwdump` (due to all the `copy-sources`) stuff, so the autodetection wouldn't work.

umarcor:
The CI issue is unrelated. I merged #1747, then rebased this and I'm testing in https://github.com/umarcor/ghdl/actions/runs/789842744. When green, I'll merge.

umarcor:
Thanks!

PR #1746, opened 2021-04-26
Migrate deprecated DebugLoc::get to DILocation::get

vdkeybus:
GHDL seems to be using a deprecated class DebugLoc. Switched to the suggested alternative. Now compiles fine for LLVM 12.

### DIFF
--- a/src/ortho/llvm6/llvm-cbindings.cpp
+++ b/src/ortho/llvm6/llvm-cbindings.cpp
@@ -1254,7 +1254,7 @@ new_var_decl(ODnode *Res, OIdent Ident, OStorage Storage, OTnode Atype)
          DebugCurrentLine, Atype->Dbg, true);
       DBuilder->insertDeclare
         (unwrap(Decl), D, DBuilder->createExpression(),
-         DebugLoc::get(DebugCurrentLine, 0, DebugCurrentScope),
+         DILocation::get(DebugCurrentScope->getContext(), DebugCurrentLine, 0, DebugCurrentScope),
          unwrap(LLVMGetInsertBlock(DeclBuilder)));
     }
 #endif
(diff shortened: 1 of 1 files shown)


umarcor:
@vdkeybus thanks for providing this enhancement/fix!

I'm curious about how did you test this. According to https://github.com/ghdl/ghdl/blob/master/configure#L259-L284, configuring with LLVM 12 should fail. Did you edit the configure script too?

tgingold:
I am merging this PR as it passes the checks, but indeed I suppose `configure` must also be updated.

vdkeybus:
You're right. I tested in the original ghdl master branch. Then I forked a
new one and applied the patches there in order to commit them. The train of
events leading to the missing necessary configure patch to go unnoticed is
like this:

1/ I did not re-run configure in the original master branch, so I was
building with llvm 12, but the check in configure was simply never executed;
2/ I forgot to specify --with-lvm-config in my local branch, so I was
probably not building with llvm in that one (did build and test 1/ though);
3/ The pipeline does not have llvm 12, so that shows that at least the new
code works with older llvm versions. Obviously the missing llvm 12 check
was not detected.

Solution: I updated configure and submitted a PR. We may need to add an
llvm 12 build eventually.

(This is only the 2nd time I'm submitting a PR, so please forgive my
fooling around here.)



Op di 27 apr. 2021 om 02:16 schreef Unai Martinez-Corral <
***@***.***>:

umarcor:
> Solution: I updated configure and submitted a PR. We may need to add an
llvm 12 build eventually.

We don't test these bleeding-edge compiler versions in the CI of this repo. However, we do further testing in ghdl/docker, and moreover there are users testing it on Fedora and Arch. Therefore, we don't need to add an LLVM 12 build explicitly here.

> (This is only the 2nd time I'm submitting a PR, so please forgive my fooling around here.)

It's absolutely ok. We all need to learn and this project has several non-intuitive details. Please, don't worry about it.

PR #1747, opened 2021-04-27
Add missing llvm version check in configure

vdkeybus:
I tested the earlier DILocation patch in original ghdl master. In order to submit it, I forked ghdl and applied the patch there. However I forgot to specify --with-lvm-config in configure (which was still effective in the original ghdl master). I now added the necessary check to pass.

### DIFF
--- a/configure
+++ b/configure
@@ -276,6 +276,7 @@ if test $backend = llvm; then
        check_version 10.0 $llvm_version ||
        check_version 11.0 $llvm_version ||
        check_version 11.1 $llvm_version ||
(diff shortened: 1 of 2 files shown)


umarcor:
Nice! Please add 12.0 to https://github.com/ghdl/ghdl/blob/master/doc/development/building/LLVM.rst as well. See the previous PR for LLVM 11 as a reference: https://github.com/ghdl/ghdl/pull/1651.
BTW, you can rebase if you want.

PR #1749, opened 2021-04-27
Increase stack size for msys2-mcode binary (fix #1742)

### LINKED ISSUE #1742: GHDL-mcode: exception raised: stack overflow
**Description**
I was just trying to run a simple UVVM testbench with GHDL for one of my IPcores developped for Vivado. I am using the UVVM VCCs for axilite and axistream. Due to Xilinx not providing VHDL behavioral models for latest Xilinx Fifo Generator cores anymore I had to use structural models for the FIFOs of my core and had to compile the Xilinx Unisim library.

As a first test I made my testbench program some axilite registers (which works fine) and afterwards issue an axistream transaction. When the last transaction is simulated GHDL crashes with the message
`exception raised: stack overflow`

This happens only with the Windows mcode version of GHDL 2.0.0-dev (1.0.0.r163.gd6a8f93b). **With the 1.0.0 Linux release everything seems to work fine and the simulation finishes**.

**Expected behaviour**
Normal simulation execution without crash.

**How to reproduce?**
Next week I will try to provide a minimal working example to reproduce, I still have to remove some unsharable code and strip the example down as much as possible.
**Context**

- OS: Windows10 1909 64bit
- Origin: Released binaries: nightly build 2.0.0-dev (1.0.0.r163.gd6a8f93b).

**Additional context**
I would appreciate if someone could give me instructions on how to get more debug information from ghdl instead of just the stack overflow exception.

bahm-lgs:
This fixes the stack overflow error when using UVVM axistream_vvc for msys2-mcode.

### DIFF
--- a/scripts/msys2-mcode/PKGBUILD
+++ b/scripts/msys2-mcode/PKGBUILD
@@ -18,7 +18,7 @@ build() {
   cd "${srcdir}/builddir"
   ../../../../configure \
     --prefix=${MINGW_PREFIX} \
-    LDFLAGS=-static \
+    LDFLAGS="-static -Wl,--stack=8388608" \
     --enable-libghdl \
     --enable-synth
   make GNATMAKE="gnatmake -j$(nproc)"


umarcor:
On the one hand, I'm really excited about this fix. On the other hand, I have several questions:

- Why 8MB and not 16MB or 32MB? What does it depend on? Can we explain/document it somewhere?
- Why mcode only? You said you could not test it with LLVM/GCC because of issues with IT and the antivirus. However, would it apply to msys2-llvm too?

None of this questions is blocking. I'm good with merging this as is. However, I' then have to apply this patch upstream: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-ghdl/PKGBUILD. A better understanding would be desirable.

bahm-lgs:
> On the one hand, I'm really excited about this fix. On the other hand, I have several questions:
> 
> * Why 8MB and not 16MB or 32MB? What does it depend on? Can we explain/document it somewhere?

Because on Linux the default is 8MB which worked and I have tested 4MB to not work, I don't know what's actually required.

> * Why mcode only? You said you could not test it with LLVM/GCC because of issues with IT and the antivirus. However, would it apply to msys2-llvm too?

I don't know. I have not used msys2-llvm so far, because I'm still blocked from using mingw on my main office PC. For building GHDL binaries I could simply use a VM, but for daily simulation usage this does not work well. And if I understood it correctly, LLVM/GCC requires mingw also for building every simulation...

> None of this questions is blocking. I'm good with merging this as is. However, I then have to apply this patch upstream: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-ghdl/PKGBUILD. A better understanding would be desirable.

I have also no deep understanding on this, this fix simply "works for me". Before merging please have someone review this who knows better.

tgingold:
Thanks!

umarcor:
The MSYS2  package upstream was updated and this fix was applied: https://github.com/msys2/MINGW-packages/pull/8621

PR #1755, opened 2021-05-09
Respect more build flags from environment

Xiretza:
**Description**
Currently, the build system isn't very good at using `CFLAGS`/`LDFLAGS`/etc from the environment, which causes all sorts of subtle problems for systems relying on them. I fixed what I could find, and GHDL can now be easily built with AddressSanitizer (`-fsanitize=address` in `{C,LD,ADA_}FLAGS`) and Arch debug packages work correctly as well.


tgingold:
Thank you for the cleanup.  Was needed !

Issue #1766, opened 2021-05-16
Powershell vendor script for OSVVM is failing with "Missing an argument for parameter 'VHDL2008'"


umarcor:
It seems that https://github.com/ghdl/ghdl/commit/47e76528 (#1730) introduced a regression.

- Extended tests run before: https://github.com/ghdl/extended-tests/runs/2392127303
- Extended tests run after: https://github.com/ghdl/extended-tests/runs/2401859382

```
compile-osvvm.ps1: D:\a\extended-tests\extended-tests\verification\OSVVM\run.ps1:2
Line |
   2 |  & $env:GHDL_PREFIX\vendors\compile-osvvm.ps1 `
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Missing an argument for parameter 'VHDL2008'. Specify a parameter of type 'System.Boolean' and try
     | again.
```

/cc @bahm-lgs @Paebbels

bahm-lgs:
It seems like compile-osvvm.ps1 uses the function Get-VHDLVariables with parameters of type "switch" and and compile-xilinx-vivado.ps1 with params of type "boolean". That's why 47e7652 fixed it for the xilinx scipt but broke it for the OSVVM script.

This could be fixed like that:
bbb2a1220f4febe776b6a093c53fa8a2ee6535f1

Paebbels:
Thanks @bahm-lgs for narrowing it down. Yesterday, I also had a quick look and was distracted by the error message and line number. I also traced it to `Get-VHDLVariables` CmdLet using `-VHDL2008`, but unfortunately run out of time to see the differences in switch vs. boolean.

bahm-lgs:
@umarcor could you rerun the extended regression tests with the code from #1767?

umarcor:
> @umarcor could you rerun the extended regression tests with the code from #1767?

@bahm-lgs I'm trying to do so: https://github.com/umarcor/ghdl-extended-tests/actions/runs/853983847. Unfortunately, GitHub Actions is having issues today: https://www.githubstatus.com/. I will let you know as soon as I can see it executed.

With regard to the specific fix, just for my own understanding, why `bool` and not `switch`? Is there any difference between them? Why do both exist?

> Or simply omit the parameter like in the UVVM script, since VHDL-2008 is the default anyway.

[...]

PR #1767, opened 2021-05-17
OSVVM-compilescript: Explicitly set boolean parameter VHDL-2008 to true

bahm-lgs:
Fix ghdl/ghdl#1766

### DIFF
--- a/scripts/vendors/compile-osvvm.ps1
+++ b/scripts/vendors/compile-osvvm.ps1
@@ -112,7 +112,7 @@ $GHDLBinary =           Get-GHDLBinary $GHDL
 New-DestinationDirectory $DestinationDirectory
 cd $DestinationDirectory
 
-$VHDLVersion,$VHDLStandard,$VHDLFlavor = Get-VHDLVariables -VHDL2008
+$VHDLVersion,$VHDLStandard,$VHDLFlavor = Get-VHDLVariables -VHDL2008:$true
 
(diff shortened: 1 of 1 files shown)


tgingold:
@umarcor, @Paebbels: Is it ok ?

Paebbels:
Thanks for the reminder !

----------
There is a problem with the type of the parameter `boolean` vs. `switch`.

Handing over the parameter via `-VHDL2008:$true` is a workaround but not fixing the origin of the problem. I'll create a correct fix and PR, so `:$true` is not needed.

----------------
@umarcor Can I update this PR or do I need a new one?

umarcor:
> Can I update this PR or do I need a new one?

You can try adding the fork as a remote and pushing on top of the feature branch. I don't see any note about "maintainers/collaborators can write to this PR". I think that's only shown to the author of the PR.

Anyway, it's ok to create another PR and closed this one as "superseeded by".

Paebbels:
@bahm-lgs sorry that this got so long delayed. I summarized all proposed changes and also what was the old, new and intended code. The fix is in #1878.

bahm-lgs:
> @bahm-lgs sorry that this got so long delayed. I summarized all proposed changes and also what was the old, new and intended code. The fix is in #1878.

Works for me, tested with xilinx-vivado, uvvm and osvvm scripts. Thanks!

Issue #1599, opened 2021-01-15
Dropping support for GNAT compiler


Paebbels:
Currently GHDL is compiled for Windows with MinGW and we can proof it works fine with mcode and llvm backends. Compiling GHDL with GNAT (Ada compiler from Ada Core) is hard to maintain. The current scripts are outdated and don't support latest PowerShell versions. See #732. These scripts are also not used in CI runs.

**Should we drop PowerShell based compile scripts?**
* Main compile script `compile.ps1`
* GHDL compile script `compile-ghdl.ps1`
* Library compile script `compile-libraries.ps1`
* Shared Module `shared.psm1`

We might want to keep the installer script. This would allow us to ship a `*.ps1` file with an internal ZIP container that self-extracts and copies contents to right places.

This functionality exists and is easier to maintain then a MSI file.

Alternatively we could consider shipping GHDL via Chocolaty / NuGet.

----------------
/cc @umarcor

umarcor:
For context, we are talking about the scripts currently located in `scripts/windows`. Until recently, those were located in `dist/windows`. For some years, those were the only scripts available for building GHDL on Windows (from powershell). Patrick did the initial research for allowing GHDL to be built "easily" on Windows. He also met some of the cliffs in his way (GNAT, GCC backend, Microsoft compiler...).

However, despite being written in powershell, those scripts used MSYS2 under the hood. Some months ago, PKGBUILD recipes were written (now located in `scripts/msys2-*`). Those recipes are equivalent to powershell scripts, but are the canonical build procedure on MSYS2. Using `makepkg-mingw` instead of custom scripts does provide several advantages, such as getting dependencies automatically, building for MINGW32 and/or MINGW64 at the same time, generating multiple packages, etc. Those recipes made it easier to upstream the packages to MSYS2's repos. Since a couple of months ago, GHDL is installable through pacman, and it's updated regurlarly. The similarity between the custom recipes here and the upstreamed one makes it easier to handle updates. Those recipes do also make setup-ghdl-ci easy to maintain, because dependencies are metadata of the packages.

At the same time, we have recently been testing how to use MSYS2 packages outside of MSYS2. Patrick has successfully used either mcode or LLVM builds in PyCharm using CPython and pyGHDL. The procedure for getting a zipfile from MSYS2 packages is not automated yet; however, the zipfiles generated through powershell/appveyor scripts do have the same issue.

[...]

Issue #1560, opened 2020-12-30
Install dependencies (DLLs) with MinGW64 builds


Paebbels:
# Problem
When loading the `libghdl...dll` into Python via `pyGHDL.libghdl`, some dependencies are missing in the GHDL installation directory, if GHDL is installed outside of MinGW64 system.

These dependencies need to be added to the `/lib` directory of GHDL while installation.

# Solution
I used [Dependencies](https://github.com/lucasg/Dependencies) to discover needed libraries.

### For `ghdl.exe`
* `zlib1.dll`

### For `libghdl-1_0_dev.dll`
* `libgnat-10.dll`
* `libgcc_s_seh-1.dll`
* `libwinpthread-1.dll`

![image](https://user-images.githubusercontent.com/956109/103338221-fd8a5180-4a7d-11eb-9f02-a793529514a5.png)

### For `ghdl.exe`
*None*

### For `ghdl-llvm.exe`
* `libgcc_s_seh-1.dll`
* `libwinpthread-1.dll`
* `libstdc++-6.dll`
* `libLLVM.dll`
* `zlib1.dll`
* `libffi-7.dll`

### For `libghdl-1_0_dev.dll`
* `libgnat-10.dll`
* `libgcc_s_seh-1.dll`
* `libwinpthread-1.dll`

### For elaboration (linking)
When GHDL elaborates the design, the generated `*.o` are linked via `gcc.exe` using `ld.exe`.  
A `lto` plugin is needed for the linker and a set of `*.a` and `*.o` files.

* for `x86_64-w64-mingw32/lib/...`
  * `crt2.o`
  * `default-manifest.o`
  * `libadvapi32.a`
  * `libdbghelp.a`
  * `libghdl.a`
  * `libkernel32.a`
  * `libmingw32.a`
  * `libmingwex.a`
  * `libmoldname.a`
  * `libmsvcrt.a`
  * `libpthread.a`
  * `libshell32.a`
  * `libuser32.a`
* for `lib/gcc/...`
  * `libz.a`
* for `lib/gcc/x86_64-w64-mingw32/10.2.0/...`
  * `crtbegin.o`
  * `crtend.o`
  * `libgcc.a`
  * `libgcc_eh.a`
  * `liblto_plugin-0.dll`

This is the full directory tree of GHDL after adding files to the right places:
```
├──bin
│  ├──gcc.exe
│  ├──ghdl.exe
│  ├──ghdl1-llvm.exe
│  ├──ld.exe
[...]

PR #1773, opened 2021-05-24
ci: generate Windows installers

Adrijaned:
**Description** 
Added a CI job steps to generate windows NSIS installers from the built windows binaries. This serves mainly as a way to decrease entry barrier for learning to write VHDL on windows, but should be generally useful for everyone running windows.

:rotating_light: This has *not* been thoroughly tested, as I don't have any windows machine currently on hand, and don't have enough free disk space for a 64-bit windows virtual machine. I did however at least test that the 32-bit installer installs and uninstalls correctly, and that the `ghdl` binary can be launched from `cmd` after installation, and there *should* not be anything too different in the 64 bit version to cause breakage.

**Further comments**
This is a fairly simple and primitive installer, currently serving as a proof of concept for me for another ghdl-related project, but should be mergeable nevertheless. If that other thing comes through and this gets merged, I will probably create another PR trying to simplify the `.nsi` scripts to reduce code duplication, and probably to include option to somehow also install `gtkWave` alongside `ghdl` as optional component.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -267,7 +267,7 @@ jobs:
     - name: '📤 Upload artifact: installer'
       uses: actions/upload-artifact@v2
       with:
-        path: scripts/msys2-${{ matrix.pkg }}/pkg/mingw-w64-${{ matrix.arch }}-ghdl-${{ matrix.pkg }}/installer-${{ matrix.arch }}-${{ matrix.pkg }}.exe
+        path: scripts/msys2-${{ matrix.pkg }}/pkg/mingw-w64-${{ matrix.arch }}-ghdl-${{ matrix.pkg }}/ghdl-installer-${{ matrix.arch }}-${{ matrix.pkg }}.exe
 
 #
 # Windows Test
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -242,6 +242,15 @@ jobs:
         cd scripts/msys2-${{ matrix.pkg }}
         makepkg-mingw --noconfirm --noprogressbar -sCLf
 
+    - name: 'Prepare NSIS build'
+      run: |
+        cp scripts/installer-${{ matrix.installs }}-${{ matrix.pkg }}.nsi scripts/msys2-${{ matrix.pkg }}/pkg/mingw-w64-${{ matrix.arch }}-ghdl-${{ matrix.pkg }}
+
+    - name: 'Create NSIS installer'
+      uses: joncloud/makensis-action@v3.3
+      with:
+        script-file: scripts/msys2-${{ matrix.pkg }}/pkg/mingw-w64-${{ matrix.arch }}-ghdl-${{ matrix.pkg }}/installer-${{ matrix.installs }}-${{ matrix.pkg }}.nsi
+
     - name: '📤 Upload artifact: builddir'
       uses: actions/upload-artifact@v2
       with:
@@ -254,6 +263,11 @@ jobs:
       uses: actions/upload-artifact@v2
       with:
         path: scripts/msys2-${{ matrix.pkg }}/mingw-*ghdl*.pkg.tar.zst
+
+    - name: '📤 Upload artifact: installer'
+      uses: actions/upload-artifact@v2
+      with:
+        path: scripts/msys2-${{ matrix.pkg }}/pkg/mingw-w64-${{ matrix.arch }}-ghdl-${{ matrix.pkg }}/installer-${{ matrix.arch }}-${{ matrix.pkg }}.exe
 
 #
 # Windows Test
--- a/scripts/installer-MINGW32-mcode.nsi
+++ b/scripts/installer-MINGW32-mcode.nsi
@@ -1,0 +1,254 @@
+!define APPNAME "GHDL"
+!define COMPANYNAME "GHDL"
+!define DESCRIPTION " VHDL 2008/93/87 simulator"
+
+;kind dependent section
+!define ARCH "mingw32"
+!define BACKEND "mcode"
+!define ARCH2 "i686"
+InstallDir "$PROGRAMFILES\${APPNAME}"
+;end kind dependent section
+
+!define VERSIONMAJOR 1
+!define VERSIONMINOR 0
+!define VERSIONBUILD 0
+
+OutFile "ghdl-installer-${ARCH2}-${BACKEND}.exe"
+
+Name "${APPNAME}"
+
+!include LogicLib.nsh
+
+Page directory
+Page instfiles
+
+!macro VerifyUserIsAdmin
+UserInfo::GetAccountType
+Pop $0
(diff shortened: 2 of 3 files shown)


.github/workflows/Test.yml
```diff
@@ -242,6 +242,15 @@ jobs:
         cd scripts/msys2-${{ matrix.pkg }}
         makepkg-mingw --noconfirm --noprogressbar -sCLf
 
+    - name: 'Prepare NSIS build'
```
umarcor, line 245: Please, create the NSIS installer in a separated job, *after* `win-test`. See https://github.com/ghdl/ghdl/actions/runs/868960417.

We want to first build GHDL, then test it, and generate the installer only if the tests are successful.

umarcor (commented):
Overall, the idea is nice, since it allows users not to install MSYS2 explicitly. However, I'm afraid it is an incomplete approach. It seems that you are creating an installer from the content of the `pkg` directory created during the build on MINGW32/MINGW64. The content of that directory is only guaranteed to work on an MSYS2 environment. The build with an mcode backend might work by chance, because it's built statically. However, the same approach will fail for the LLVM backend. Please, see #1560 and #1599 for further info about the challenge.

I tried to find somewhere in the `*.nsi` files that the required DLLs are copied, but I couln't find it. Am I missing something?

umarcor:
> and that the ghdl binary can be launched from cmd after installation

Launching the GHDL alone is not enough testing. You should run the testsuite, or some of the tests at least, for ensuring that no dependencies are missing.

>there should not be anything too different in the 64 bit version to cause breakage.

The 64 bit version is very different because mcode backend is not supported on Windows x86-64. Therefore, a completely different backend is used. mcode is a in-memory statically built-in backend. LLVM is an external compiled solution, which is not built statically.

> will probably create another PR trying to simplify the .nsi scripts to reduce code duplication, and probably to include option to somehow also install gtkWave alongside ghdl as optional component.

Note that GTKWave uses MSYS2 in CI and then generates standalone zipfiles by copying additional files from manually maintained lists. See https://github.com/gtkwave/gtkwave/tree/master/MSYS2, https://github.com/gtkwave/gtkwave/pull/20 and https://github.com/gtkwave/gtkwave/issues/18.

I did discuss with @Paebbels about using a similar solution for GHDL. So, copy the `standalone_pkg.sh` from GTKWave, and write `*dll.inc`, `*lib.inc` and `*exe.inc`, with the lists provided in https://github.com/ghdl/ghdl/issues/1560#issue-776330650. However, that is something I cannot afford to maintain myself. I can help with CI, management of artifacts, assets, releases, etc. but I cannot take care of keeping those `*.inc` files up to date. Nonetheless, if you or anyone else wants to take care of it, that'd be awesome. It would allow to use the `*.nsi` files proposed in this PR. Furthermore, it might not be necessary to maintain the list manually. See "Existing tools for automatic discovery of MSYS2 package dependencies (files)" (https://github.com/ghdl/ghdl/issues/1560#issuecomment-752750494).

With regard to combining multiple tools in one installer, see the comment about "something similar to Ninite" in https://github.com/ghdl/ghdl/issues/1599#issuecomment-761107708.

Adrijaned:
Oof, that's a lot more complicated than I had hoped for. Now, 64-bit with llvm looks way more complicated than 32-bit mcode, and I don't even have a need for it nor do I expect a lot of people who might just want to try learning VHDL without too much obstacles on the way to have.

If we decreased the scope of this PR to just create installer for the mcode x86 binary, how should I go around testing that the install works on a fairly bare system? So far I've only succesfully manually run the installed binary on two very simple designs, both time utilizing only the `-a`, `-e`, `-r`, and `--wave` flags.

umarcor:
> Oof, that's a lot more complicated than I had hoped for.

Yeah, "*(...) software is hard. It's harder than anything else I've ever had to do.*" (https://en.wikiquote.org/wiki/Donald_Knuth)

> Now, 64-bit with llvm looks way more complicated than 32-bit mcode, and I don't even have a need for it nor do I expect a lot of people who might just want to try learning VHDL without too much obstacles on the way to have.

See #1785. That's a fair point. Still, mcode has dependencies which are not included in the tarball. See https://github.com/ghdl/ghdl/blob/master/scripts/msys2-mcode/PKGBUILD#L10-L14 and #1560. So, even if you want to install the 32 bit mcode version only, you need to at least include zlib, libgnat, gcc_s_seh and libwinpthread. Moreover, we cannot guarantee that those DLLs will not conflict with others in the system. Alternatively, you could enhance the build plumbing, for generating a completely static binary. Last, you could try using AdaCore's GNAT, instead of the MSYS2 recipes.

> If we decreased the scope of this PR to just create installer for the mcode x86 binary, how should I go around testing that the install works on a fairly bare system? So far I've only succesfully manually run the installed binary on two very simple designs, both time utilizing only the `-a`, `-e`, `-r`, and `--wave` flags.

Create an additional `windows-latest` job in CI where you do nothing other than download the artifact, checkout the repo and run the testsuite. You run the testsuite with `GHDL=path/to/ghdl ./testsuite/testsuite.sh`.

That is currently done for MSYS2. There is a `win-build` job matrix (https://github.com/ghdl/ghdl/blob/master/.github/workflows/Test.yml#L200) and then a `win-test` job matrix (https://github.com/ghdl/ghdl/blob/master/.github/workflows/Test.yml#L262). In the test jobs, we install the testsuite dependencies only: diffutils, gcc, python-pip, python-setuptools and `testsuite/requirements.txt` (https://github.com/ghdl/ghdl/blob/master/.github/workflows/Test.yml#L287-L314). You would need to install those using some alternative solution, since the purpose is precisely not to use MSYS2. Alternatively, you can skip some of the testsuites (e.g. pyunit) for avoiding the dependency on Python.

Paebbels:
Some long time ago, GHDL had NSIS support to create MSI installer files. The problem with NSIS is - and I think it's not solved yet or ever - it doesn't create correct installations. Mainly adding GHDL to PATH was completely broken. And if I remember correctly, it only GHDL wasn't added correctly, it fucked up the PATH variable for the whole system ...

Is this problem now solved with latest NSIS?

Paebbels:
See this very very old ticket started in 2014: https://sourceforge.net/p/ghdl-updates/tickets/10/

Paebbels:
While NSIS 2 is a dead end (latest version 2016).
image  
Source: https://sourceforge.net/projects/nsis/files/NSIS%202/


There is a new NSIS 3 ...
https://nsis.sourceforge.io/Download

Looking at the list of bugs (quite long with 179 issues), the PATH bug is still listed:
* https://sourceforge.net/p/nsis/bugs/1247/

This might be also interesting to see:
* GitHub Actions and NSIS: https://github.com/NSIS-Dev/github-action-examples
* Old NSIS file: https://sourceforge.net/p/ghdl-updates/ghdl-updates.git/ci/master/tree/dist/mcode/windows/ghdl.nsi

PR #1784, opened 2021-06-10
OpenBSD port

ams:
**Description**
Quick port to OpenBSD (6.9, AMD64) to use the mcode backend.  Test suite passes.

There are some egregious fixes here to make the test suite work on OpenBSD, like removing `--strip-trailing-cr` to diff, fixing some C99 specific things (OpenBSD defaults to C90) and other misc. changes.

The changes specifically to get GHDL working on OpenBSD are: 76fb1cf9d3a6d59fab5127b47f22ab08f2eaa801, 236ef20f39a7f519079e7ad6316b00a747980602  and 51d78ee6ffdcf308f2ee74f3d6f285eae595f02c.

Log files:

[configure.txt](https://github.com/ghdl/ghdl/files/6630447/configure.txt)
[testsuite.txt](https://github.com/ghdl/ghdl/files/6630445/testsuite.txt)
[compile.txt](https://github.com/ghdl/ghdl/files/6630446/compile.txt)

### DIFF
--- a/configure
+++ b/configure
@@ -228,6 +228,7 @@ if test $backend = mcode; then
   case "$build" in
     i[3-6]86*) mcode64="" ;;
     x86_64*) mcode64="64" ;;
+    amd64*)  mcode64="64" ;;
     *)
       mcode64=""
       echo "WARNING: GHDL for mcode is supported only on x86"
@@ -376,6 +377,7 @@ if test $backend = mcode; then
     *darwin*) ortho_flags="Flags_Macosx${mcode64}" ;;
     *mingw32*) ortho_flags="Flags_Windows${mcode64}" ;;
     *linux*) ortho_flags="Flags_Linux${mcode64}" ;;
+    *openbsd*) ortho_flags="Flags_Openbsd${mcode64}" ;;
     *) echo "Unsupported $build build for mcode"; exit 1;;
   esac
   echo "Generate ortho_code-x86-flags.ads"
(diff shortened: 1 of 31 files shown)


tgingold:
Interesting.

I think I can merge the changes that are specific to OpenBSD.  Maybe I will reuse the macosx configuration for mcode rather than adding identical config files.

I am not sure about the purpose of the symlinks.  I suppose this was for locally running the testsuite.

The changes for diff will introduce failures for windows.  Maybe we should create a command in testenv.sh for diffing and not using strip trailing cr on OpenBSD.

ams:
> Interesting.

:-)

I was slightly surprised that nothing strange was required.

> I think I can merge the changes that are specific to OpenBSD.  Maybe I will reuse the macosx configuration for mcode rather than adding identical config files.

No opinion, I started out that way and it worked.

> I am not sure about the purpose of the symlinks.  I suppose this was for locally running the testsuite.

Correct.

> The changes for diff will introduce failures for windows.  Maybe we should create a command in testenv.sh for diffing and not using strip trailing cr on OpenBSD.

Want me to write something up for that? I can't test it though, but maybe the CI pipeline will catch it?

ams:
Pushed a fix that should DTRT when it comes to the diff command.

tgingold:
I have just merged your patches with some minor changes.  Tell me if I broke something on OpenBSD.

Thank you for the PR!

ams:
Just verified that the testsuite still passes, thanks!

tgingold:
Great!  Thanks.

PR #1795, opened 2021-06-18
pyGHDL: add types to nodes py

tgingold:
@Paebbels: here is my patch to add type annotations to nodes.py

I have tested it quickly on my computer, and I am using CI to check it more thoroughly.
I am pretty sure black will cry.

We need to settle on the type name convention.  We often strip `_` from Ada type names (like `Iir_Kind` -> `IirKind`), but not always (like `Location_Type`).  I think we should always strip the `_`.

There is still one mismatch: `Token_Type` is `Tok` in python, so that the literals are directly translated: `Tok_String` becomes `Tok.String`

I think this is my first self pull-request!

### DIFF
--- a/pyGHDL/libghdl/_decorator.py
+++ b/pyGHDL/libghdl/_decorator.py
@@ -31,7 +31,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ============================================================================
 #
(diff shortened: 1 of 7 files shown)


Paebbels:
I'm Ok with the implementation.

Issue #1793, opened 2021-06-18
Assert failure in vhdl-nodes when analysing some sources from NEORV32


umarcor:
While trying the new [DOM.py](https://github.com/ghdl/ghdl/blob/master/pyGHDL/cli/DOM.py) example with sources from [NEORV32: rtl/core](https://github.com/stnolting/neorv32/tree/master/rtl/core), I found some to be properly handled, other to produce legit exceptions, and a few to produce an assert failure in libghdl. Precisely:

- Successful
  - `rtl/core/neorv32_cpu_decompressor.vhd`
- `raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : vhdl-nodes.adb:861`
  - `rtl/core/neorv32_busswitch.vhd`
  - `rtl/core/neorv32_cpu_alu.vhd`
  - `rtl/core/neorv32_cpu_regfile.vhd`
  - `rtl/core/neorv32_cpu.vhd`
- Aggregate
  - `rtl/core/neorv32_application_image.vhd`: `Unknown expression kind 'Aggregate'(185) in expression '1305'`
  - `rtl/core/neorv32_bootloader_image.vhd`: `Unknown expression kind 'Aggregate'(185) in expression '1305'`
- Parenthesis
  - `rtl/core/neorv32_boot_rom.vhd`: `Unknown expression kind 'Parenthesis_Name'(259) in expression '1346'`
  - `rtl/core/neorv32_cfs.vhd`: `Unknown expression kind 'Parenthesis_Name'(259) in expression '1391'`
  - `rtl/core/neorv32_cpu_control.vhd`: `Unknown expression kind 'Parenthesis_Name'(259) in expression '1538'`
  - `rtl/core/neorv32_dmem.vhd`: `Unknown expression kind 'Parenthesis_Name'(259) in expression '1367'`
  - `rtl/core/neorv32_gpio.vhd`: `Unknown expression kind 'Parenthesis_Name'(259) in expression '1359'`
  - `rtl/core/neorv32_icache.vhd`: `Unknown expression kind 'Parenthesis_Name'(259) in expression '1413'`
  - `rtl/core/neorv32_imem.vhd`: `Unknown expression kind 'Parenthesis_Name'(259) in expression '1374'`
- String
  - `rtl/core/neorv32_cpu_cp_fpu.vhd`: `Unknown expression kind 'String_Literal8'(11) in expression '1383'`
[...]

PR #1797, opened 2021-06-18
Add unit tests for pyGHDL DOM

umarcor:
This PR adds a ~~job matrix~~ pytest script for testing DOM ~~on MSYS2~~. By using the feature introduced in #1794, all the `*.vhdl` sources in the repository are processed. It needs ~4-6~ 8-12 min, however, it's still faster than some other jobs in the same workflow.

~~My understading is that these jobs cannot fail. When the exceptions are handled, the exit code is 0. When assertion failure happen (#1793) it seems that the exit code is 0 too. As a result, they are not very useful as an automated test yet, but just for human inspection. When the missing features are completed, we might improve `DOM.py` for it to actually fail.~~


.github/workflows/Test.yml
```diff
@@ -313,13 +313,75 @@ jobs:
...
+    - name: '📥 Download artifact: package'
+      uses: actions/download-artifact@v2
+
+    - name: '🚧 Install package and Python dependencies'
+      run: |
+        pacman --noconfirm -U artifact/mingw-w64-${{ matrix.sys.arch }}-ghdl-${{ matrix.sys.pkg }}-*.zst
+        pip3 install .
+
+    - name: Test the DOM and pretty print all VHDL sources
+      if: matrix.suite == 'sanity pyunit vpi vhpi'
```
Xiretza, line 364: Unlike `win-test` (where this was presumably copied from), this job doesn't define `matrix.suite`, so this is never run.

umarcor: Interesting, you are correct, but the jobs did run indeed 😆 (see below). It's obviously a bug in GitHub's evaluation of the `if`.

I will fix that.

Xiretza: It's just the step that is skipped: https://github.com/ghdl/ghdl/pull/1797/checks?check_run_id=2857078893

umarcor: Correct. My refactoring mistake: https://github.com/umarcor/ghdl/runs/2854455105?check_suite_focus=true#step:8:18.

Running now: https://github.com/ghdl/ghdl/runs/2857276874?check_suite_focus=true#step:7:18

.github/workflows/Test.yml
```diff
@@ -313,13 +313,74 @@ jobs:
...
+
+    - name: '🚧 Install package and Python dependencies'
+      run: |
+        pacman --noconfirm -U artifact/mingw-w64-${{ matrix.arch }}-ghdl-${{ matrix.pkg }}-*.zst
+        pip3 install .
+
+    - name: Test the DOM and pretty print all VHDL sources
+      run: |
+        cat > test_pyGHDL.sh <<EOF
+        #!/usr/bin/env bash
```
Xiretza, line 366: This should `set -e` or equivalent to avoid swallowing the exit code.

umarcor: Yeah, but that is on purpose. I don't think it makes sense to enforce it if the DOM.py scripts handles exception as follows: https://github.com/ghdl/ghdl/blob/master/pyGHDL/cli/DOM.py#L53-L54. So, currently, it wouldn't be meaningful. We know it fails a lot because the feature is not complete. 

We expect multiple changes in the following days, since @Paebbels is working really hard on this feature. This job is for having a quick coarse feedback about the progress. Then, we will enhance this for properly reporting how many files were successful, which failed, etc.

In fact, `set -e` should be the default if the call is not wrapped (which is just for aesthetic purposes). This is the MSYS2 entrypoint: https://github.com/msys2/setup-msys2/blob/master/main.js#L169. However, see https://github.com/umarcor/ghdl/runs/2857691698?check_suite_focus=true. It didn't fail on the first error.

umarcor: Since several crashes were fixed yesterday and today, I updated the PR for using pytest, instead of a bash wrapper.

umarcor:
Currently, there seems to be some crash with mcode backend, which prevents the test from finishing. With LLVM, some unrelated tests are failing, which I don't understand why; however, the test can finish and results are reported.

Paebbels:
@umarcor is this ready for merge even with errors on your repo?

umarcor:
> @umarcor is this ready for merge even with errors on your repo?

I don't think so. I converted it to a draft now. 

I don't understand why adding a new file for pytest does break other tests in testsuite/pyunit.

umarcor:
@Paebbels see https://github.com/umarcor/ghdl/runs/2865041837?check_suite_focus=true#step:7:9. That is generated with a shell script: https://github.com/umarcor/ghdl/commit/6f18de382f6980968a94fab50d8a75fd1d6f55d1#diff-78765f54ba6e205a4b8e7b6daa8cfe64a9636173f9928b8afe138b0f57a392f8.

I'd like to provide a similar functionality using a Python script, unittest and pytest's parametrize and xfail marks. With the bash script, we either pass (ignoring failures) or will fail, blocking CI. Pytests allows us to define some tests as expected to fail, which won't block CI.

Unfortunately, it seems that loading DOM once for all the files/tests is not working.

umarcor:
This is working as desired now:

```
========== 17 passed, 1140 xfailed, 1607 xpassed in 465.68s (0:07:45) ==========
```

Instead of importing and calling DOM, it is called as an external script/tool. That "fixes" the issue with other tests being broken. 

I added options `-v` and `-s` to the pytest call because the captured output is not shown otherwise. When the remaining features are added and we remove the xfail marks, we might remove those verbosity options too.

Since this is now a pyunit test, it's executed on all jobs, not only on Windows.

PR #1798, opened 2021-06-19
Python-C/Ada Bindings - Updated decorator

Paebbels:
**Description**
* Added new types (c_int64, c_double) to the decorator.
* Fixed typo `Iir_Kind` &rarr; `IirKind` in `pnodes.py`
* If a return type is an enum derived from `IntEnum`, it is automatically converted from `c_int32/int` to the enum.  
   As to functions are used in the decorator, there is no `if...then...else` runtime overhead.
* Add decorator usage everywhere in auto generated API.
* Changed all TypeVars to use more specific CDLL types.

----------------
**TODOs:**
* [x] Review changes  
  &rArr; **new PR**
* [ ] Split type translation function for directions PY->C and C->PY into 2 functions  
  &rArr; **new PR**
* [x] Review / discuss if TypeVars should be bound to Python types (done by @Paebbels) or to CDLL types (done by @tgingold).  
  &rArr; **done**
* [ ] `Has_` functions had no type for parameter `kind`.  
  I used `IirKind`. Is this correct?

### DIFF
--- a/pyGHDL/dom/Expression.py
+++ b/pyGHDL/dom/Expression.py
@@ -30,6 +30,14 @@
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ============================================================================
+from typing import List
+
+from pyGHDL.dom.Aggregates import OthersAggregateElement, SimpleAggregateElement, RangedAggregateElement, IndexedAggregateElement, NamedAggregateElement
+from pyGHDL.dom.Symbol import EnumerationLiteralSymbol
+from pyGHDL.libghdl import utils
+
+from pyGHDL.dom.Common import DOMException
+from pyGHDL.dom._Utils import GetIirKindOfNode
 from pyGHDL.libghdl.vhdl import nodes
 from pydecor import export
 
@@ -66,7 +74,8 @@ from pyVHDLModel.VHDLModel import (
     ShiftLeftArithmeticExpression as VHDLModel_ShiftLeftArithmeticExpression,
     RotateRightExpression as VHDLModel_RotateRightExpression,
     RotateLeftExpression as VHDLModel_RotateLeftExpression,
-    Expression,
+    Aggregate as VHDLModel_Aggregate,
+    Expression, AggregateElement,
 )
 
 __all__ = []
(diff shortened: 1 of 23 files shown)


tgingold:
I am not sure about binding typevars to python vs ctypes.  How do you know the real type ?  For example, what is the type of 'float' ? Is it a float32 or a float64 ?  At least with ctypes that's clear.

Paebbels:
Yesterday night I had the same thoughts for `c_int**` vs. `c_uint**` and `c_*int32` vs. `c_*int64` but I didn't want to decide on a rush.

Yes I think CDLL types are better, because we need the size and signedness information for the interface. Otherwise we would need to assume, C/Ada uses e.g. only `c_int64` in integer interfaces, but never a unsigned variant and a 32-bit variant.

While this could be agreed on, I think main problem could be libghdl build for MinGW32 vs. MinGW64. How should Python know what library is used on the other side.

tgingold:
I am not sure to understand what you mean for MinGW32 vs MinGW64.  That's two different python binaries, using two different libghdl dlls.

Paebbels:
Please have a look at TODO 4. Other stuff is done so far and ready for merge.

pyGHDL/dom/Expression.py
```diff
@@ -350,3 +381,46 @@ def __init__(self, left: Expression, right: Expression):
...
+                r = GetExpressionFromNode(nodes.Get_Choice_Range(item))
+                value = GetExpressionFromNode(nodes.Get_Associated_Expr(item))
+                choices.append(RangedAggregateElement(r, value))
+            elif kind == nodes.Iir_Kind.Choice_By_Name:
+                name = EnumerationLiteralSymbol(nodes.Get_Choice_Name(item))
+                value = GetExpressionFromNode(nodes.Get_Associated_Expr(item))
+                choices.append(NamedAggregateElement(name, value))
+            elif kind == nodes.Iir_Kind.Choice_By_Others:
+                expression = None
+                choices.append(OthersAggregateElement(expression))
```
tgingold, line 418: Choice_By_Others have a value.
I think this can be slightly refactorized: `value` can be extracted before the `if`.

Paebbels: Oh, you're right. It will be fixed with the next set of enhancements.

PR #1801, opened 2021-06-20
pyGHDL: add ghdl-dom entrypoint

umarcor:
Similarly to `ghdl-ls`, this PR adds a `ghdl-dom` entrypoint when pyGHDL is installed, so that user can analise their sources directly:

```sh
$ ghdl-dom path/to/a/file
# or
$ ghdl-dom path/to/a/file path/to/anothe/file ...
```

For instance:

```sh
# ghdl-dom libraries/ieee/math_real.vhdl libraries/ieee2008/numeric_bit-body.vhdl
Unknown expression kind 'String_Literal8'(11) in expression '1294'.
Unknown expression kind 'Aggregate'(185) in expression '1298'.
```

### DIFF
--- a/pyGHDL/cli/DOM.py
+++ b/pyGHDL/cli/DOM.py
@@ -40,7 +40,7 @@ class Application:
         print("\n".join(buffer))
 
 
-def main(items):
+def main(items=argv[1:]):
     _exitcode = 0
 
     if len(items) < 1:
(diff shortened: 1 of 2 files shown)


umarcor:
I updated this PR so that ghdl-dom returns different error codes depending on the exception.
That allows generating reports such as [report.txt](https://github.com/ghdl/ghdl/files/6693939/report.txt) (rename to `*.html`).

PR #1803, opened 2021-06-22
Fix codacy issues

umarcor:
This PR fixes several issues reported by Codacy, and `black` is used in `libraries/openieee/*.py`.

### DIFF
--- a/pyGHDL/dom/NonStandard.py
+++ b/pyGHDL/dom/NonStandard.py
@@ -80,8 +80,9 @@ class Design(VHDLModel_Design):
         self.__ghdl_init()
 
     def __ghdl_init(self):
-        """Initialization: set options and then load libraries."""
-
+        """
+        Initialization: set options and then load libraries.
+        """
         # Initialize libghdl
         libghdl_finalize()
         libghdl_initialize()
--- a/pyGHDL/dom/NonStandard.py
+++ b/pyGHDL/dom/NonStandard.py
@@ -80,9 +80,7 @@ class Design(VHDLModel_Design):
         self.__ghdl_init()
 
     def __ghdl_init(self):
-        """
-        Initialization: set options and then load libraries.
-        """
+        """Initialization: set options and then load libraries."""
         # Initialize libghdl
         libghdl_finalize()
         libghdl_initialize()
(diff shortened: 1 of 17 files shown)


scripts/pnodespy.py
```diff
@@ -30,18 +30,27 @@ class {0}(IntEnum):
...
             # Call 'make' in 'src/vhdl' to regenerate:
             #
-            from enum import IntEnum, unique
-            from pydecor import export
-
-            from pyGHDL.libghdl._decorator import BindToLibGHDL
         """), end=''
     )
 
+    if includeIntEnumunique:
```
Paebbels, line 41: I think it's better to use this form:

```py
print(dedent("""\
    # Auto generated Python source file from Ada sources
    # Call 'make' in 'src/vhdl' to regenerate:
    #
    {sysImports}
    
    from pydecor import export
    
    {moduleImports}
    """.format(
        sysImports = "from enum import IntEnum, unique" if includeIntEnumunique else "",
        moduleImports  "from pyGHDL.libghdl._decorator import BindToLibGHDL" if includeBindToLibGHDL else "",
    )
))
```

This doesn't break the reading flow so much.

pyGHDL/dom/NonStandard.py
```diff
@@ -80,8 +80,9 @@ def __init__(self):
         self.__ghdl_init()
 
     def __ghdl_init(self):
-        """Initialization: set options and then load libraries."""
-
+        """
```
Paebbels, line 83: I think a online was here ok unless the docstring becomes longer.

Paebbels (changes requested):

scripts/pnodespy.py
```diff
@@ -30,18 +30,27 @@ class {0}(IntEnum):
         print("    {0} = {1}".format(k, n))
 
 
-def print_file_header():
+def print_file_header(includeIntEnumunique=True,includeBindToLibGHDL=True):
```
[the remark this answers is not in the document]
umarcor: I uppercased `U` and I added the space.

Black does not always modify those things. That's why it's non deterministic...

pyGHDL/dom/NonStandard.py
umarcor: Changed.

scripts/pnodespy.py
umarcor: Done.

Paebbels (approved):

PR #1804, opened 2021-06-22
DOM: Functions and types

Paebbels:
**Description**
* Added a null check to utility functions and raise a `ValueError`.
* Preparation for deferred constants.
* Added `EnumerationLiteral` and `EnumeratedType`
* Added `ArrayType`
* Added `RecordTypeElement` and `RecordType`
* Added `AccessType`
* Added `SharedVariable`
* Added generics and parameters to `Function`.
* Added generics and parameters to `Procedure`.
* Added `GetSelectedName`

**Changes:**
* Converted `CheckForErrors` from method to function.
* Print more detailed errors from libghdl incl. filename and problem position.
* Removed `GHDLMixin`.
* Construct design units and other ModelEntities with lists or generators instead of adding items to properties.
* Code cleanups.
* Refactored code and created `GetSubTypeIndicationFromNode` and `GetSubTypeIndicationFromIndicationNode`.
* Added missing type hints.
* Split `ConstrainedSubTypeSymbol` into `ConstrainedScalarSubTypeSymbol` and `ConstrainedCompositeSubTypeSymbol`.
* Renamed test `AllVHDLSources` to `Sanity`.
* Added missing file headers.
* Run *black* in `testsuit/pyuint`


pyGHDL/requirements.txt
```diff
@@ -1,2 +1,3 @@
 pydecor>=2.0.1
-pyVHDLModel==0.10.2
+#pyVHDLModel==0.10.2
```
umarcor, line 2: Before merging, we should release pyVHDLModel 0.10.3, publish it to PyPI and replace this modification with just bumping 0.10.2 to 0.10.3,

pyGHDL/dom/DesignUnit.py
```diff
@@ -61,54 +61,39 @@
...
+        ports = GetPortsFromChainedNodes(nodes.Get_Port_Chain(entityNode))
+        declaredItems = GetDeclaredItemsFromChainedNodes(
             nodes.Get_Declaration_Chain(entityNode), "entity", name
-        ):
-            entity.DeclaredItems.append(item)
+        )
+        bodyItems = []
 
-        return entity
+        return cls(name, generics, ports, declaredItems, bodyItems)
```
umarcor, line 81: In cases such as this one, what do you think about the following style?

```py
        return cls(
            GetNameOfNode(entityNode),
            GetGenericsFromChainedNodes(nodes.Get_Generic_Chain(entityNode)),
            GetPortsFromChainedNodes(nodes.Get_Port_Chain(entityNode)),
            declaredItems,
            []
        )
````

I do understand the value of intermediate variables when those are used more than once, or when the "generation" of them is complex. However, most of these are two chained calls with a single input and a single output.

Comment (umarcor) [reply pyGHDL/dom/DesignUnit.py:81]: Moreover, does it make sense to put the call to `Get_*_Chain` into the `Get*FromChainedNodes`, and rename it to `Get*FromEntityNode`?

File: pyGHDL/dom/DesignUnit.py [hunk 3]
```diff
@@ -119,63 +104,50 @@ class Component(VHDLModel_Component):
     @classmethod
     def parse(cls, componentNode: Iir):
         name = GetNameOfNode(componentNode)
+        generics = GetGenericsFromChainedNodes(nodes.Get_Generic_Chain(componentNode))
+        ports = GetPortsFromChainedNodes(nodes.Get_Port_Chain(componentNode))
```
umarcor, line 108: See comment about `Get*FromPrimaryNode` above.

pyGHDL/dom/DesignUnit.py
Paebbels: Maybe later when we are through all translations.

pyGHDL/requirements.txt
Paebbels: It's fixed to `v0.10.3`.

PR #1805, opened 2021-06-22
Fix more codacy issues

umarcor:
A few more fixes to issues reported by Codacy.

### DIFF
--- a/scripts/pnodespy.py
+++ b/scripts/pnodespy.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-"""Like pnodes but output for Python"""
+"""Like pnodes but output for Python."""
 
 from __future__ import print_function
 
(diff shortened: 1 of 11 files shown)


scripts/pnodespy.py
```diff
@@ -84,13 +84,13 @@ def Get_Location(node: Iir) -> LocationType:
             @export
             @BindToLibGHDL("{classname}__get_{kname_lower}")
             def Get_{kname}(obj: Iir) -> {rtype}:
-                \"\"\"{gettercomment}\"\"\"
```
Paebbels, line 87: Should we really remove this or can @tgingold provide this from Ada files?

umarcor: I think that most of the getters/setters don't have any description/comment in Ada. Hence, we might just write a general comment here.

PR #1806, opened 2021-06-24
More DOM improvements

Paebbels:
**Description**
* Reworked Names
* Wrap calls into libghdl with a try-except block to catch `OSError`.
* Print `LibGHDLException` text in case it was wrapped into a `DOMException`.
* Added `Atrribute` and `AttributeSpecification`.
* Added `UseClause`.
* Handle generics on packages.
* Added `PackageInstantiation`.
* Split `RangeExpression` into `AscendingRangeExpression` and `DescendingRangeExpression`.
  *Note:* `RangeExpression` is now a `BinaryExpression`.
* Added `ParameterFileInterfaceItem`.
* Added `NullLiteral`.
* Added `AllName`.
* Load sourcecode from Python as string into libghdl.  
  *Note:* No file I/O from libghdl itself.
* Added `File` (object).
* Added `PhysicalType`
* Added `ProtectedType` and `ProtectedTypeBody`.
* Fixed testcases.


pyGHDL/libghdl/_decorator.py
```diff
@@ -181,14 +181,34 @@ def wrapper(func: Callable):
...
+                except OSError as ex:
+                    errors = [str(ex)]
+                    raise LibGHDLException(
+                        "Caught exception when calling '{func}' in libghdl.".format(
+                            func=subprogramName
+                        ),
+                        errors,
+                    ) from ex
+
+                return returnType(returnValue)
```
tgingold, line 195: Do we really want this ?  In which cases OSError is raised ?

testsuite/pyunit/lsp/LanguageServer.py
```diff
@@ -197,6 +199,7 @@ def test_Request_Response(self):
 class Test004_Error_Project(JSONTest):
     subdir = Path("004errprj")
 
+    @mark.xfail
```
tgingold, line 202: Why ?  lsp doesn't depend on dom.

pyGHDL/libghdl/_decorator.py
Paebbels: It's raised on every wrong call via libghdl to e.g. IIR nodes. So if a node does not support a data field, it causes a OSError. It just wraps such OSErrors in a generic `LibGHDLException`. This allows then a clean exception handling.

testsuite/pyunit/lsp/LanguageServer.py
Paebbels: This is the test that fails since long time only Windows.
You couldn't fix/reproduce it on you side.

It's about the JSON compares with `/` vs. `\`.

Paebbels:
```
XFAIL [ 85%]
Test004_Error_Project
[0].params.message: mismatch (ref: json error in project file C:\git\GitHub\GHDL\ghdl\testsuite\pyunit\lsp/004errprj/hdl-prj.json:1:3, result: json error in project file C:\git\GitHub\GHDL\ghdl\testsuite\pyunit\lsp\004errprj\hdl-prj.json:1:3)
FAILURE between output and C:\git\GitHub\GHDL\ghdl\testsuite\pyunit\lsp\004errprj\replies.json (for C:\git\GitHub\GHDL\ghdl\testsuite\pyunit\lsp\004errprj\cmds.json)
Writing result output to result.json
```

tgingold: Is it possible to mark it xfail only on windows platforms ?

pyGHDL/libghdl/_decorator.py
tgingold: Ok, but what raise the exception ?  Is it python that convert the Ada exception to a python one ?

Paebbels: I'm not sure if it's an Ada exception, because a `STORAGE_ERROR` and `ASSERT_FAILURE` from Ada (I assume it's internally an exception) are just printed to stderr from libghdl. It would be nice if we could convert them to Python exceptions.

You can see it in the extended-tests: https://paebbels.github.io/extended-tests/  
(open some of the xfailed from OSVVM or UVVM)
> `raised STORAGE_ERROR : stack overflow or erroneous memory access`

or
> `raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : no field Direction`

If the number of parameters is wrong, I think CDLL would raise an exception on the Python side. So having an OSError might be a memory access violation.
The general Python documentation writes:
> This exception is raised when a system function returns a system-related error, including I/O failures such as “file not found” or “disk full” (not for illegal argument types or other incidental errors).

I'm not sure how CDLL uses the OSError.

testsuite/pyunit/lsp/LanguageServer.py
Paebbels: I'm not aware of a method to make `xfail` platform specific. I have reverted the change and pushed again.

This should finish in 20 minutes. I needed to force push again due to black complaints.

umarcor: We can xfail or skip specific tests on Windows or on MSYS2 only:

- https://github.com/ghdl/ghdl-cosim/blob/419bd73fd5ff8e22faa430925a994dae96cbcc0d/tests/common.py#L13
- https://github.com/ghdl/ghdl-cosim/blob/master/tests/test_vhpidirect_vffi.py#L16-L21
- https://github.com/ghdl/ghdl-cosim/blob/master/tests/test_vhpidirect_shared.py#L26-L29

tgingold:
Thanks!

PR #1808, opened 2021-06-28
CLI Update for DOM

Paebbels:
**Description**
* Incremented pyVHDLModel version to v0.11.1.
* Added needed dependencies for the new CLI of DOM.
* New command line interface for DOM (`ghdl-dom`).
  * Use command `pretty` to pretty-print the internal DOM structure.
    * Use `-f <file>` or `--file <file>` to specify a source file.  
      (Note can be used multiple times.)
    * Use `-F <file> ... <file>` or `--files <file> ... <file>` to specify multiple source files.
* Renamed `DOM.py` to `dom.py`.
  (Note: the entry point wrapper `ghdl-dom`, which is installed via setuptools is not changed.)
* Reduced initializer code, because pyVHDLModel offers now parameters in `__init__(...)`.
* Renamed parameter `name` to `identifier` if it's of type `str`, so `name` can be used for type `Name`.
* Implemented `AttributeSpecification`.
* Added matching expressions.
* Added allocations
* Added `GenericTypeInterfaceItem`.
* Added minimal implementations for
  * packages, procedures and functions as generic
  * PSL vunit, vprop, vmode and default clock
* Skipped implementation of
  * disconnect specification
  * group and group template
* Reworked `Range`.
* Reworked Symbols.
* Fixed `PhysicalType`.
* Split function `GetTypeFromNode` into two parts.  
  (new: `GetAnonymousTypeFromNode`)
* Updated routines in `prettyprint.py`.
* Added support for packages in declaration regions.
* Fixed Codacy warnings. E.g. added *return*-statements to dummy function wrappers which are using the `BindToLibGHDL` decorator.
* Improved pyunit test code.


Paebbels:
I have rebased. It looks like it will also pass because master is ok again.

Paebbels:
@tgingold CI is now OK. I also added the list of changes to this PR. You can review and merge.

pyGHDL/requirements.txt
```diff
@@ -1,3 +1,6 @@
 pydecor>=2.0.1
-pyVHDLModel==0.10.4
+pyAttributes==2.1.0
```
umarcor, line 2: Shall we create a requirements file for the DOM which is separated from pyGHDL?
It's ok to install all of those packages by default, but some users might want the LSP only, so they can get along with fewer dependencies.
So: `requirements.dom.txt`, `requirements.lsp.txt` and `requirements.txt` (including both of the others).

pyGHDL/cli/dom.py
```diff
@@ -0,0 +1,374 @@
...
+        self.WriteNormal("License:    {0}".format(__license__))
+        authors = __author__.split(", ")
+        self.WriteNormal("Authors:    {0}".format(authors[0]))
+        for author in authors[1:]:
+            self.WriteNormal("            {0}".format(author))
+        self.WriteNormal("Version:    {0}".format(__version__))
+        self.exit()
+
+    # ----------------------------------------------------------------------------
+    # create the sub-parser for the "token-stream" command
```
umarcor, line 269: Is that "token-stream" correct or a pasto?

Paebbels: You're right, that's pesto (from pyVHDLParser 😄)

pyGHDL/requirements.txt
Paebbels: Hmmm or the other way around, a `requirements.txt` for the minimal stuff and the `requirements.xxx.txt` for the extended usecases. I would prefer the latter approach, because that's what we do e.g. in docs where re recursively source the main file and add additional packages for docs.

I'm not sure if `requirements.xxx.txt` will be understood by dependency checkers. we might need separate directories from them. But this makes it quite complex, so maybe you're approch is better as it provides a simple file at top for everyone and for user who complain we have specific files ...

umarcor: I'm thinking about providing a default file and requiring users to understand others if they want a minimal setup. Hence, I don't think the structure is very relevant, as long as we document it. Up to you!

Maybe?

- `requirements.min.txt`
- `requirements.dom.txt` or `dom/requirements.txt`
- `requirements.lsp.txt` or `lsp/requirements.txt`
- `requirements.txt` (importing the three before; this is the default)

PR #1809, opened 2021-06-29
ci: generate standalone zipfiles and add CPython jobs

umarcor:
This PR enables the mcode build on MINGW64. Although code generation is not supported, features other than simulation do work.

Moreover, a job is added after testing on Windows, for generating "standalone" zipfiles for mcode32 and mcode64. Precisely, the MSYS2 packages are extracted and a shell script contributed by @Paebbels is used for adding DLLs.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -68,7 +68,7 @@ jobs:
         path: doc/_build/man/ghdl.1
 
 #
-# pyGHDL
+# pyGHDL Bindings and Formatting
 #
 
   fmt:
@@ -194,7 +194,7 @@ jobs:
         path: ghdl-macos*${{ matrix.backend }}.tgz
 
 #
-# Windows Build
+# Windows MSYS2 Build
 #
 
   win-build:
@@ -256,7 +256,7 @@ jobs:
         path: scripts/msys2-${{ matrix.pkg }}/mingw-*ghdl*.pkg.tar.zst
 
 #
-# Windows Test
+# Windows MSYS2 Test
 #
 
   win-test:
@@ -314,7 +314,7 @@ jobs:
       run: GHDL=ghdl ./testsuite/testsuite.sh ${{ matrix.suite }}
 
 #
-# Windows pyGHDL
+# Windows pyGHDL pip installation
 #
 
   win-pip:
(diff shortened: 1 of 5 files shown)


umarcor:
This is now updated for testing pyGHDL in the following environments:

- Install on MSYS2 through pip.
- CPython with GHDL on MSYS2 and pyGHDL used through PYTHONPATH. This is tested with MINGW32-mcode, MINGW64-mcode and MINGW64-LLVM
- CPython with GHDL installed through a tarball and pyGHDL used through PYTHONPATH. Currently, standalone zipfiles are generated for mcode32 and mcode64. Hence, this is tested with mcode only.

As seen in CI results, 64 bits tests are successful, but 32 bits fails. So, 2/3 of the CPython + MSYS2 jobs are successful, and 1/2 of the CPython + standalone is successful.

@Paebbels, @tgingold, how shall we proceed? We can either remove the failing tests, or keep them but ignore their result. That will not CI fail, but the status of the workflow will be a red cross instead of a green checkmark.

---

By the way, I updated the icons and job names, so it is hopefully more descriptive.

- 🚧: build.
- 🚦: test.
- 🐧: GNU/Linux.
- 🍎: macOS.
- 🟪: MINGW32.
- 🟦: MINGW64.
- 🥡: standalone.
- 🐍: Python.
- ⛴️: build container.
- 🛳️: run container.

---

I also marked Test004_Error_Project as XFAIL when CPython is used, because there is some error (which is not produced on MSYS2).

Paebbels:
@umarcor can we have a meeting and discuss some details?

umarcor:
All the CPython + MSYS2 jobs are successful now; CPython + standalone mcode64 works too; but CPython + standalone mcode32 fails. It is complaining about `FileNotFoundError: Could not find module 'D:\a\ghdl\ghdl\GHDL-standalone\lib\libghdl-2_0_0_dev.dll' (or one of its dependencies). Try using the full path with constructor syntax.`. However, that is probably because of some other DLL which that library depends on. Actually, the zipfile generation scripts seems not to resolve the dependencies on MINGW32: https://github.com/ghdl/ghdl/runs/3099086379?check_suite_focus=true#step:7:16. Therefore, I disabled it (commented it out) so that we can merge this PR.

Paebbels:
All is now working and fine. Some minor remaining tasks will be changed with new PRs.

PR #1811, opened 2021-06-30
Added DLL search path for Python ≥3.8.

Paebbels:
**Problem Description**
If GHDL is installed standalone outside of MinGW (tested with MSYS2/MinGW64), all DLL dependencies are resolved correctly. This is independent of GHDL / libghdl being build with mcode or llvm backend.

However, if GHDL / libghdl and it's dependencies are not copied outside of MSYS2 (e.g. `C:\msys64\mingw64\...` to `C:\Tools\GHDL`), then Pythons ctypes.CDLL has problems in finding all dependencies. This is due to the fact that `%PATH%` is not correctly used to search for dependencies.

**Solution**
With CPython 3.8, a new API (`os.add_dll_directory`) was added to add a DLL search path. This is only necessary on Windows platforms, because the problem does not exist on Linux and more over Linux would offer a `LD_LIBRARY_PATH` variable to support shared library searches.

This PR adds code to support using GHDL that was installed via `pacman` inside of MinGW64 to be used from Python packages running in a Windows native Python installation (CPython).

**Change**
```diff
    # Load the shared library
    _libghdl_path = _get_libghdl_path()

+    # Add DLL search path(s)
+    if sys_platform == "win32":
+        from os import add_dll_directory as os_add_dll_directory

+        ghdlBinPath = _libghdl_path.parent.parent / "bin"
+        os_add_dll_directory(str(ghdlBinPath))

    # Load libghdl shared object
    libghdl = CDLL(str(_libghdl_path))
```

### DIFF
--- a/pyGHDL/libghdl/__init__.py
+++ b/pyGHDL/libghdl/__init__.py
@@ -31,10 +31,9 @@
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ============================================================================
-
 from ctypes import c_char_p, CDLL
-import os
-import sys
+from sys import platform as sys_platform, version_info as sys_version_info
+from os import environ as os_environ
 from pathlib import Path
 from shutil import which
 from typing import List
@@ -63,7 +62,7 @@ class LibGHDLException(GHDLBaseException):
 def _get_libghdl_name() -> Path:
     """Get the name of the libghdl library (with version and extension)."""
     ver = __version__.replace("-", "_").replace(".", "_")
-    ext = {"win32": "dll", "cygwin": "dll", "darwin": "dylib"}.get(sys.platform, "so")
+    ext = {"win32": "dll", "cygwin": "dll", "darwin": "dylib"}.get(sys_platform, "so")
     return Path("libghdl-{version}.{ext}".format(version=ver, ext=ext))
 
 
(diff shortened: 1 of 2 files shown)


Paebbels:
@tgingold any idea why LSP suddenly fails with encoding problems?

See https://github.com/ghdl/ghdl/runs/2953246138?check_suite_focus=true#step:7:286
```py
    def read_request(self):
        headers = {}
        while True:
            # Read a line
            line = self.conn.readline()
            # Return on EOF.
            if not line:
                return None
            if line[-2:] != "\r\n":
                raise ProtocolError("invalid end of line in header")
            line = line[:-2]
            if not line:
                # End of headers.
                log.debug("Headers: %r", headers)
                length = headers.get("Content-Length", None)
                if length is not None:
                    body = self.conn.read(int(length))
                    return body
                else:
>                   raise ProtocolError("missing Content-Length in header")
E                   pyGHDL.lsp.lsp.ProtocolError: missing Content-Length in header
```

Paebbels:
@tgingold it's ready for review.

(I have no idea why the VM for macOS died in the middle.)

tgingold:
Ok, but what about python < 3.8 ?  Does it make sense to check the python version ?

Paebbels:
I have added a version check. Versions before 3.8 behave as before.
```py
    if (
        sys_platform == "win32"
        and sys_version_info.major == 3
        and sys_version_info.minor >= 8
    ):
```

PR #1813, opened 2021-07-03
Read requirement files recursively

Paebbels:
**Description**

This PR fixes issues in `setup.py` and it's the function to read dependencies from `requirements.txt`.

Since last PR for DOM, the `requirements.txt` has been split into multiple files, so users can choose what to install depending on the usecase. Common requirement files are included via `-r ...` option. The new read function can now handle these recursive includes. Same includes are also deduplicated.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,13 +63,17 @@ def get_requirements(file: Path) -> List[str]:
     requirements = []
     with file.open("r") as fh:
         for line in fh.readlines():
-            if line.startswith("#"):
+            line = line.strip()
+            if line.startswith("#") or line == "":
                 continue
+            elif line.startswith("-r"):
+                filename = line[3:].strip()
+                requirements += get_requirements(file.parent / filename)
             elif line.startswith("https"):
-                _splitItems = line.strip().split("#")
+                _splitItems = line.split("#")
                 requirements.append("{} @ {}".format(_splitItems[1], _splitItems[0]))
             else:
-                requirements.append(line.strip())
+                requirements.append(line)
     return requirements
 
 
@@ -96,6 +100,8 @@ documentationURL = (
         namespace=gitHubNamespace, projectName=projectName
     )
 )
+requirements = list(set(get_requirements(requirementsFile)))
+
 
 # Assemble all package information
 setuptools_setup(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -96,6 +99,8 @@ def get_version():
         namespace=gitHubNamespace, projectName=projectName
     )
 )
+requirements = list(dict.fromkeys(get_requirements(requirementsFile)))
```
Xiretza, line 102: If the order of requirements is not significant (I don't think it is?), `list(set(get_requirements(requirementsFile)))` might be a bit clearer here.

setup.py
```diff
@@ -63,8 +63,11 @@ def get_requirements(file: Path) -> List[str]:
     requirements = []
     with file.open("r") as fh:
         for line in fh.readlines():
-            if line.startswith("#"):
+            if line.startswith("#") or line.strip() == "":
```
umarcor, line 66: It seems that we are doing the same strip operation on the same variable in three different places. What about `_line = line.strip()` right after L65, so we can then remove `strip()` from L66, L69 and L72?

setup.py
```diff
@@ -63,8 +63,11 @@ def get_requirements(file: Path) -> List[str]:
...
                 continue
+            elif line.startswith("-r"):
+                filename = line[3:].strip()
```
umarcor, line 69: If we move strip above, `line[3:]` can be written in L70 without variable `filename`.

setup.py
Paebbels: for new Python versions (so starting around 3.6), `list(set(...))` is not recommended anymore.

setup.py
Paebbels: No the strip operations are not identical.

Btw. don't name local variables `_...` (I saw this multiple times in your code.)
These are private members of classes.

setup.py
Xiretza: Interesting, do you have a source for that? I can't really imagine `dict.fromkeys` being more idiomatic than `set` even though there is no dict involved.

setup.py
Paebbels: There is no performance improvement by removing variables.

At first you remove documentation from code and you complicate reviews, as it's unclear what information is in the substring `[3:]`. At second you looking forward for a conflict with black, which will then cripple the code.

setup.py
Paebbels: Thanks @Xiretza. I did tests with `timeit.repeat(...)` myself and the articles about list deduplication are not correct.
A modified version has been force-pushed.

setup.py
Paebbels: @umarcor Oh there are 2 more strips ...

See update version.

umarcor:
I rebased on top of master.

PR #1822, opened 2021-07-25
Paebbels/dom

Paebbels:
# New Features
* Handle multiple identifiers in generics, ports, parameters and objects.
* `ghdl-dom` now also accepts `-D` for directories to scan.
* Resolve architectures to entities.
* Context reference
* Library clause
* Use clause
* Handle contexts of design units
* New `OpenName`
* Translate concurrent statements:
  * Component instantiation
  * Entity instantiation
  * Configuration instantiation
  * If..generate statement
  * Case..generate statement
  * For..generate statement
  * Block statement
  * Process statement
  * Concurrent simple signal assignment
  * Concurrent procedure call
* Translate sequential statements:
  * If statement
  * Case statement
  * For loop
  * Sequential simple signal assignment
  * Sequential procedure call
  * Sequential assert statement
  * Sequential report statement
  * Wait statement
* Print hierarchy in pretty-print
* New binding to `str_table` `string8_address`

# Changes
* Adjusted to renaming of `pyVHDLModel.VHDLModel` to `pyVHDLModel.SyntaxModel`.
* Adjust DOM to a change in pyVHDLModel: some Identifiers being now a list of identifiers.
* Reordered items in GHA workflow `Test.yml`.
* Improved ranges

# Bug fixes
* Fixed typo in IIR translation of `Greater_Than_Or_Equal_Operator`: should be `GreaterEqualExpression`.
* Wrap type marks in a `SimpleName`.
* Fixed syntax of lists in GHA workflow `Test.yml`.
* Fixed handling of bit-string literals.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -16,671 +16,692 @@ jobs:
 #
 
   doc:
-    runs-on: ubuntu-latest
     name: '📓 Docs'
+    runs-on: ubuntu-latest
+
     steps:
+      - name: '🧰 Checkout'
+        uses: actions/checkout@v2
 
-    - name: '🧰 Checkout'
-      uses: actions/checkout@v2
+      - name: '⛴ Build ghdl/doc'
+        run: |
+          docker build -t ghdl/doc . -f- <<-EOF
+          FROM ghdl/debug:base
+          ENV PYTHONPATH=/opt/ghdl/pyGHDL
+          COPY . /opt/ghdl
+          RUN cd /opt/ghdl && ./configure && make && make install
+          EOF
 
-    - name: '⛴️ Build ghdl/doc'
-      run: |
-        docker build -t ghdl/doc . -f- <<-EOF
-        FROM ghdl/debug:base
-        ENV PYTHONPATH=/opt/ghdl/pyGHDL
-        COPY . /opt/ghdl
-        RUN cd /opt/ghdl && ./configure && make && make install
-        EOF
+      - name: '📚 Run gnatdoc'
+        run: |
+          cat > run.sh <<-EOF
+          #!/usr/bin/env sh
+          gnatdoc -P./ghdl
+          mkdir /src/public
+          mv gnatdoc /src/public/gnatdoc
+          EOF
+          chmod +x run.sh
+          docker run --rm -v $(pwd):/src -w /opt/ghdl ghdl/doc /src/run.sh
 
-    - name: '📚 Run gnatdoc'
-      run: |
-        cat > run.sh <<-EOF
-        #!/usr/bin/env sh
-        gnatdoc -P./ghdl
-        mkdir /src/public
-        mv gnatdoc /src/public/gnatdoc
-        EOF
-        chmod +x run.sh
-        docker run --rm -v $(pwd):/src -w /opt/ghdl ghdl/doc /src/run.sh
+      - name: '📓 BuildTheDocs (BTD)'
+        if: github.event_name != 'pull_request'
+        uses: buildthedocs/btd@v0
+        with:
+          token: ${{ github.token }}
 
-    - name: '📓 BuildTheDocs (BTD)'
-      if: github.event_name != 'pull_request'
-      uses: buildthedocs/btd@v0
-      with:
-        token: ${{ github.token }}
+      #- run: nroff -man doc/_build/man/ghdl.1
 
-    #- run: nroff -man doc/_build/man/ghdl.1
+      - name: '📤 Upload artifact: HTML and LaTeX'
+        if: github.event_name != 'pull_request'
+        uses: actions/upload-artifact@v2
+        with:
+          name: doc
+          path: |
+            doc/_build/html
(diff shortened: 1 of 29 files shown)


pyGHDL/dom/_Translate.py
```diff
@@ -472,45 +467,75 @@ def GetGenericsFromChainedNodes(
...
+
+    port = nodeChain
+    while port != nodes.Null_Iir:
         kind = GetIirKindOfNode(port)
         if kind == nodes.Iir_Kind.Interface_Signal_Declaration:
             from pyGHDL.dom.InterfaceItem import PortSignalInterfaceItem
 
             portSignal = PortSignalInterfaceItem.parse(port)
 
+            if nodes.Get_Has_Identifier_List(port):
```
Paebbels, line 528: @tgingold This solves how to handle multiple identifiers, right?

1. A manual iteration is needed
2. A lookahead is needed for certain node types (here a port signal)
3. If the subtype indication is Null, add this identifier to the existing currently created port.
4. If the subtype indication is present, a new port was found.

Am I right, that no other element is in between these nodes, so no need to check the kind of `nextPort`?

-----------------
In pyVHDLModel, it's not expanded to multiple ports. It's one port with a list of identifiers.

.github/workflows/Test.yml
```diff
@@ -16,671 +16,692 @@ jobs:
...
-          doc/_build/latex
-          doc/_build/man
-
-    - name: '📤 Upload artifact: ghdl.1'
-      if: github.event_name != 'pull_request'
-      uses: actions/upload-artifact@v2
-      with:
-        name: man
-        path: doc/_build/man/ghdl.1
+    steps:
```
umarcor, line 22: Please, either remove the space before, or leave a blank line both before and after `steps`.

.github/workflows/Test.yml
```diff
@@ -16,671 +16,692 @@ jobs:
...
         shell: pwsh
+
     steps:
+      - name: '${{ matrix.icon }} Setup MSYS2'
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: MINGW${{ matrix.bits }}
+          update: true
+
+      - name: '⚙ git config'
```
umarcor, line 439: This icon/emoji seems to be different from the other "gears". The same with the '🛠️' in the two steps below.

.github/workflows/Test.yml
```diff
@@ -16,671 +16,692 @@ jobs:
...
         include:
-        #- { pkg: '32-llvm', pyarch: x86 } ! not yet functional
-        #- { pkg: '64-llvm', pyarch: x64 }
-        #- { pkg: '32-mcode', pyarch: x86 } #! the tarball generation on MINGW32 needs to be fixed
-        - { pkg: '64-mcode', pyarch: x64 } #! simulation with mcode is not yet supported on win64
-    name: '🚦🐍🥡 ${{ matrix.pkg }}'
+#          - {name: 'llvm32', pkg: '32-llvm', pyarch: x86 } ! not yet functional
+#          - {name: 'llvm64', pkg: '64-llvm', pyarch: x64 }
+#          - {name: 'mcode32', pkg: '32-mcode', pyarch: x86 } #! the tarball generation on MINGW32 needs to be fixed
+          - {name: 'mcode64', pkg: '64-mcode', pyarch: x64 } #! simulation with mcode is not yet supported on win64
```
umarcor, line 506: If field `name` is added, better use `pkg` and `bits`, as in other jobs. So the "name" would be `${{ matrix.pkg }}${{ matrix.bits }}`.

.github/workflows/Test.yml
```diff
@@ -16,671 +16,692 @@ jobs:
...
-      run: |
-        ghdl-dom help
-        ghdl-ls --help
-
-    - name: '🚦 Test pyunit testsuite'
-      run: |
-        cd testsuite
-        python3 -m pytest -vsrA pyunit
+    steps:
+      - name: '⚙ git config'
```
umarcor, line 513: Same issue with emojis as commented above.

.github/workflows/Test.yml
```diff
@@ -16,671 +16,692 @@ jobs:
...
+      - name: '🟪 Setup MSYS2'
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: MINGW64
+          update: true
+          install: >
+            mingw-w64-x86_64-python-pip
+            mingw-w64-x86_64-python-setuptools
+
+      - name: '⚙ git config'
```
umarcor, line 642: Emojis/icons..

.github/workflows/Test.yml
umarcor: Understood that the difference is because of the font used by the editor (web gui).

Paebbels:
@umarcor some problems are fixed now. Especially the doubled use of class name `Context` and circular references after solving that issue.

All checks have passed except for many false positives from Codacy.

-------------
Maybe we need to assemble a MWE for pylint to write an issue...

umarcor:
🚀

PR #1825, opened 2021-07-29
libghw: add missing semicolon to physical type formatting

Xiretza:
**Description**

When printing the type section, physical types do not have a trailing semicolon. This causes problems when comparing the golden outputs in the testsuite with my [Rust libghw/ghwdump reimplementation](https://gitlab.com/xiretza/ghw-rs), which always prints a trailing semicolon for any type table entry. While the rest of the expected output is not completely valid VHDL either, this just looks like an oversight.


tgingold:
Thanks.

PR #1827, opened 2021-07-30
TEST DOM crash with multiple files

umarcor:
This PR is not for merging, but for showing a problem when analysing multiple files with ghdl-dom at once.
In the regular CI tests we analyse each file separatedly, in order to avoid this problem.
Actually, we hit it before and Tristan did fix it partially.

In this PR, repo NEORV32 is cloned and `rtl/core/*.vhd` are analysed at once. That generates a segmentation fault.

```
set -e
git clone https://github.com/stnolting/neorv32
ghdl-dom pretty -F neorv32/rtl/core/*.vhd
```

```
Cloning into 'neorv32'...
================================================================================
                         pyGHDL.dom - Test Application                          
================================================================================
Parsing file 'neorv32\rtl\core\neorv32_application_image.vhd'
D:\a\_temp\fbe4ff3b-599e-4360-9380-9eb730caa92d.sh: line 3:   588 Segmentation fault      ghdl-dom pretty -F neorv32/rtl/core/*.vhd
Error: Process completed with exit code 139.
```

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -493,6 +493,13 @@ jobs:
           cd testsuite
           python3 -m pytest -vsrA pyunit
 
+      - name: Test NEORV32
+        shell: bash
+        run: |
+          set -e
+          git clone https://github.com/stnolting/neorv32
+          ghdl-dom pretty -F neorv32/rtl/core/*.vhd
+
 #
 # Windows CPython pyGHDL Test with standalone zipfile and pyGHDL wheel
 #
@@ -571,6 +578,13 @@ jobs:
         run: |
           cd testsuite
           python3 -m pytest -vsrA pyunit
+
+      - name: Test NEORV32
+        shell: bash
+        run: |
+          set -e
+          git clone https://github.com/stnolting/neorv32
+          ghdl-dom pretty -F neorv32/rtl/core/*.vhd
 
 #
 # Release


tgingold:
That's an issue in the python code.

I suppose this is a liveness issue on the identifiers string.
Does it work if name_table.Get_Name_Ptr always returns 'foo' ?

Paebbels:
I'm not sure how `name_table.Get_Name_Ptr` should cause this.

What we see (and feel) is: The more or bigger files we add, then it gives a segmentation fault. Depending on the number or size of files or comes earlier or later.

Strings returned by `name_table.Get_Name_Ptr` are copied into Python. Each string is UTF de/encoded, so it should be a copy of the original string. So if it becomes invalid memory from ctypes or libGHDL, it should cause a segmentation fault in Python because it was copied, right?

The CLI program `ghdl-dom` has currently one command `pretty` for testing purposes. With `-F` you can handover a list of VHDL files (e.g. by shell expansion) for local testing.


Can you explain more in detail what memory/string you assume to be deallocated thus an access to it gives a segmentation fault?

tgingold:
StringLiteral.parse is incorrect: it calls `Get_Name_Ptr` with a string id.

Paebbels:
Sorry, I don't get it.
Isn't `Get_String8_Id` meant to return an ID that can be used by `Get_Name_Ptr` to get the string?

Hmmm I can't find `Get_String8_Id` in LSP.

----------------
Can you propose a corrected code snippet?

```py
@export
class StringLiteral(VHDLModel_StringLiteral, DOMMixin):
    def __init__(self, node: Iir, value: str):
        super().__init__(value)
        DOMMixin.__init__(self, node)

    @classmethod
    def parse(cls, literalNode: Iir) -> "StringLiteral":
        stringID = nodes.Get_String8_Id(literalNode)
        value = name_table.Get_Name_Ptr(stringID)
        return cls(literalNode, value)
```

tgingold:
Get_String8_Id returns a String8_Id, but Get_Name_Ptr requires a Name_Id.  That doesn't match.

The function String8_Address (from str_table.ads) should be called.  But I think there is no corresponding binding function in libghdl (yet).

umarcor:
I rebased this branch after Paebbels/ghdl#4. There is now a different error:

https://github.com/ghdl/ghdl/runs/3213889525?check_suite_focus=true#step:13:46
```
 Parsing file 'neorv32\rtl\core\neorv32_busswitch.vhd'
LIB: Caught exception when calling 'vhdl__nodes__get_subtype_indication' in libghdl.
  [WinError 541541187] Windows Error 0x20474343
```

Paebbels:
@umarcor please try again with changes from https://github.com/Paebbels/ghdl/commit/184134099af726cbf6751a123f02e5631d39f3da

I tested the failing file locally and it worked now.

umarcor:
@Paebbels, almost there. The Debian Buster job is failing (GPL) and the mcode32 has some encoding issue:

```
 Parsing file 'neorv32\rtl\core\neorv32_cpu_cp_muldiv.vhd'
FATAL: An unknown or unhandled exception reached the topmost exception handler!
  Exception type:      UnicodeDecodeError
  Exception message:   'utf-8' codec can't decode byte 0x82 in position 39: invalid start byte
  Caused in:           Get_String8_Ptr in file 'c:\hostedtoolcache\windows\python\3.8.10\x86\lib\site-packages\pyGHDL\libghdl\str_table.py' at line 60
```

However, the other 3/4 environments are successful.

Paebbels:
Any idea about UTF-8 decoding problem in mcode32 (did not raise in mcode64 or llvm64)
https://github.com/ghdl/ghdl/runs/3215090195?check_suite_focus=true#step:13:336

```
Parsing file 'neorv32\rtl\core\neorv32_cpu_cp_muldiv.vhd'
FATAL: An unknown or unhandled exception reached the topmost exception handler!
  Exception type:      UnicodeDecodeError
  Exception message:   'utf-8' codec can't decode byte 0x82 in position 39: invalid start byte
  Caused in:           Get_String8_Ptr in file 'c:\hostedtoolcache\windows\python\3.8.10\x86\lib\site-packages\pyGHDL\libghdl\str_table.py' at line 60
```

umarcor:
In a debug session with @Paebbels, we found that the strings we are getting from libghdl are not terminated. Therefore, the Python conversion function was trying to read past the limit of the string until it found problems. We fixed it in https://github.com/ghdl/ghdl/pull/1827/commits/afee8309e4b644e0e94c1938c0f4e211ae3038fa, which Patrick will add to #1822.

@tgingold, is it possible to get the terminated string from libghdl instead of using the length in Python explicitly?

tgingold:
Sorry, I didn't read the spec before replying about str_table.
In fact, this is even worst than that: string8 in str_table are not even ASCII strings, they are strings of enumeration value.
So for strings of `std.standard.string`, this corresponds to an ASCII strings.  So you also need the length to extract it.
But for `std_logic_vector` (or `bit_vector`, or any other strings), the elements of the strings are the position of the enumerated values, so they cannot be directly converted to python `str`.

Paebbels:
We changed the binding to this:
```py
@export
def Get_String8_Ptr(Id: String8Id, Length: int) -> str:
    """
    Get the address of string8 ID. Note that as soon as a character is appended
    (using Append_String8) or a string8 is resized (using Resize_String8), an
    address previously returned is not valid anymore.

    :param Id: String8Id for the string to query.
    :return:   String8 as string.
    """
    return _String8_Address(Id)[:Length].decode("utf-8")
```

It uses the `Get_String_Length` from the IIR node:
```py
@export
class StringLiteral(VHDLModel_StringLiteral, DOMMixin):
    def __init__(self, node: Iir, value: str):
        super().__init__(value)
        DOMMixin.__init__(self, node)

    @classmethod
    def parse(cls, literalNode: Iir) -> "StringLiteral":
        if nodes.Get_Bit_String_Base(literalNode) is nodes.NumberBaseType.Base_None:
            value = str_table.Get_String8_Ptr(
                nodes.Get_String8_Id(literalNode), nodes.Get_String_Length(literalNode)
            )
            return cls(literalNode, value)
        else:
            print("[NOT IMPLEMENTED] Bit String Literal not supported yet")
```

As far as I know, the parser can't know if a string is a text string or bit string: `"hello"` vs. `"0101"` problem in VHDL.
When does GHDL convert the latter string to an array of std_logic enumerations?

I found the `Get_Bit_String_Base` function and `nodes.NumberBaseType` enumeration. I think I can assume it's such an enumeration array if the modifiers `b`, `o`, `d`, `x` are used. But what about no modifier?

It's also currently unclear to me how to get the oder modifiers:
* length
* signed/unsigned

umarcor:
I rebased this PR after merging #1822 into master. This seems not to fail anymore.

@Paebbels, what do you suggest? Shall we consider this to be fixed or should be make it break again (maybe adding more sources at once)?

umarcor:
In a video meeting with @Paebbels, we decided to close this because the remaining problem is not related to having multiple files. See #1846.

PR #1846, opened 2021-08-24
TEST execute dom Sanity tests with a single instance

umarcor:
This is to keep track of a very annoying problem which we could not understand yet.

Currently, the testsuite of the DOM executes each test isolated, by calling ghdl-dom as an external program. It ensures that each file is analysed on its own. We did so because of the problems we had when analysing too many files. Those problems are solved now, so we don't need to keep using `executable` from `sys`.

However, for some unexplained reason, if we do "reuse" the instance some other apparently unrelated tests in other files do fail. Precisely, `pyunit/dom/SimpleEntity.py` and `pyunit/dom/SimplePackage.py` fail, despite the change in this PR being related to `pyunit/dom/Sanity.py`.

We don't know whether that's related to libghdl, pyGHDL or pytest. This is not a regression, but something we avoided since the beginning (https://github.com/ghdl/ghdl/pull/1797#issuecomment-864403511).


Paebbels:
Tests are stable and no special effect are visible anymore.

PR #1861, opened 2021-09-12
add cpp header check for libghw

mateoconlechuga:
C++ compilers mangle the names in the libghw header. This prevents that from happening so they can properly be linked to a C++ application.


tgingold:
Thanks!

PR #1865, opened 2021-09-14
Warning on incomplete process sensitivity lists new

### LINKED ISSUE #1425: Warning on incomplete process sensitivity lists
I'm usually checking my RTL simulations against the behavior of an actual FPGA implementation. It's always frustrating to see different behavior in the simulation and on the actual hardware.
In many cases I've seen, this had been caused by an incomplete sensitivity list of a process.
Checking this manually can be very time consuming in a big project. For the time being, the fastest way for me to find these issues is to compile with ModelSim, which shows warnings for signals that are used in a process and are not part of the sensitivity lists.

It would be nice to have a warning in GHDL (`--warn-sensitivity`) that shows all signals used in a process body, which are not part of the process' sensitivity list. 

I know that some designs are making use of this to accelerate simulation. Thus, it may be appropriate to disable this warning by default.

Blebowski:
Close #1425

**Description** 

This pull request implements `--check-synthesis` analysis option which is intended to check for good RTL
design practices and reveal potential synthesis/simulation mismatch. Currently, following checks are executed:
 1. Check for incomplete sensitivity lists
 2. Check for over-specified sensitivity lists.
 Note that clocked processes are skipped in this dummy implementation (not checked). A process is classified as
clocked process if it contains 'event attribute or a call to one of "rising_edge" or "falling_edge" functions.

The question of detecting what is clocked process and how is "expected" sensitivity list constructed might
still be fine-grained in future (e.g. constructs within --pragma synthesis off should be ignored and not reported as
false positively missing from sensitivity lists).

- [x] DO make sure that the build is successful.

**Further comments**

A test is not added since this is only a draft of the feature, further also a check of "clocked" processes will be added.
Next, it might be good to walk through the nodes, and emit warnings on grammar constructs which are not recommended
for synthesis by: IEEE 1076.6-2004 - IEEE Standard for VHDL Register Transfer Level (RTL) Synthesis (e.g. shared
variable). Similarly, this standard defines that "after" clasues shall be ignored by synthesis, so there could be warning
if this would be used.

Effectively, the ultimate goal would be having "RTL linter" with this feature.

:heart: Thank you!

PS: Sorry for many commits, I had trouble re-generating pyGHDL and all the Internal node files. It might be good to
squash the commits.


doc/using/InvokingGHDL.rst
```diff
@@ -500,6 +500,19 @@ Options
   Currently, VITAL checks are only partially implemented.
   See section :ref:`VHDL_restrictions_for_VITAL` for more details.
 
+.. option:: --check-synthesis
```
tgingold, line 503: I am wondering if the name of the flag is the right one.  As this is a flag for warnigs, I would propose something like  `-Wsynthesis` (or `-Wsynth-sensitivity-list` if not too long).

Then you would be able to directly use the warning mechanisms: Use `Is_Warning_Enabled` to test if this option is present.

src/vhdl/vhdl-canon.adb
```diff
@@ -80,6 +84,30 @@ package body Vhdl.Canon is
...
+   begin
+      if not Canoning_Process then
+         return;
+      end if;
+
+      case Get_Kind (Node) is
+      when Iir_Kind_Function_Call =>
+         Func_Identifier := Get_Identifier (Get_Implementation (Node));
+         if Name_Table.Image (Func_Identifier) = "rising_edge" or
+            Name_Table.Image (Func_Identifier) = "falling_edge"
```
tgingold, line 98: Technically this is not correct.  A user can define its own function `rising_edge` that does something differently (OK, that doesn't happen).
You can check if the function is `rising_edge` using code like this: `Get_Implicit_Definition(Imp) = Iir_Predefined_Ieee_1164_Rising_Edge`.
You will certainly use a case statement to handle all the functions needed.

src/vhdl/vhdl-sem_lib.adb
```diff
@@ -128,6 +129,21 @@ package body Vhdl.Sem_Lib is
...
       Vhdl.Canon.Canonicalize (Unit);
 
+      -- Synthesis checks
+      --------------------
+
+      -- Note: Calling Synthesis checks after Canon is more computationally
+      --       demanding than before. This is because, concurent assignments
+      --       has been converted to processes and are therefore checked too!
+      --       We stick with this solution because Canon pass fixes possibly
+      --       broken chain of concurent statements for us!
```
tgingold, line 139: It is not clear to me what the broken chains are.

src/vhdl/vhdl-synthesis_checks.adb
```diff
@@ -0,0 +1,165 @@
...
+            if Get_Named_Entity (Real_Iir) = Gold_Iir then
+               Match := True;
+               exit;
+            end if;
+            Next (Real_It);
+         end loop;
+
+         if Match = False then
+            Warning_Msg_Sem (Warnid_Synth_Sensitivity_Lists, +Proc,
+                             "Incomplete sensitivity list! Missing signal: " &
```
tgingold, line 76: You can use a dedicated mechanism to insert names: `... Missing signal: %i", +Gold_Iir);`

src/vhdl/vhdl-synthesis_checks.adb
```diff
@@ -0,0 +1,165 @@
...
+               exit;
+            end if;
+            Next (Gold_It);
+         end loop;
+
+         if Match = False then
+            Warning_Msg_Sem (Warnid_Synth_Sensitivity_Lists, +Proc,
+                             "Over-specified sensitivity list! " &
+                             "Redundant signal: " &
+                             Image_Name_Id (Get_Identifier (Real_Iir)));
```
tgingold, line 116: Here too.

src/vhdl/vhdl-synthesis_checks.adb
```diff
@@ -0,0 +1,165 @@
...
+      if Iir_Kinds_Library_Unit (Get_Kind (El)) = Iir_Kind_Architecture_Body
+      then
+         Stmt := Get_Concurrent_Statement_Chain (El);
+         while Stmt /= Null_Iir loop
+            Kind := Get_Kind(Stmt);
+
+            -- FIXME: We should to ignore processes if we have "pragma
+            --        synthesis off" defined! Such processes will not be
+            --        synthesized, therefore shall not be checked!
+            if Kind = Iir_Kind_Sensitized_Process_Statement then
```
tgingold, line 140: You can use `Get_Process_Origin` to check if the process is a real user process (in that case the origin is Null_Iir) or if it was the result of Canon.

src/vhdl/vhdl-synthesis_checks.adb
```diff
@@ -0,0 +1,165 @@
...
+               Gold_List := Get_Golden_Sensitivity_List (Stmt);
+               Real_List := Get_Sensitivity_List (Stmt);
+
+               -- Skip clocked processes when checking !
+               if Get_Is_Clocked_Process (Stmt) = False and
```
tgingold, line 145: I don't think the `Is_Clocked_Process` is really needed, but that's a discussion for later.

src/vhdl/vhdl-synthesis_checks.ads
```diff
@@ -0,0 +1,24 @@
...
+--
+--  You should have received a copy of the GNU General Public License
+--  along with this program.  If not, see <gnu.org/licenses>.
+with Vhdl.Nodes; use Vhdl.Nodes;
+
+package Vhdl.Synthesis_Checks is
+
+   -- Check for issues which cause simulation synthesis mismatches
+   --  - check incomplete and over-specified sensitivity lists
+   procedure DoSynthesisChecks(Unit: Iir_Design_Unit);
```
tgingold, line 22: We don't use straight CamelCase.  Please insert underscores!

tgingold (changes requested):
I think this is a very useful warning, and I hope the requested changes are not too much.

doc/using/InvokingGHDL.rst
Blebowski: Hi, this has been actually done, "Warnid_Synth_Sensitivity_Lists" has been added and by default it is enabled.

The idea is that `--check-synthesis` will enable all Synthesis checks and "Warnid_Synth_Sensitivity_Lists" will
be a sub-set of these checks. Currently, since Sensitivity list check is the only one, these two options are equal.
Later, I would like to add others.

src/vhdl/vhdl-canon.adb
Blebowski: I thought of that, and I though that user would most probably not do that so I went with the naive solution first.
I will modify it to use the "Get_Implicit_Defintion" approach.

src/vhdl/vhdl-sem_lib.adb
Blebowski: I thought that link lists of sequential statements could be "broken", having no links between individual statements  (I saw a coment somewhere in the code from which I understood that canon pass would handle this situation). Clearly, I got it wrong and no such thing could occur.

src/vhdl/vhdl-synthesis_checks.adb
Blebowski: Ok, will do.

src/vhdl/vhdl-synthesis_checks.adb
Blebowski: Ok,

src/vhdl/vhdl-synthesis_checks.adb
Blebowski: Thanks, I did not know how to avoid running it on "cannoned" processes from parallel assignments.

src/vhdl/vhdl-synthesis_checks.ads
Blebowski: Ok

src/vhdl/vhdl-synthesis_checks.adb
Blebowski: Probably no if the detection of "golden" sensitivity list works also for clocked processes. So far I went with simple flag indicating this and just ignoring them. Later, I would like to have check of clocked processes too. This is especially important if you have reset of a flop as OR of multiple reset signals (altough it is not a good RTL practice), this typically leads to forgetting one of the reset signals in sensitivity list.

doc/using/InvokingGHDL.rst
tgingold: I am fine with multiple flags.  But still I think all command line flags for warnings must start with `-W` to easily identify them as warnings.  So what about `-Wsynthesis-sensitivity` for this issue and `-Wsynthesis` to enable all warnings about synthesis ?

Blebowski: Hi, thanks. It sounds reasonable. Plus, if the nested options will have the same prefix as the "master" option, it will be clear that these belong together.

Issue #1063, opened 2020-01-01
[Compliance-Tests] vhdl_2008/tb_matching_operator crash during analysis


eine:
Source file https://github.com/VHDL/Compliance-Tests/blob/master/vhdl_2008/tb_matching_operator.vhd contains multiple tests for VHDL 2008 "matching operator" features. Unfortunately, none of them can be tested because GHDL crashes.

```vhdl :file:test.vhd
library vunit_lib;
context vunit_lib.vunit_context;

library ieee;
use ieee.std_logic_1164.all;

entity tb_matching_operators is
  generic (runner_cfg : runner_cfg_t);
end entity tb_matching_operators;

architecture tb of tb_matching_operators is
begin
  test_runner: process is
  begin
    test_runner_setup(runner, runner_cfg);
    check_equal('1' ?= '1', '1');
    test_runner_cleanup(runner);
    wait;
  end process test_runner;
end architecture tb;

```

```py :run:
#!/usr/bin/env python3

from os.path import join, dirname
from vunit import VUnit, VUnitCLI

root = dirname(__file__)

ui = VUnit.from_argv(vhdl_standard="2008")
ui.add_library("vhdl_2008").add_source_files(join(root, "*.vhd"))

ui.main()

#:image: ghdl/vunit:llvm
```

```
=== Command used: ===
/usr/local/bin/ghdl -a --workdir=/root/vunit_out/ghdl/libraries/vhdl_2008 --work=vhdl_2008 --std=08 -P/root/vunit_out/ghdl/libraries/vunit_lib -P/root/vunit_out/ghdl/libraries/vhdl_2008 /root/test.vhd

=== Command output: ===

******************** GHDL Bug occurred ***************************
Please report this bug on https://github.com/ghdl/ghdl/issues
GHDL release: 0.37-dev (tarball) [Dunoon edition]
Compiled with GNAT Version: 7.4.0
Target: x86_64-linux-gnu
/root/
Command line:
/usr/local/bin/ghdl1-llvm --workdir=/root/vunit_out/ghdl/libraries/vhdl_2008 --work=vhdl_2008 --std=08 -P/root/vunit_out/ghdl/libraries/vunit_lib/ -P/root/vunit_out/ghdl/libraries/vhdl_2008/ -P/usr/local/lib/ghdl/ieee/v08/ -P/usr/local/lib/ghdl/ -c -fpic -o /root/vunit_out/ghdl/libraries/vhdl_2008/test.o /root/test.vhd
[...]

PR #1872, opened 2021-09-21
Implement Matching Operators

bpadalino:
Implement the matching operators ?<, ?<=, ?>, ?>=, ?/=.

Added implementations for the matching operators according to the LRM.

Tested against the Compliance Tests and appears to pass.

### DIFF
--- a/src/vhdl/vhdl-evaluation.adb
+++ b/src/vhdl/vhdl-evaluation.adb
@@ -1556,6 +1556,124 @@ package body Vhdl.Evaluation is
       end if;
    end Eval_Logic_Match_Equality;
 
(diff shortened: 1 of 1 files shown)


Paebbels:
What about `?=` and `case?`?

umarcor:
`?=` is supported already: https://github.com/VHDL/Compliance-Tests/runs/3669209802?check_suite_focus=true#step:3:2355 (#1063)

tgingold:
Thank you.  There are minor style deviation but I will adjust that later.

Many evaluation of operations with static operands are not yet implemented.  Most of them are rarely used. 
With vhdl 2008, many operations defined in the ieee packages can now be static and I am wondering how to merge the evaluation with synthesis (which does the same).

PR #1873, opened 2021-09-22
Add test for matching operators.

bpadalino:
Add an exhaustive test for the matching operators.

Added a test for the matching operators that were recently added.

Note that currently this test fails, but I am not sure where in the code the incorrect value is coming from.

The tables failing are `?>` and `?>=` with the either L or R being `'-'`.

Here is the entire [output](https://github.com/ghdl/ghdl/files/7211615/output.txt) of the test, when run with severity `error` instead of `failure`.

### DIFF
--- a/testsuite/gna/issue1872/match_operators.vhdl
+++ b/testsuite/gna/issue1872/match_operators.vhdl
@@ -1,0 +1,168 @@
+library ieee ;
+    use ieee.std_logic_1164.all ;
+
+library std ;
+    use std.textio.all ;
+
+entity match_operators is
+end entity match_operators ;
+
+architecture arch of match_operators is
+
+    -- Index is (L,R)
+    type result_table_t is array(std_ulogic, std_ulogic) of std_ulogic ;
+
+    function "not"(x : result_table_t) return result_table_t is
+        variable rv : result_table_t := (others =>(others =>'-'));
+    begin
+        for l in x'range(1) loop
+            for r in x'range(2) loop
+                rv(l, r) := not x(l,r) ;
+            end loop ;
+        end loop ;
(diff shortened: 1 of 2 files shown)


testsuite/gna/issue1872/match_operators.vhdl
```diff
@@ -0,0 +1,185 @@
...
+
+entity match_operators is
+end entity match_operators ;
+
+architecture arch of match_operators is
+
+    constant NUM_VALUES : integer := std_ulogic'pos(std_ulogic'right) - std_ulogic'pos(std_ulogic'left) ;
+
+    -- Index is (L,R)
+    type result_table_t is array(0 to NUM_VALUES, 0 to NUM_VALUES) of std_ulogic ;
```
tgingold, line 15: This is a little bit overcomplicated.  It is possible to use any discrete type as an index.  So you could write:
``` type result_table_t is array (std_ulogic, std_ulogic) of std_ulogic```

File: testsuite/gna/issue1872/match_operators.vhdl [hunk 2]
```diff
@@ -0,0 +1,185 @@
...
+
+    function "not"(x : result_table_t) return result_table_t is
+        variable rv : result_table_t := (others =>(others =>'-'));
+    begin
+        for l in 0 to NUM_VALUES loop
+            for r in 0 to NUM_VALUES loop
```
tgingold, line 21: And then iterate using either `for l in std_ulogic loop` or even better: `for l in x'range(1) loop`

bpadalino:
Rebased the test against master which has #1874, so the test now fully passes.

tgingold:
Thanks!

PR #1874, opened 2021-09-22
Add explicit ?>= and ?> functions for translation.

bpadalino:
Instead of swapping L/R arguments to try to create `?>=` and `?>`,
create a function for each which performs the not operation of `?<`
and `?<=` as defined by the LRM.

This causes the match_operators test from #1873 to pass.

### DIFF
--- a/src/ghdldrv/ghdlrun.adb
+++ b/src/ghdldrv/ghdlrun.adb
@@ -690,6 +690,10 @@ package body Ghdlrun is
            Grt.Std_Logic_1164.Ghdl_Std_Ulogic_Match_Lt'Address);
       Def (Trans_Decls.Ghdl_Std_Ulogic_Match_Le,
            Grt.Std_Logic_1164.Ghdl_Std_Ulogic_Match_Le'Address);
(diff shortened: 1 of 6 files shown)


src/ghdldrv/ghdlrun.adb
```diff
@@ -690,6 +690,10 @@ package body Ghdlrun is
            Grt.Std_Logic_1164.Ghdl_Std_Ulogic_Match_Lt'Address);
       Def (Trans_Decls.Ghdl_Std_Ulogic_Match_Le,
            Grt.Std_Logic_1164.Ghdl_Std_Ulogic_Match_Le'Address);
+      Def (Trans_Decls.Ghdl_Std_Ulogic_Match_Ge,
+           Grt.Std_Logic_1164.Ghdl_Std_Ulogic_Match_Ge'Address);
+      Def (Trans_Decls.Ghdl_Std_Ulogic_Match_Gt,
+           Get.Std_Logic_1164.Ghdl_Std_Ulogic_Match_Gt'Address);
```
tgingold, line 696: Typo: `Grt.` instead of `Get.`

tgingold:
Thanks!

PR #1879, opened 2021-09-23
Add parsing of case? statement and simple test.

bpadalino:
Also add the Matching flag to the Iir_Kind_Case_Statement.

**Description**
This adds the parsing for matching case statements (case?) as well as adds the Matching flag for the AST node.


src/vhdl/vhdl-parse.adb
```diff
@@ -7624,15 +7624,15 @@ package body Vhdl.Parse is
...
    --      [ CASE_label : ]
-   --          CASE expression IS
+   --          CASE [?] expression IS
    --              case_statement_alternative
    --              { case_statement_alternative }
-   --          END CASE [ CASE_label ] ;
+   --          END CASE [?] [ CASE_label ] ;
    --
-   --  [ LRM93 8.8 ]
+   --  [ LRM93 10.9]
```
umarcor, line 7635: Is `LRM93 10.9` correct?

bpadalino: Indeed it isn't.  Resolved.

testsuite/gna/issue717/matching_case.vhdl
```diff
@@ -0,0 +1,57 @@
...
+	port(
+		iClk : in std_ulogic;
+		inRstAsync : in std_ulogic;
+		iCtrl : in std_ulogic_vector (2 downto 0);
+		oData : out std_ulogic_vector(gWidth-1 downto 0)
+	);
+end entity CounterWithReset;
+
+architecture RTL of CounterWithReset is
+	signal CountStateM, CountStateNextM : std_ulogic_vector(oData'RANGE);
```
umarcor, line 16: Maybe `CountStateM` can be of type unsigned, to avoid all the castings below?

pyGHDL/libghdl/vhdl/nodes.py
```diff
@@ -1096,6 +1096,7 @@ class Iir_Kinds:
     ]
 
 
+
```
umarcor, line 1099: Execute `./scripts/update_py_bindings.sh` right before committing your changes, in order to have Python sources regenerated. Then, run `black pyGHDL` to have consistent formatting. Black is a Python package, which you can install through pip.

testsuite/gna/issue717/matching_case.vhdl
bpadalino: Changed.  I mostly just took the example from 717.  There's definitely a lot that this doesn't test.  Just very basic.

pyGHDL/libghdl/vhdl/nodes.py
bpadalino: OK.  I believe I re-ran those and updated.  Let's see.

umarcor: It seems that you setup keeps removing some lines in `pyGHDL/libghdl/vhdl/nodes.py`. Can you check that black is up to date (CI uses `black-21.9b0`) and that no other editor/IDE settings are modifying the sources?

bpadalino: OK, I'll keep this unresolved until we figure it out.

I think this latest push might have fixed?  Let's wait for the CI to tell us.

umarcor: Yes, it did! Thanks!

umarcor (approved):
LGTM! I'll let @tgingold merge, though.

tgingold:
Thanks!

PR #1885, opened 2021-10-03
Upstreaming patches from Debian packaging

andreasbombe:
The Debian packaging uses a few patches against upstream to make packages build correctly and according to policy. A bunch of these are not Debian specific and should have been upstreamed a while ago, and here they are now. These are just a few lines of changes mainly in the Makefile.in.

The CURDIR change actually fixed a build failure under certain conditions, but I forgot under which conditions that happened and could not reproduce it anymore. Anyway, making sure not to use the PWD from the environment variables is probably a good idea since its value in the environment is not really dependable.

### DIFF
--- a/doc/ghdl.1
+++ b/doc/ghdl.1
@@ -94,10 +94,8 @@ Dump VCD to FILENAME (a waveform dump, viewable with\-\-for instance\-\-\fBgtkwa
 Back annotate SDF onto design using TYPE (min|typ|max), instance PATH, and SDF file FILENAME, i.e. \fI./sdf_design \-\-sdf=typ=top/inst=inst.sdf\fP
 
 .SH "SEE ALSO"
-.TP
-.B gtkwave (1)
+.BR gtkwave (1)
 .PP
-.br 
 The texinfo manual fully documents GHDL. You may also browse it at
 \fB<https://ghdl.github.io/ghdl/>\fP.
 .SH "AUTHOR"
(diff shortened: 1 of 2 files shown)


tgingold:
I think this is OK.  But, do you really need to override `GHDL_VER_HASH` and `GHDL_VER_REF` ?

doc/ghdl.1
```diff
@@ -94,10 +94,8 @@ Dump VCD to FILENAME (a waveform dump, viewable with\-\-for instance\-\-\fBgtkwa
 Back annotate SDF onto design using TYPE (min|typ|max), instance PATH, and SDF file FILENAME, i.e. \fI./sdf_design \-\-sdf=typ=top/inst=inst.sdf\fP
 
 .SH "SEE ALSO"
-.TP
-.B gtkwave (1)
+.BR gtkwave (1)
```
umarcor, line 97: This file is actually auto-generated through Sphinx. It is included in the repo, so that packagers don't need to execute Sphinx explicitly, allowing them to package GHDL without using Python. However, the changes done here might be reverted at any time.

@andreasbombe can you afford to execute Sphinx for updating `ghdl.1` during packaging?

andreasbombe: I didn't realize that. Sphinx would be an additional build dependency, but then again ghdl is already heavy. Anyway, it was just a quick fix I had in addition to some changes for documenting Debian stuff.

andreasbombe: I didn't realize that. Sphinx would be an additional build dependency, but then again ghdl is already heavy. Anyway, it was just a quick fix I had in addition to some changes for documenting Debian stuff. I'll look into how to deal with this going forward.

PR #1901, opened 2021-10-25
synth: Support alias declarations in vunit

tmeissner:
This PR is intended to add support of `alias` definitions in PSL vunits.

I'm not sure if the handling of non-object aliases is correct, I assume such are aliases to procedures or entities? So I also added a simple testcase for that:

```vhdl
function incr (a : integer) return integer is
begin
  return a + 1;
end function incr;
```

```vhdl
vunit verif5 (assert2(behav)) {
 -- Alias to non-object
alias incr_a is incr[integer return integer];
...
assert always cnti_a = incr_a(prev(cnti_a));
}
```


tmeissner:
@tgingold Are the changes in this PR reasonable? If not I could save work and not bump this PR to the current HEAD of master 😉

tgingold:
Yes, it is OK.  I have resolved the conflicts and I am waiting for the checks.  Sorry for the delay.

tgingold:
Now merged.  Thanks!

PR #1928, opened 2021-12-05
Update .editorconfig with settings for Ada files

std-max:
**Description**
Update .editorconfig with settings for Ada files.
Indent style and indent size values are the one that are currently used by the GHDL project.

(This PR aims to prevent editor with .editorconfig plugin to force the default tab indentation)


umarcor (approved):

tgingold:
Thanks!

PR #1960, opened 2022-02-03
pyGHDL: reformat after update to black version 22

Xiretza:
Black was updated to version 22, now CI is failing because it has some new opinions on formatting.


Paebbels:
Quite opinionated for an *unopinionated* formatter ...

Issue #1958, opened 2022-02-02
Tracking issue: VHDL-2019 support


Xiretza:
Just like VHDL-2008, VHDL-2019 brings a lot of new features to the language. While some of them are quite out there and probably hard to implement, there are also a few small quality-of-life improvements.

This issue will serve both as a place for discussion about the implementation of VHDL-2019 features in general, as well as to track the implementation of all the Language Changes.

# [List of language changes](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/VHDL2017)
- [ ] [LCS-2016-001](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_001) "Partially Connected Vectors on Port Map"
- [ ] [LCS-2016-002](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_002) "Allow access and protected type parameters on function interfaces"
- [ ] [LCS-2016-004](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_004) "Allow PT methods with AT and FT parms"
- [ ] [LCS-2016-006a](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_006a) "File IO / Textio updates"
- [ ] [LCS-2016-006c](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_006c) "File IO / Directory I/O"
- [ ] [LCS-2016-006e](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_006e) "Allow access to system environment variables (baseline)"
- [ ] [LCS-2016-006f](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_006f) "Expose standard conditional analysis identifiers"
- [ ] [LCS-2016-007](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_007) "Sequential Declaration Regions"
- [ ] [LCS-2016-011](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_011) "Date/Time functions"
- [ ] [LCS-2016-012](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_012) "'IMAGE and TO_STRING for composite types"
- [ ] [LCS-2016-014](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_014) "Composites of Protected Types"
[...]

PR #1984, opened 2022-02-28
vhdl-sem_types: allow methods to return file and protected types in VHDL-2019

Xiretza:
This implements [LCS2016_004](http://www.eda-twiki.org/cgi-bin/view.cgi/P1076/LCS2016_004). It is required for the VHDL-2019 standard library to analyze.


Paebbels:
@Xiretza is this change all what's needed for that LCS?

If so:
* should we add test cases for each feature?
* can we check off this LCS in #1958 ?

Xiretza:
No, this is definitely not all that's required - for one, this doesn't even consider method parameters, only return types. It also might need changes outside the analyzer. This is just the bare minimum to get the standard library to work, so `--std=19` can actually be used and tested properly.

PR #2001, opened 2022-03-12
Fix hardcoded values in gcc backend's default_paths

DanielG:
The gcc backend's Make-lang.in still had hardcoded install paths. Since
this is copied to the gcc build system we don't have access to our
configure variables, so just treat it as a subst_file in the configure
script.

While we're at it we get rid of the separate sed invocation for
default_paths and have that go through config.status instead.


tgingold:
Thanks.

PR #2003, opened 2022-03-13
Parallelize suite_driver with xargs -P

DanielG:
The vests suite is still sequential since it doesn't seem to go through
suite_driver which is still limiting the speedup.

### DIFF
--- a/testsuite/suite_driver.sh
+++ b/testsuite/suite_driver.sh
@@ -9,31 +9,35 @@ ANSI_GREEN="\033[32m"
 ANSI_RED="\033[31m"
 ANSI_NOCOLOR="\033[0m"
 
-_suite="$1"
-shift
+parse_cmdline () {
+  _suite="$1"
+  shift
 
-# This is the only place where test dirs are specified.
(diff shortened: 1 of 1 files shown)


tgingold:
Why not but there is `gna/testsuite.py` which already run tests in parallel.
I would prefer to use the python script, it is easier to maintain.
WDYT ?

DanielG:
I don't speak python so to me it's certainly less maintainable ;)

If you really want something maintainable I'd go with a generic test driver framework like TAP then you don't have to home grow and maintain it to begin with. https://en.wikipedia.org/wiki/Test_Anything_Protocol

Paebbels:
I would prefer a testsuite run by `pytest`.

Besides pure execution, it would give us reports, statistics etc.

DanielG:
pytest looks to be in Debian so that works for me. I'd still like to get this merged in the mean time since it provides an immediate benefit at essentially no cost. I'm certainly not going to go and port the testsuite to some other framework ;)

tgingold:
Fine.

PR #2005, opened 2022-03-13
Fix include-dir paths returned by cmdline _again_

DanielG:
Looks like I managed to break --libghdl-include-dir again with 87df6ea6f5f49984884b6c5c336aad4e6b8dcdbe right after fixing it :) 

--libghdl-include-dir should point at a directory containing a ghdl/
subdirectory while --vpi-include-dir should point right at the ghdl/
subdir.

Note this reverses the appending of ghdl/ in the Makefile. If the explicit style is still preferred we could now go back on the quasi revert of 786ca17cea32431349608a37b48b608b0e60048f too.

If I wanted to add a test for the libghdl option in what suite would I put that? Or should I add a new "libghdl" directory?


tgingold:
If you want to do some tests about configure, do it directly in the CI flow.
(And just build and run a few tests).

DanielG:
Oh, no. Not the configure. I wan to test if something `gcc $(ghdl --libhgdl-include-dir)` will compile a trivial test program.

tgingold:
Ah, just add a test in gna/

DanielG:
Hmm, CI environment doesn't have a cc1plus :)

```
 + gcc -xc++ -I/usr/local/include -L/usr/local/lib test.c -lghdl-3_0_0_dev
  gcc: error trying to exec 'cc1plus': execvp: No such file or directory
```

tgingold:
Why do you need c++ (instead of c) ?

DanielG:
Because ghdl/synth.h is C++ :)

tgingold:
If you want to test, you can just include synth_gates.h

DanielG:
I would have liked to test the linking aspects too TBH. I guess I _could_ just manually depend on a mangled C++ symbol if you don't want to add a c++ compiler to the CI env though.

tgingold:
Don't worry, symbols are not mangled!

DanielG:
Yup, just realized that. This should work now.

testsuite/gna/issue2005/testsuite.sh
```diff
@@ -0,0 +1,16 @@
...
+. ../../testenv.sh
+
+set -x
+
+if c_compiler_is_available; then
+	lib=$("$GHDL" --libghdl-name)
+	lib=${lib#lib} # strip "lib" prefix
+	lib=${lib%.so} # strip .so suffix
+
+	$CC -I"$("$GHDL" --libghdl-include-dir)" -L"$(dirname "$("$GHDL" --libghdl-library-path)")" test.c -l$lib
```
tgingold, line 12: Could be much simpler: `$CC "-I$("$GHDL" --libghdl-include-dir)" test.c "$("$GHDL" --libghdl-library-path)"`

tgingold:
That's OK but we now have to fix the failures.

testsuite/gna/issue2005/testsuite.sh
DanielG: There's a reason I do it the long way around. This way I'm testing both --libghdl-library-path and --libghdl-library-name.

PR #2007, opened 2022-03-14
Fix include-dir paths returned by cmdline _again_

DanielG:
--libghdl-include-dir should point at a directory containing a ghdl/
subdirectory while --vpi-include-dir should point right at the ghdl/
subdir.

#2005 was closed by accident I think.


testsuite/gna/issue2005/testsuite.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if c_compiler_is_available; then
+	lib=$("$GHDL" --libghdl-name)
+	lib=${lib#lib} # strip "lib" prefix
+	lib=${lib%.so} # strip .so suffix
+	libdir="$(dirname "$("$GHDL" --libghdl-library-path)")"
+
+
+	$CC -I"$("$GHDL" --libghdl-include-dir)" test.c "$("$GHDL" --libghdl-library-path)"
+	$CC -I"$("$GHDL" --libghdl-include-dir)" -L$libdir test.c -l$lib
```
DanielG, line 15: I'm testing both ways now, straight up library path and searching --libghdl-name in the --libghdl-library-path dir.

tgingold:
ok.

umarcor:
> #2005 was closed by accident I think.

Agree. Probably Tristan clicked on "comment and close" instead of "comment" (the buttons are next to each other).

PR #2009, opened 2022-03-17
Makefile: install.local fixes

Xiretza:
The recent changes to the install path handling introduced some regressions in the in-tree testing workflow. These patches fix the `install.*.local` targets so they run correctly and at the right time.


tgingold:
Thanks!

Issue #2056, opened 2022-05-12
Compilation fails with GCC 12: pragma Unreferenced given for "Success" [enabled by default]


Xiretza:
**Description**
Compiling GHDL with GCC 12 fails with an error.

**Expected behaviour**
GHDL compiles.

**How to reproduce?**

```
./configure \
	--prefix=/usr/ \
	--enable-libghdl \
	--enable-synth

make
```

```
Use full IEEE library
Build machine is: x86_64-pc-linux-gnu
create pic/ subdirectory
Creating Makefile
Creating default_paths.ads
Creating ghdl.gpr
Creating scripts/gcc/Make-lang.in
for d in ieee/v87 ieee/v93 ieee/v08 std/v87 std/v93 std/v08 src/ieee src/ieee/v87 src/ieee/v93 src/ieee2008 src/std src/std/v87 src/std/v93 src/std/v08 src/synopsys src/synopsys/v08 src/upf src/vital95 src/vital2000; do \
  mkdir -p lib/ghdl/$d; \
done
Generate ortho_code-x86-flags.ads
Generate elf_arch.ads
Generate ghdlsynth_maybe.ads
gcc -c  -fPIC -DWITH_GNAT_RUN_TIME -g -Wall -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -g -fvar-tracking-assignments -ffile-prefix-map=/build/ghdl-mcode-git/src=/usr/src/debug -o jumps.o /build/ghdl-mcode-git/src/ghdl/./src/grt/config/jumps.c
/build/ghdl-mcode-git/src/ghdl/./src/grt/config/jumps.c:132:1: warning: ‘grt_overflow_handler’ defined but not used [-Wunused-function]
  132 | grt_overflow_handler (int signo, siginfo_t *info, void *ptr)
      | ^~~~~~~~~~~~~~~~~~~~
[...]

PR #2058, opened 2022-05-16
fix(configure): disable -Werror by default

Xiretza:
Because the build system does not have direct control over the compiler, it cannot ensure that no warnings are issued in downstream compilations. Such warnings can occur due to newer compiler versions with more sophisticated diagnostics (example: #2056, #2057), older compiler versions with diagnostics bugs, or simply different, untested compilers. With `-Werror` enabled by default, these harmless warnings result in complete compilation failures.

The option remains enabled in CI to ensure upstream code quality.


umarcor (approved):
LGTM!

tgingold:
Yes, I think this is a good compromise.

tgingold:
Thanks.

PR #2059, opened 2022-05-16
fix(testsuite): show diffs on comparison failure

Xiretza:
I just had a [spurious test failure](https://github.com/Xiretza/ghdl/runs/6456572177?check_suite_focus=true#step:7:798) in one of my branches (`gna/issue880` failed only in MINGW64 llvm), but unfortunately the only message was "report mismatch" because the output of the `diff` was swallowed. Replacing all occurrences of `diff.*-q` and `diff.*/dev/null` should make future test failures more debuggable.


umarcor (approved):
LGTM!

PR #2067, opened 2022-05-30
use --version flag to check for C compiler

sudden6:
**Description** Use the `--version` option over `-v` to check for existing compilers

`-v` does not only print the compiler version, but runs some commands according to `gcc --help` and `clang --help`. This has side effects, which breaks things on some setups. I also think that's what was originally intended.


tgingold:
Thanks!

PR #2087, opened 2022-06-09
Support non glibc Linux systems

Guiltybyte:
HAVE_BACKTRACE is now only defined as 1 on Linux systems if glibc is present, This is done by checking if the `__GLIBC__`
macro is defined, which it will be if glibc is on the system. This allows for successful building on non glibc systems (for example musl) which do not define a `backtrace()` function.

Tested build on Void Linux w/ musl & Ubuntu Linux w/ glibc


tgingold:
Thanks!

PR #2094, opened 2022-06-13
netlists-disp_verilog: Use blocking assignments in non-clocked blocks

antonblanchard:
When using `ghdl --synth --out=verilog` on Microwatt, I get a lot of verilator warnings of the sort:

```
%Warning-INITIALDLY: microwatt.v:112548:16: Delayed assignments (<=) in initial or final block
                                          : ... Suggest blocking assignments (=)
112548 |     ext_rst1_n <= 1'b0;
       |                ^~
```

The following patch fixes the warnings. I tried to find every assignment within an `always @*` process as well as any `initial` statements.


tgingold:
Thanks.

PR #2108, opened 2022-06-22
Fix docstring and a small bug in PoSh precompile script

### LINKED ISSUE #2107: Problems passing ghdl binary for PoSh precompile scripts / missleading documentation
**Description**

Hi.

I'm using the PowerShell precompile scripts to compile Intel and Lattice vendor libraries.

Prior to d890f27e480cdb37428f1b7c4bb37475ed38348d the environment variable `GHDL` as well as the `-GHDL` parameter had to contain the path to the GHDL
binary **directory** (e.g. `<myGHDLdir>/bin/`). Now, you have to pass the path to the GHDL **executable** (e.g. `<myGHDLdir>/bin/ghdl.exe`). 

The [docstring / help](https://github.com/ghdl/ghdl/blob/a4f663eb442310aa17d895d267a9e212c54c5efe/scripts/vendors/compile-intel.ps1#L84) tells me to pass the binary directory:

~~~
    -GHDL <String>
        Set GHDL binary directory.
~~~

but returns an error when I do:

~~~
.\compile-intel.ps1 -All -GHDL <mypath>\GHDL\MINGW32-mcode-standalone\bin
Use adv. options '-GHDL' to set the GHDL executable.
~~~

The related function for the PoSh scripts is [`Get-GHDLBinary`](https://github.com/ghdl/ghdl/blob/778ee5cbb8f7a936f8d4cc93ec6b4671bec2008e/scripts/vendors/shared.psm1#L137)

**Expected behaviour**

When the docs tell me to pass the binary directory, the scripts should work when I do so.

So I would suggest to fix the docstring, or even better to accept both, passing the directory or the executable.

If it is not wanted to make both variants work, I would like to update the error message for the PoSh scripts to something
like `This is not a path to a file. Please pass the path to the GHDL executable e.g. "<mypath>/GHDL/bin/ghdl.exe"`.

**How to reproduce?**
Set environment variable `GHDL` or pass `-GHDL` parameter containing a path to the binary directory to any PoSH
precompile script (e.g. `compile-intel.ps1`).

**Context**

- OS: Win10
- Origin:
  - [x] Released binaries: `[tarball_url](https://github.com/ghdl/ghdl/releases/download/nightly/MINGW32-mcode-standalone.zip)`

SalttownUser:
**Update**

As discussed [here](https://github.com/ghdl/ghdl/pull/2108#discussion_r906032565) and [here](https://github.com/ghdl/ghdl/issues/2107#issuecomment-1165541878) I changed the PR to only fix the docstring in the preocompile scripts and a small bug in `shared.psm1`.

**Description**

- 290b4d41ba91ee4fb4af91ba8d92108ea8f391ce fixes the try/catch when calling `Get-Command`. Before, the catch condition
   is never met what leads to this error message, when `ghdl..exe` is not on PATH:

   ```
   Get-Command: The term 'ghdl.exe' is not recognized as a name of a cmdlet, function, script file, or executable program.
   Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
   
   Test-Path: Value cannot be null. (Parameter 'The provided Path argument was null or an empty collection.')
   
   Use adv. options '-GHDL' to set the GHDL executable.
   ```

   You can see that the scripts exists because `Test-Path` is called with an undifined variable instead of hitting the catch condition.

- 81e290191fc019e86d7b6d1a4dd2b0808068c5df updates `Get-GHDLBinary` to accept both path to directory and path path to executable.
- db736b6cc763c4b69ab581e2ed9394c78288dc2c updates the docstring of the PoSh scripts

Closes #2107

I did split the commits in case you don't like changes from 81e290191fc019e86d7b6d1a4dd2b0808068c5df  but hopfully will accept the other commits. What do you think?

### DIFF
--- a/scripts/vendors/shared.psm1
+++ b/scripts/vendors/shared.psm1
@@ -156,16 +156,19 @@ function Get-GHDLBinary
 	{	$GHDLBinary = $env:GHDL   }
 	else
 	{	try
-		{	write-host "calling which ..."
-			$GHDLBinary = (Get-Command "ghdl.exe").Source }
+		{	Write-host "Calling Get-Command ..."
+			$GHDLBinary = (Get-Command "ghdl.exe" -ErrorAction Stop).Source }
 		catch
-		{	Write-Host "Use adv. options '-GHDL' to set the GHDL executable." -ForegroundColor Red
+		{	Write-Host "Cannot find ghdl.exe." -ForegroundColor Red
+			Write-Host "Use adv. options '-GHDL' to set the GHDL executable." -ForegroundColor Red
 			Exit-CompileScript -1
 		}
 	}
 
 	if (-not (Test-Path $GHDLBinary -PathType Leaf))
-	{	Write-Host "Use adv. options '-GHDL' to set the GHDL executable." -ForegroundColor Red
+	{
+		Write-Host "$GHDLBinary is not a file." -ForegroundColor Red
+		Write-Host "Use adv. options '-GHDL' to set the GHDL executable." -ForegroundColor Red
 		Exit-CompileScript -1
 	}
 
--- a/scripts/vendors/compile-altera.ps1
+++ b/scripts/vendors/compile-altera.ps1
@@ -81,7 +81,7 @@ param(
 	[string]$Source =     "",
 	# Set output directory name.
 	[string]$Output =     "",
-	# Set GHDL binary directory.
+	# Set path to GHDL's executable, e.g. <MyGHDLPath>/bin/ghdl.exe
 	[string]$GHDL =       ""
 )
 
--- a/scripts/vendors/compile-intel.ps1
+++ b/scripts/vendors/compile-intel.ps1
@@ -81,7 +81,7 @@ param(
 	[string]$Source =     "",
 	# Set output directory name.
 	[string]$Output =     "",
-	# Set GHDL binary directory.
+	# Set path to GHDL's executable, e.g. <MyGHDLPath>/bin/ghdl.exe
 	[string]$GHDL =       ""
 )
 
--- a/scripts/vendors/compile-lattice.ps1
+++ b/scripts/vendors/compile-lattice.ps1
@@ -94,7 +94,7 @@ param(
 	[string]$Source =     "",
 	# Set output directory name.
 	[string]$Output =     "",
-	# Set GHDL binary directory.
+	# Set path to GHDL's executable, e.g. <MyGHDLPath>/bin/ghdl.exe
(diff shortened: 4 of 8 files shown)


Paebbels:
@tgingold can you clarify what is / was / should be the meaning of `GHDL` in other places?

Is it like `GCC` or `GHDLLS` the path to the executable itself or the path to the directory of the binary?

There is also `GHDL_PREFIX` which points to `.../lib/ghdl`.

tgingold:
For me, GHDL is always the path or the name of the executable.
Then the prefix if needed can be extracted from it.

umarcor:
For context, clarifying and documenting GHDL, PREFIX and GHDL_PREFIX is listed as a pending task in #1557.

scripts/vendors/shared.psm1
```diff
@@ -150,21 +150,32 @@ function Get-GHDLBinary
...
-			$GHDLBinary = (Get-Command "ghdl.exe").Source }
-		catch
-		{	Write-Host "Use adv. options '-GHDL' to set the GHDL executable." -ForegroundColor Red
-			Exit-CompileScript -1
+	if ($GHDL -eq "")
+	{	$GHDL = $env:GHDL	}
+
+
+	if ($GHDL -ne "") {
+		if (Test-Path $GHDL -PathType Container) # is directory
```
umarcor, line 158: I think it is an interesting feature to support GHDL either being either a directory or a binary. However, that would introduce an inconsistency with the bash scripts and pyGHDL.

SalttownUser:
@Paebbels What is your opinion about the Codacy Style Check result. I don't think it makes sense to change all `Write-Host` in this PoSh script. I also don't like the idea of fixing/changing `Write-Host` only in the lines chaged by me as it is the default output for the script.

Paebbels:
@SalttownUser do you have a Gitter account for further (chat) discussions?

SalttownUser:
Not yet. But I signed in to Gitter with my Github account and joined ghdl a minute ago.

scripts/vendors/shared.psm1
SalttownUser: I have removed the feature and just fixed the docstring and the bug where `Get-Command` is called.

SalttownUser:
@Paebbels @umarcor  Hi. I removed commit 81e290191fc019e86d7b6d1a4dd2b0808068c5df. Is there something else to do?

tgingold:
For me it's OK, but I'd like to hear from @Paebbels or @umarcor before merging.

Paebbels (approved):

PR #2120, opened 2022-07-03
Added logging mechanism to pyGHDL

eng73:
This PR introduces a basic logger to replace various print statements. The intention is to allow other tools using pyGHDL to be able customise the log level/format/stream.

:rotating_light: Before submitting your PR, please read [contribute](http://ghdl.github.io/ghdl/contribute.html#fork-modify-and-pull-request) in the [Docs](http://ghdl.github.io/ghdl), and review the following checklist:

- [x] DO indicate which issues are either fixed or closed by this PR. See [GitHub Help: Closing issues using keywords](https://help.github.com/articles/closing-issues-via-commit-messages/).
No issues are being closed by this PR.

**Further comments**

:heart: Thank you!


Paebbels:
The printouts are only in `pyGHDL.dom`, so please move the logger instance to that package.  
`pyGHDL` itself has no outputs and will never have them.

As all messages are swallowed by the logger, a printout in `https://github.com/ghdl/ghdl/blob/master/pyGHDL/cli/dom.py` is missing.

Paebbels (changes requested):
The printouts are only in `pyGHDL.dom`, so please move the logger instance to that package.  
`pyGHDL` itself has no outputs and will never have them.

As all messages are swallowed by the logger, a printout in `https://github.com/ghdl/ghdl/blob/master/pyGHDL/cli/dom.py` is missing.

pyGHDL/cli/dom.py
```diff
@@ -316,7 +317,7 @@ def HandlePretty(self, args):
         for line in PP.formatDesign(self._design, 1):
             buffer.append(line)
 
-        print("\n".join(buffer))
+        logger.info("\n".join(buffer))
```
Paebbels, line 320: This output is not printed by logger.

pyGHDL/dom/Attribute.py
```diff
@@ -118,7 +118,7 @@ def parse(cls, attributeNode: Iir) -> "AttributeSpecification":
             if nameKind == nodes.Iir_Kind.Simple_Name:
                 names.append(SimpleName(name, GetNameOfNode(name)))
             elif nameKind == nodes.Iir_Kind.Signature:
-                print("[NOT IMPLEMENTED] Signature name in attribute specifications.")
+                logger.warning("[NOT IMPLEMENTED] Signature name in attribute specifications.")
```
Paebbels, line 121: Messages don't print in-order because of logger.

Paebbels (changes requested):
The given implementation change doesn't have the same behavior.

* Messages do not print in order  
  | Your PR | master branch |
  | --- | --- |
  | image | image |
* Messages from CLI are not shown at all.  
  | Your PR | master branch |
  | --- | --- |
  | image | image |



The purpose of `pyGHDL.cli.dom` is to test and demonstrate capabilities of `pyGHDL.dom`. So a printout of the translated structure is essential.

Also printing error messages at the point where they appear is also essential.

pyGHDL/cli/dom.py
eng73: As this is a test application, I propose reverting these back to print statements.

umarcor:
Is ordering still a problem after the last update to this PR?

Paebbels:
Yes it is.

I see 2 solutions to the problem:
1. we add a mode or similar to deactivate these printouts - note pyGHDL.dom is an alpha state extension - so it's not finished and fully ready for production, as features are missing.
2. we remove print outs from code and keep it on a dev-branch for the developer(s).

Currently, print out are use to denote what's not translated to pyVHDLModel. Throwing an exception is no option, because it's not meant to be an exception and it would interrupt the translation, so almost no input could be translated.

Should I investigate if we can add a flag like "development mode" or so?

umarcor:
I think it makes sense to have a "debug" mode, which can be enabled by either a CLI option or an environment variable. Maybe `GHDL_DEBUG=true` (so it can potentially be used by ghdl itself)? Or just `PYGHDL_DEBUG=true`?

Paebbels:
I'll need some time for this change.

umarcor:
No rush :wink:

Issue #2165, opened 2022-08-05
Change frelaxed-rules to accept subprogram bodies whose parameter lists differ to their declaration in presence of optional keywords


Farmadupe:
**Is your feature request related to a problem? Please describe.**
In a subprogram interface list, every parameter has an object class (constant, signal, variable, file) This object class can be explicitly written in the subprogram specification, but for every parameter there is a default (constant for mode in, variable for other modes). For example, the following two subprogram declarations are semantically identical:
```vhdl
procedure p(arg : bit);
procedure p(constant arg : bit);
```

VHDL allows forward declaration of functions. When there is a forward declaration, the 2008 LRM requires the interface lists of subprogram declaration and body to match lexically according to the following wording:
```
4.10 Conformance rules
Whenever the language rules either require or allow the specification of a given subprogram to be provided
in more than one place, the following variations are allowed at each place:
— A numeric literal can be replaced by a different numeric literal if and only if both have the same
value.
— A simple name can be replaced by an expanded name in which this simple name is the suffix if, and
only if, at both places the meaning of the simple name is given by the same declaration.
Two subprogram specifications are said to lexically conform if, apart from comments and the preceding
allowed variations, both specifications are formed by the same sequence of lexical elements and if
corresponding lexical elements are given the same meaning by the visibility rules.
```

GHDL follows this wording exactly, and hence raises an error when a subprogram declaration and a subprogram body mismatch but are semantically the same. For example on the following snippet:

```vhdl
entity e is end; 
architecture a of e is

    -- Body mismatches declaration because 'arg' is not marked constant. 
    -- However constant is the default value when not specified
    function f1(constant arg : bit) return bit;
    function f1(         arg : bit) return bit
    is
    begin
        return '0';
    end;
begin 
end;

-------------------------------------------------------------------------------------

> ghdl -a -frelaxed-rules bug.vhd
bug.vhd:7:14: body of function "f1" does not conform with specification at line 6:14
    function f1(         arg : bit) return bit
```

While GHDL is totally correct, other simulators do raise an error in this circumstance. This means that code that is considered "de facto VHDL " because it compiles in other simulators cannot be used in GHDL without modification.

**Describe the solution you'd like**
Request that the -frelaxed-rules option of GHDL be modified to raise no error upon compilation. This would prevent having to make trivial modifications to source files to allow them to work with GHDL.

[...]

PR #2167, opened 2022-08-07
Document changes to -frelaxed from #2165

Farmadupe:
**Description** 
Document how -frelaxed treats conformance rules for subprograms after #2165 

**Note**
I have not performed the below activity because I do not know how to build the docs.
```
```


tgingold:
Thanks!

Issue #2036, opened 2022-04-13
ghdl-ls crashes on windows when hdl-prj.json is used.


Farmadupe:
**Description**
When using ghdl-ls to analyze, the server executable crashes when analyzing any source file with more than one design unit.

**How to reproduce?**
This minimal case elicits the error for me:

**pkg_1.vhd**
```vhd :file: pkg_1.vhd
package a is end;
package b is end;
```

**hdl-prj.json**
```json :file: hdl-prj.json
{"files": [{ "file": "pkg_1.vhd", "language": "vhdl" }]}
```

Representative stack traces (line numbers may differ because I annotated ghdl-ls python with some print statements):
```
Traceback (most recent call last):
  File "C:\Users\t\AppData\Local\Programs\Python\Python310\lib\site-packages\pyGHDL\lsp\lsp.py", line 122, in handle
    response = fmethod(**params)
  File "C:\Users\t\AppData\Local\Programs\Python\Python310\lib\site-packages\pyGHDL\lsp\vhdl_ls.py", line 112, in textDocument_documentSymbol
    doc = self.workspace.get_or_create_document(textDocument["uri"])
  File "C:\Users\t\AppData\Local\Programs\Python\Python310\lib\site-packages\pyGHDL\lsp\workspace.py", line 107, in get_or_create_document
    res = self.create_document_from_uri(doc_uri)
  File "C:\Users\t\AppData\Local\Programs\Python\Python310\lib\site-packages\pyGHDL\lsp\workspace.py", line 100, in create_document_from_uri
    sfe = document.Document.load(source, os.path.dirname(path), os.path.basename(path))
  File "C:\Users\t\AppData\Local\Programs\Python\Python310\lib\site-packages\pyGHDL\lsp\document.py", line 51, in load
    sfe = files_map.Reserve_Source_File(dirid, fileid, buf_len)
  File "C:\Users\t\AppData\Local\Programs\Python\Python310\lib\site-packages\pyGHDL\libghdl\_decorator.py", line 192, in inner
    raise LibGHDLException(
pyGHDL.libghdl.LibGHDLException: Caught exception when calling 'files_map__reserve_source_file' in libghdl.


Traceback (most recent call last):
  File "C:\Users\t\AppData\Local\Programs\Python\Python310\lib\site-packages\pyGHDL\libghdl\_decorator.py", line 189, in inner
    return functionPointer(*args)
OSError: [WinError 541541187] Windows Error 0x20474343

```

[...]

PR #2168, opened 2022-08-07
Resolve crashes caused by inconsistent capitalization on Windows (fix issue #2036)

Farmadupe:
Currently ghdl-ls crashes inside libghdl on windows when hdl-prj.json is used due to capitalization issues (issue #2036). This commit updates ghdl-ls to intercept all sources of filenames (hdl_prj.json and language server RPCs) and normalizes them before they are passed to libghdl.

ghdl-ls now appears to work without crashing on Windows, whether or not hdl_prj.json is used.

I have not yet tested the change on linux, but hopefully it should be unaffected, because all changes are protected by runtime guards that detect the OS.

I am not used to python so my code may be unpythonic.


Paebbels (changes requested):
Please use `pathlib` instead of old `os.path` to solve this problem.

Farmadupe:
Changed lsp.py to use Pathlib  to remove quotes.

Also updated function path_to_uri to replace handwritten code with equivalent stdlib functionality.

Now passes testsuite locally on linux. I cannot run the testsuite locally on windows due to temporary libghdl.dll incompatibility (I don't build libghdl.dll myself).

tgingold:
I think the recursion in `normalize_rpc_file_uris` might be too heavy, but that's ok.
So a client doesn't need to normalize the path.  Interesting...

PR #2195, opened 2022-09-23
pyGHDL: added missing type annotations. Fix #2192

### LINKED ISSUE #2192: Exception when importing pyGHDL.libghdl.vhdl.scanner
**Description**
A Python import statement of the form `import pyGHDL.libghdl.vhdl.scanner` fails with the Exception `ValueError: Function Get_Token_Offset is not annotated with types.`

**Expected behaviour**
The import should work without errors.

**How to reproduce?**
Just run the import statement. 
As I see it, the functions `Get_Token_Position` and `Get_Position` are missing type annotations for their return values. When I add `-> int` everything works as expected. 

**Context**

- OS: Fedora 36
- Origin: Built from sources: `5914540da`

fhuemer:
Added return type annotations to the functions `Get_Token_Position` and `Get_Position` in `scanner.py` to fix #2192

### DIFF
--- a/pyGHDL/libghdl/vhdl/scanner.py
+++ b/pyGHDL/libghdl/vhdl/scanner.py
@@ -97,7 +97,7 @@ def Get_Token_Offset() -> int:
 
 @export
 @BindToLibGHDL("vhdl__scanner__get_token_position")
-def Get_Token_Position():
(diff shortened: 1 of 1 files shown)


tgingold:
Thanks!

PR #2241, opened 2022-11-07
Escape port name in dot output.

cderrien:
This escapes port names in dot output to prevent any signal name starting with the character 'r' from being escaped.
Example : "\rstn" -> "CR stn"
This will also prevent \n and \t from being escaped.

This does not fix an issue.


tgingold:
Thanks!

PR #2242, opened 2022-11-07
Added id to warnings related to attributes.

cderrien:
This adds a warning id to two warnings related to attributes encountered during synthesis.
It is meant to be useful if enum_encoding is used, so that the warnings can be ignored.

This does not fix any issue.


tgingold:
Thanks!

PR #2243, opened 2022-11-07
Restore latch inference.

cderrien:
Do not merge this ! This is not a real pull request.

I was wondering why yosys was freaking out on any hdl with latches in it when I saw this.
Mostly I am wondering why this is here. I applied the patch locally and tried synthesising a few designs with latches.
So far, it seems to work all right.

Is there a reason this is disabled that I am not seeing ?

### DIFF
--- a/src/synth/netlists-inference.adb
+++ b/src/synth/netlists-inference.adb
@@ -914,11 +914,7 @@ package body Netlists.Inference is
            (Loc, "latch infered for net %n (use --latches)", (1 => +Name));
       end if;
 
-      if False then
-         return Infere_Latch_Create (Ctxt, Val, Prev_Val, Last_Mux, Loc);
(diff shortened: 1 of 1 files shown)


tgingold:
Probably this is correct.  However it would be nice to have tests!

cderrien:
What sort of tests do you want ?
I have tried testing this with all examples provided in previous issues concerning latches (it worked; I could copy them here), but these are all minimal tests. 
It would be hard to provide tests for false negatives (ghdl synthesises a combinatorial loop instead of a latch) or false positives (ghdl synthesises a flip flop instead of a latch or combinatorial loop).

tgingold:
Yes, you're right; there are already tests for latches.

tgingold:
Thanks!

PR #2255, opened 2022-11-28
f-strings

Paebbels:
This PR converts almost all `"".format(...)` calls into f-string syntax.

The changes apply to pyGHDL.DOM, pyGHDL.libghdl, pyGHDL.lsp, pnodes, some unit tests, but **not** to provided testcases having Python code shipped with them.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -74,7 +74,7 @@ try:
     with open(prologPath, "r") as prologFile:
         rst_prolog = prologFile.read()
 except Exception as ex:
-    print("[ERROR:] While reading '{0!s}'.".format(prologPath))
+    print(f"[ERROR:] While reading '{prologPath!s}'.")
     print(ex)
     rst_prolog = ""
 
--- a/pyGHDL/cli/dom.py
+++ b/pyGHDL/cli/dom.py
@@ -224,7 +224,7 @@ class Application(LineTerminal, ArgParseMixin):
             try:
                 self.SubParsers[args.Command].print_help()
             except KeyError:
-                self.WriteError("Command {0} is unknown.".format(args.Command))
+                self.WriteError(f"Command {args.Command} is unknown.")
 
         self.WriteNormal("")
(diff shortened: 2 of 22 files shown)


Paebbels:
macOS is failing in CI, because it runs with Python 3.11 instead of 3.10.

tgingold:
Yes, but why is Python 3.11 a problem ?

Paebbels:
I'm already investigating the 3.11 problem. It comes from pyTooling and is related to some low-level code. It looks that changes in 3.11 had an effect on pyTooling.

For my other packages, I didn't upgrade all of them to 3.11 to buy some time. But here it creates more presure on me to find a solution, because macOS already uses it.

Paebbels:
I fixed it. Sorry, there was an overlap between multiple macOS and Python 3.11 issues. Here the clarifications:
* In pyTooling Python 3.11 gives problems on Windows, due to no wheel file for `lxml` that either compiles on Windows or ships with the wheel.
* In pyTooling, macOS jobs where failing due to a missing program on UNIX compared to Linux. The computation of relative paths is now solved in a different way, thus macOS jobs are completing.
* In GHDL, macOS jobs where failing because there is a new additional attribute in enumerations. This new attribute breaks the decorator `EnumLookupTable`. I added another check to exclude low-case attributes in addition to the already excluded underscore attributes. This assumes that GHDL always uses upper-case enumeration literals.  
  I could reproduce locally on my Windows too. It was just visible on macOS here, because it was the first job switching to Python 3.11 in GitHub Actions.

This decorator is used to precompute and auto-generate lookup tables for enumerations from GHDL like:
```python
@export
@EnumLookupTable(nodes.Iir_Kind)
def kind_image(k: int) -> str:
    """String representation of Nodes.Iir_Kind :obj:`k`."""
```

The remaining `macOS - llvm` is due to synthesis but independent of f-strings, right?

tgingold:
Thanks!

PR #2256, opened 2022-11-28
Updaing used GitHub Actions

Paebbels:
This PR updates used GitHub Actions in the workflow. It should reduced the number of warnings issued by GHA about node.js 12.

It also enables dependabot for dependency scanning and creating PRs with dependency updates (if PR gets accepted).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,31 @@
+version: 2
+updates:
+  # Maintain Python packages
+  - package-ecosystem: "pip"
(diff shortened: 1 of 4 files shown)


Paebbels:
Job `pyGHDL` fails due to a push of an outdated `pyGHDL/libghdl/errorout.py` file. This will be fixed by #2255 (and by a cherry-picked commit d8365b9f23dabbb309d2b2310fcfe454faf07afa embedded in this PR).

PR #2275, opened 2022-12-19
Testing complex documentation examples

Paebbels:
This PR shows 2 problems:
1. in the single line complex example, an assertion error is raised, because a normal comment is gathered as documentation comment. Thus it has no marker.  
   > no :ctx1: in [349..443] '-- package body should be supported too to keep parity, but I have currently no usecase for it.'(95)
2. it also contains the same complex example using multi-line comments. Here a fatal windows error is thrown. I assume it's an internal exception in libghdl.  
   See line 303 in the `Comments.py`.

I think I can't further debug from my side. Feel free to take over my branch and finish the merge.


testsuite/pyunit/libghdl/examples/Complex_MultiLine.vhdl
```diff
@@ -0,0 +1,138 @@
...
+	 */
+	generic (
+		/* :FREQUENCY: comment before a generic
+		 * :FREQUENCY: might be multiline
+		 */
+		constant FREQUENCY : positive;
+		constant BITS      : positive; /* :BITS: comment after a generic are mostly single line,
+																		* :BITS: but could be multi line too
+																		*/
+		/* in case comment is before and after
```
Paebbels, line 52: Is this missing `*/` the cause for the internal error?

I keep it for now, so you can improve libghdl with a detection of the condition and emiting an error message before I fix the syntax.

Paebbels: After adding the missing `*/`, there are 2 observations:
1. It fixes the Windows error.
2. The mentioned wrong association for the single line complex example, does not exist for complex multi line.

(Or: are there no multi-line comments gathered, so nothing is compared -> thus no errors at all?)

tgingold:
For 1, I think the test must be changed.  Comments are simply gathered and attached to nodes.  I don't think we can differentiate normal and documentation comments without a marker.

For 2, I am investigating.

Paebbels:
For 1:
I think it can be solved.  
See this example where it distinguishes already 2 comments and associates them with architecture and a declared object.

```vhdl
architecture a of e is
  -- inner comment for architecture.
  
  -- comment for a signal
  signal s : bit;
begin

end architecture;
```

Similar it can be solved for this situation:

```vhdl
architecture a of e is

  -- comment for a constant
  constant c : boolean := true;
  
  -- just a comment, no documentation

  -- comment for a signal
  signal s : bit;
begin

end architecture;
```

I would summarize these rules:
* Multiline documentation comments must be consecutive.  
  Thus, if a comment with regex `--.*?\n` (comment with any text followed by a line break) is followed be an empty line of regexp `\s*?\n` (whitespaces ended by a line break), then multiline documentation comments are separated comment blocks.
* If documentation comments are written before VHDL entities, then no empty line is allowed before.  
  ```vhdl
  -- not associated to architecture
  
  architecture a of e is
  begin
  end architecture;
  ```
* If documentation comments are written inside VHDL entities, then no empty line is allowed between entity and first documentation comment line.  
  ```vhdl
  architecture a of e is

    -- not associated to architecture
  begin
  end architecture;
  ```
* If a documentation comment is written after a VHDL entity, its first documentation comment line must be written in the same line following this regexp: `;\s*?--.*?\n`:
  ```vhdl
  constant c : bit := '1';  -- constant comment
  ```
* More documentation comment lines can follow an "after" comment line, if there is no empty line in between. Any whitespace for indentation is allow in front of the comment of further lines.  
  ```vhdl
  constant c : boolean := true;  -- constant comment
                                 -- on multiple lines
           -- wrong indentation doesn't matter

                                 -- not associated to the constant
  ```

tgingold:
I am not sure how well this will work.

What about the comment at the beginning of a file ?  It is usually followed by a newline.
Also big comments usually contain empty lines.

I think it is better to currently attach as many comments as possible.  They can be filtered later and we can tune also later if needed.

Paebbels:
Big documentation comments are not a problem. Other tools also enforce no empty source code line in between. It doesn't mean, there can't be an "empty comment".

```vhdl
architecture a of e is
  -- headline of a component description
  --
  -- description of a component
  -- further text
  --
  -- Parameters:
  -- * param 1
```

With rules as stated above, you can have big multi-line comments with separating lines, but a comment sign plus a comment with no content or whitespace content will be used.

Can you give an example for a comment at file start where there might be problems?

tgingold:
A lot of vhdl files start like this:
```vhdl
--  comment
--  copyright
-- ...
--  description
-- ..

library ieee;
...
```
I think there is no need to settle that now.  It isn't something difficult to deal with.

Paebbels:
For these "global" per file comments, I would define it as:
* Consecutive VHDL comments before any VHDL reserved word (like `library`, `use`, `context`, `entity`, `package`, etc.)

```vhdl
-- file description
-- 
-- license: MIT
-- 
-- more documentation text must be in the same comment block.

-- not part of documentation, just normal comment


library ieee;
```

Paebbels:
Is `2.` fixed / investigated or preserved, so can I push new code to this branch?

tgingold:
For those global comments, you also see headers like:
```vhdl
--  Copyright
-- ...

--  Description
-- ...

library ieee;
```

I would simply keep the current implementation as it is and see how it works in practice.
The additional filtering could also be done by the python code.

tgingold:
For 2. I haven't been able to reproduce the issue.  `Complex_Multiline` is now tested.

I think we should close this PR, as it contains too many conflicts.

Paebbels:
I rebased this PR. CI looks good so far, but some jobs are still running: https://github.com/ghdl/ghdl/actions/runs/3759992146

I marked questionable comments with `--!`, but added expected `:name:` marker to get the CI successful.

Paebbels:
CI is all green.

Paebbels:
P.S. I don't think multi line has same results as single line comments. In single line I marked the mentioned line from first comment with a `:name:` marker. In multi line that marker wasn't added, but it's successful. I suspect the tests are false positives.

tgingold:
Thanks.  And yes we need a failing test for block comments.  They aren't yet gathered I think.

PR #2105, opened 2022-06-21
pyGHDL updates

Paebbels:
# New Features
* Use VHDL library files (ieee2008, synopsys, mentor, vital2000) for additional pyGHDL.DOM testcases.
* Added `vhdlVersion` parameter to `Document` class, so VHDL-AMS can be enabled per source file.
* Added new libghdl C-bindings for:
  * `flags.AMS_Vhdl` which refers to the `--ams` CLI option.
  * `flags.MB_Comment` which refers to the `--mb_comment` CLI option.
  * `flags.Explicit` which refers to the `-fexplicit` CLI option.
  * `flags.Relaxed` which refers to the `-frelaxed` CLI option.

# Changes
* Changed `"...".format(...)` to f-strings.
* Changed default encoding used by pyGHDL.libghdl and pyGHDL.DOM from `utf-8` to `latin-1`.  
  This was required by VHDL library source files. The global encoding setting is reachable via `pyGHDL.libghdl.ENCODING` and can be monkey-patched.
* Updated autoapi template.
* Updated dependency to pyVHDLModel to v0.15.0, which fixes error messages created by GHDL's Sphinx documentation build.

# Bug Fixes
* After reading error messages from the error message buffer. A call to `Clear_Errors` was missing.
* Added missing typehints.

PR #2284, opened 2022-12-24
New wrapper for flags

Paebbels:
This PR proposes a solution to get rid of `flags_name.value = True` syntax and uses proper descriptor protocols for hiding the C-binding wrapper.

**Old:**
```py
Flags_Gather_Comments.value = True
```

**New:**
```py
Flags().Gather_Comments = True
```

@tgingold if you like the idea, I would do these additional steps:
* Eliminate the parenthesis (instantiating a singleton):
  ```py
  Flags.Gather_Comments = True
  ```
* Expand this technique to other flags too. E.g. in parse, vhdl, ...

------
Originally started PR was #2105.


tgingold:
I'd really prefer to have a layered approach.  libghdl should be just a ctype interface, and a more pythonic interface would be implemented on top of it.

Could you at least keep the wrappers as globals in flags.py ?  That would also keep backward compatibility.
Than I am OK for the Flags class.

Paebbels:
With my proposal, I would like to enhance what ctypes doesn't do well. They have implemented a descriptor interface, but introduced another level of value access by using `.value` syntax.

> Could you at least keep the wrappers as globals in flags.py

I didn't fully get this point.  
Do you mean:
* keeping the old ctypes wrappers in parallel to the Flags class, OR
* providing aliases to the descriptors, so old names still work?

My goal is to provide a clean and easy syntax for bare variables from libghdl, which ctype doesn't provide. No further overhead or abstraction is planned on pyGHDL.libghdl.

I'll do further improvements like getting rid of the `()` and then you can see a more final proposal.

--------------
In an ideal world, I wouldn't implement this using a singletone + (instance) properties / descriptors. I would prefer to use class properties (like @property + @classmethod). Unfortunately, this feature is broken in Python and it's even a vulnerability. I'm trying to push on fixing it in Python itself. If so, it won't be available before 3.12 as I don't expect back porting.

tgingold:
> With my proposal, I would like to enhance what ctypes doesn't do well. They have implemented a descriptor interface, but introduced another level of value access by using .value syntax.

I think I perfectly understand your proposal and I think this is a good idea.

> Could you at least keep the wrappers as globals in flags.py
> 
> I didn't fully get this point.
> Do you mean:
> 
> 	• keeping the old ctypes wrappers in parallel to the Flags class, OR
> 	• providing aliases to the descriptors, so old names still work?

I would simply keep the wrappers as they are and just add the Flags() class.  No need to move the wrappers inside the class.

And yes, I agree that getting rid of the () would also be nice!

Paebbels:
I can rebase, but there is additional work needed from my side.

PR #2293, opened 2022-12-31
Limit Sphinx to 5.3 <= XX < 6.0

Paebbels:
This should delay the problem with Sphinx until the RTD theme is fixed with respect to docutils 0.18/0.19.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,6 +1,6 @@
 -r ../pyGHDL/requirements.txt
 
-sphinx>=5.0.2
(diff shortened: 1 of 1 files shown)


tgingold:
Thanks.

Paebbels:
@umarcor  
I think when this https://github.com/buildthedocs/sphinx.theme/issues/7 gets updated, then we should be fine again for higher Sphinx versions.

If If Sphinx and docutils are newer releases, the navigation bar is broken. It doesn't show sub sections in the navigation with `[+]`.

PR #2305, opened 2023-01-09
Dot output ports

cderrien:
This fixes the output ports being absent from the dot output of synthesis, and removes the dummy instance that was present.
There is no relevant issue.


tgingold:
Thanks!

Issue #2314, opened 2023-01-13
Standalone installation targets in makefile for GHDL 'mcode' compiled with MinGW32/MinGW64/UCRT64 


Paebbels:
**Is your feature request related to a problem? Please describe.**

To ease using `pyGHDL` for developers, a MinGW64 or UCRT64 standalone GHDL mcode is preferred. So the embedded `libghdl` can be used with a natively installed CPython 64-bit (not the Python within MSYS2). 

Currently, such a setup is achieved by:
1. configuring GHDL with a `PREFIX` outside of MSYS2 like `C:\Tools\GHDL`
2. building GHDL using `make`.
3. installing GHDL using `make install`.
4. manually copying some DLLs from MSYS2 installation directory to the GHDL installation directory.

An additional make target like `install_standalone` could trigger the `install target` and additionally do some copy/install operations so manual copying is replaced. This also ensures that MSYS2 updates are applied to the GHDL standalone installation as well (overwriting files with new variants).

**Describe the solution you'd like**

* Add a new make target.
* Restrict supported platforms to:
  * MinGW32 - mcode
  * MinGW64 - mcode
  * UCRT64 - mcode
* Copy these files:
  * for MinGW32:
    * `/c/msys64/mingw32/bin/libgcc_s_dw2-1.dll` &rarr; `$PREFIX/lib/libgcc_s_dw2-1.dll`
    * `/c/msys64/mingw32/bin/libgnat-12.dll` &rarr; `$PREFIX/lib/libgnat-12.dll`
    * `/c/msys64/mingw32/bin/libwinpthread-1.dll` &rarr; `$PREFIX/lib/libwinpthread-1.dll`
  * for MinGW64:
    * `/c/msys64/mingw64/bin/libgcc_s_seh-1.dll` &rarr; `$PREFIX/lib/libgcc_s_seh-1.dll`
    * `/c/msys64/mingw64/bin/libgnat-12.dll` &rarr; `$PREFIX/lib/libgnat-12.dll`
    * `/c/msys64/mingw64/bin/libwinpthread-1.dll` &rarr; `$PREFIX/lib/libwinpthread-1.dll`
  * for UCRT64:
    * `/c/msys64/ucrt64/bin/libgcc_s_seh-1.dll` &rarr; `$PREFIX/lib/libgcc_s_seh-1.dll`
    * `/c/msys64/ucrt64/bin/libgnat-12.dll` &rarr; `$PREFIX/lib/libgnat-12.dll`
[...]

Issue #2313, opened 2023-01-13
Build libghdl for MinGW64, UCRT64 and Linux with dependencies included


Paebbels:
**Is your feature request related to a problem? Please describe.**

I would like to ship precompiled binaries of `libghdl` with Python wheels to ease installation of `pyGHDL` as a standalone package via PIP and PyPI without the need of a full/local GHDL installation.

**Describe the solution you'd like**

`mcode` backend variants are preferred, because:
* no `*.o` files are generated by `libghdl`.
* `mcode` is faster in RAM for small files. Moreover, currently the performance of `llvm` is not needed.
* `libghdl` is currently used for parsing, no synthesis, no simulation.
* as local CPython installations will be 64-bit installations, `mcode 32-bit` is not needed.
* `mcode` has less depdendencies (`*.so`/`*.dll` files)

When looking at MinGW64 and UCRT64 variants of GHDL, these dependencies are used:
* `libgcc_s_seh-1.dll`
* `libwinpthread-1.dll`
* `libgnat-12.dll`

It looks like all of them have an equivalent `*.a` file, so the linker could link all dependencies together into a single `libghdl` shared library.

GHDL's libraries `libghdlvpi` and `libghw` are currently not needed for `pyGHDL`.

**Describe alternatives you've considered**

The current alternative is building and installing GHDL from sources using MSYS2/MinGW64. Afterwards, some DLLs are copied manually into the GHDL installation directory to create a standalone variant.

Such a standalone variant is also provided via nightly builds as a ZIP file. Unfortunately, updating requires again manual downloads and overwriting files.

**Additional context**

Python's wheel packages can:
* contain platform independent Python code
* platform specific precompiled binaries
* source code, that's compiled on the target machine using locally available compilers  
  (not every machine has an Ada compiler - and Windows even lacks a C compiler)

Paebbels:
I think only a variant for `pyGHDL` should use such a shared library. It's not intended for GHDL to use such a all-in-one shared library.

A proof of concept can be started for the MSYS2 environments MinGW64 and UCRT64. When this works, it can be expanded to Linux and macOS.

chenailuoding1:
I installed GHDL on Windows through MSYS2. Then I wanted to use pyGHDL in Python, but an error occurred.May I ask you some questions

Paebbels:
Please see my answer in #2808.

PR #2315, opened 2023-01-13
Added rules for DLLs from MSYS2.

Paebbels:
This PR tries to solve #2314.

@tgingold 
Open questions:
* how to get the GNAT major version?  
  This version number is used in the DLL filename.
* How to distinguish 32-bit MSYS2 environments (here MinGW32) from 64-bit environments (MinGW64 or UCRT64).  
  One environment uses `dw2` the other `seh` exception handling.


Paebbels:
I saw, you're using some `true`/`false` flags via make target names. Maybe it would be cleaner to not add a new make target, but offer a configure option like `--standalone`, which then sets a `true`/`false`.

Paebbels:
The latest commit contains this:
* `--enable`/`--disable-standalone`.
* new variable `LIBGCC_KIND` in configure derived from `build`.

I'm now locally testing. I assume we should add a check if standalone is used with another backend than mcode and if not used in MSYS2.

tgingold:
I have created #2320 to add new makefile targets.

Should we rewrite Test.yml to build standalone windows packages without reusing the pacman package ?
Or also create the standalone windows packages while creating the msys packages ?

Paebbels:
> Should we rewrite Test.yml to build standalone windows packages without reusing the pacman package ?

The current solution for packaging standalone mcode variants is a "dirty hack" I made with @umarcor. We used MSYS2 packages, because they contain all needed files, so we didn't need to figure out what to repackage for standalone.

So we used that package, installed it, added die DLLs and then packaged it again as ZIP.

If we could install GHDL mcode directly including DLLs and use this directory for creating a ZIP file, it would simplify a lot and reduce CI job dependencies.

tgingold:
I am not sure to understand how you want to create the standalone package.

There are multiple solutions:
1. the current one: unpack the msys pacman package, add dlls, zip
2. mixed: after building the pacman package, add the dlls and zip.  So during the same jobs two artifacts are produced.
3. full jobs: build mcode, add dlls and create the zip.

I don't really like 3), as it will be difficult to test.  I am ok with 1 and 2.

But the real question is how to produce the python wheel for #2313.

PR #2322, opened 2023-01-17
gh workflow: reduce tests, keep only windows

tgingold:
WIP for testing build of platform specific python wheels


tgingold:
To be reopen later.

PR #2364, opened 2023-02-26
doc: change theme to 'furo'

umarcor:
The sphinx_rtd_theme is based on a framework that was developed between 2013-2015, but has not been maintained since then: https://github.com/snide/wyrm. Despite the codebase still being functional, grunt and bower were replaced in the JS community years ago. Moreover, since RTD does not support building the site outsite of their own infrastructure, we use a fork from https://github.com/buildthedocs/sphinx.theme to support some features. In the last years, sphinx_rtd_theme has been lagging behind Python (docutils) changes, which produces frequent visualisation problems in the documentation. On top of that, maintaining/rebasing the fork adds latency (e.g. https://github.com/ghdl/ghdl/pull/2293#issuecomment-1445131943). Therefore, although we could hope they update the stack and keeping rebasing, I believe we should better change the theme.

We have been long looking for an alternative to sphinx_rtd_theme|sphinx_btd_theme, but we could not find any that convinced us. However, in the last 2y a pypa member, psf fellow, and sphinx-doc contributor developed [pradyunsg/furo](https://github.com/pradyunsg/furo). It's used by urllib3, pip, black...: https://github.com/pradyunsg/furo#used-by. See https://github.com/pradyunsg/furo#elevator-pitch:

> - Intentionally minimal --- the most important thing is the content, not the scaffolding around it.
> - Responsive --- adapting perfectly to the available screen space, to work on all sorts of devices.
> - Customisable --- change the color palette, font families, logo and more!
> - Easy to navigate --- with carefully-designed sidebar navigation and inter-page links.
> - Good looking content --- through clear typography and well-stylised elements.
> - Good looking search --- helps readers find what they want quickly.
> - Biased for smaller docsets --- intended for smaller documentation sets, where presenting the entire hierarchy in the sidebar is not overwhelming.

The only drawback in our case might be the "Biased for smaller docsets". However, I believe the GHDL documentation looks better with furo than sphinx_rtd_theme, particularly because it has two sidebars. See https://umarcor.github.io/ghdl/.


Paebbels:
It doesn't look like https://umarcor.github.io/ghdl/ is showing the new theme.

image

umarcor:
@Paebbels, refresh (Ctrl+F5).

Paebbels:
Here is my comparison:

| Old design | New design |
| ---------- | ---------- |
| image | image |

Advantages of `furo`:
* It's not outdated.
* It doesn't present Python class fields or methods in multiple ragged columns.  
  image
* Admonition styling seams to be better:  
  | Warning | Hint | Note | Important | Attention |
  | --- | --- | --- | --- | --- |
  | image | image | 
image
 | image | image |


Drawbacks of `furo`:
* Navigation is not synchronized  
  Opening a Python class leads to lots of scrolling and searching to find the next class or package.  
  Also in side a page, the right navigation is also not synchronized.  
  image
* Bigger font size
  * Less content per page 
  * More scrolling
* It's not using the full screen size
  * Less content per page
  * More scrolling
* More spacing between headline in navigation


---------
If we get font sizes and enormous space consumption handled by user defined CSS, then the biggest and most annoying problem would be the synchronization of navigation bars!

umarcor:
@Paebbels, with regard to synchronisation, I believe they are in sync, but not autoscrolled. As you scroll the content, the "highlighted" item will change in the right sidebar. If you move to a different section in the same page by clicking on the right sidebar, it will keep the scroll position of the sidebars. However, when you change to a different page, the item is shown bold in the left sidebar and when loaded it scrolls both sidebars at the top; they are not moved automatically. Therefore, I think you are missing the "autoscrolling after page loading"?

umarcor:
With regard to spacing and font size, I agree that `17.6px` might be excessive. Yet, I would be comfortable with `16px` as the default size on "regular and large" screens, and `14px` on smaller devices. See https://github.com/pradyunsg/furo/discussions/369#discussioncomment-5195302.

Even though the theme does not use the full width on FullHD screens, I believe that would be better addressed by aligning the sidebars at the edges, rather than increasing the width of the body signifficantly. That is, I would increase the whitespace between the content and the sidebars, and I would do them wider. This is because, as you know from LaTeX (see also https://www.mediawiki.org/wiki/Reading/Web/Desktop_Improvements/Features/Limiting_content_width#Goals_and_Motivation), we read better if paragraphs are not too wide.

Paebbels:
I miss when a page is loaded, the navigation shows that page, so I can click the next link in navigation to go to the next page. Now I need to scroll a while to find my current page, so I can click. This is very inconvenient.

Paebbels:
Yes, paragraphs shouldn't be to wide, so they look like paragraphs, but not one-liners. On the other hand (I) and others have big screens and it could fit more content and not scroll so much.

PR #2366, opened 2023-02-28
ci: enable UCRT64 LLVM and test MINGW64 LLVM

umarcor:
In https://github.com/ghdl/ghdl/commit/ed54e7f29fcadd7c443a5cf708b0723a82fa27aa, UCRT64 LLVM was disabled. That was/is provided through ghdl-setup-ci, and it's used by projects such as OSVVM (see https://github.com/OSVVM/OsvvmLibraries/blob/main/.github/workflows/Test.yml#L67). Those are now failing: https://github.com/OSVVM/OsvvmLibraries/actions/runs/4294470208.

This PR enables UCRT64 LLVM builds and tests. It also enables MINGW64 LLVM tests.

Alternatively, we can stop supporting UCRT64 and suggest users to consume MINGW64 LLVM only. Yet, I think that MSYS2 is making UCRT64 the default...

/cc @JimLewis

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -592,7 +592,7 @@ jobs:
       matrix:
         sys:
 #          - {icon: '🟪', pkg: 'llvm',  sys: 'MINGW32' } ! not yet functional
-#          - {icon: '🟦', pkg: 'llvm',  sys: 'MINGW64' }
+          - {icon: '🟦', pkg: 'llvm',  sys: 'MINGW64' }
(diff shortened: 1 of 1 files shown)


JimLewis:
@umarcor  Thanks for the notice.   I noticed that it caused the regression to fail.

Paebbels:
I also think UCRT64 is the way to go, but MinGW64 might be around for a long time.

tgingold:
I don't see any strong reason why both ucrt64 and mingw64 cannot work.  Let's see.

umarcor:
There was an error but after rerunning it's all green. Hence, I'm mergig.

PR #2367, opened 2023-02-28
ci: reusable workflows

umarcor:
This PR is based on #2366, so I'll keep it as a draft until we solve that.

Instead of having all the jobs defined in a single workflow (which makes it uncomfortable to maintain), this PR splits them into multiple local reusable workflows. They are called in the main workflow, and not triggered by anything else. Therefore, the outcome is the same we have now (same number of jobs, same artifacts, equivalent dependencies).

The "reusable workflow" feature did not exist when we wrote the current pipeline. There were composite actions, but that did not satisfy our needs.

As a fortunate side-effect, the jobs in the sidebar of the view are easier shown hierarchically: https://github.com/umarcor/ghdl/actions/runs/4296855104


.github/workflows/Coverage.yml
```diff
@@ -0,0 +1,93 @@
+name: Coverage
+
+on:
+  workflow_call:
+
+jobs:
+
+
+#
+# pyGHDL Coverage (MINGW64)
```
Paebbels, line 10: Can we reused actions from `pyTooling/actions`?

I think running coverage can be on Linux or Windows. Maybe we should extend `pyTooling/actions` to support Windows too.

.github/workflows/Coverage.yml
```diff
@@ -0,0 +1,93 @@
...
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: coverage
+          path: coverage.xml
+
+#
+# Coverage Publish (Ubuntu)
+#
+  coverage-publish:
```
Paebbels, line 66: This could be reused too, right?

.github/workflows/Coverage.yml
umarcor: Agree. However, let's keep it out of this PR, since I cannot deal with it now.
After this, split, we can focus on the similariries between Coverage.yml and pyTooling Actions.

tgingold (approved):
Yes, I am ok with the split, it indeed makes the flow easier to maintain.

PR #2406, opened 2023-04-03
pyGHDL: fix range handling in subtype indication

c-thaler:
According to §3.1, range can be a range expression or an attribute name. This must be respected when accessing the fields of the node. Trying to access the range expression fields on an attribute name node led to a libGHDL exception.

The bug is triggered by declarations using an attribute name as range:
```
signal int : integer range entries'range;
```

### DIFF
--- a/pyGHDL/dom/_Translate.py
+++ b/pyGHDL/dom/_Translate.py
@@ -333,7 +333,13 @@ def GetScalarConstrainedSubtypeFromNode(
     typeMarkName = GetNameOfNode(typeMark)
     simpleTypeMark = SimpleName(typeMark, typeMarkName)
     rangeConstraint = nodes.Get_Range_Constraint(subtypeIndicationNode)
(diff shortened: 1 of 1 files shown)


tgingold:
Thanks!

Paebbels:
@c-thaler do you have a Gitter account, maybe we could coordinate there.

c-thaler:
> @c-thaler do you have a Gitter account, maybe we could coordinate there.

Yes, I'm on Gitter with my github account.

Paebbels:
Gitter doesn't list you as `c-thaler`, `Christian` or just `Thaler`, it notes, some entries might be blocked due to privacy settings...

can you see me - `paebbels`?

c-thaler:
@Paebbels can't find you either

Maybe the invitation link works: https://matrix.to/#/@c-thaler-5e8798b4d73408ce4fdf357e:gitter.im

Issue #2418, opened 2023-04-19
Synthesis crashes when encouterning vector slices


bourquenoud:
**Description**
When synthesising, GHDL crashes if there are slices with constant width and dynamic bounds present in the code.

**How to reproduce?**
VHDL code:

```vhd :file: ent.vhd
library ieee;
use ieee.std_logic_1164.all;
use ieee.numeric_std.all;

entity ent is
    port (
        sel : in std_logic_vector(1 downto 0)
    );
end entity ent;

architecture arch of ent is
    signal reg : std_logic_vector(7 downto 0);
begin
    process(sel)
    begin
        reg <= (others => '0');

        -- This is the line that causes the error
        reg(to_integer(unsigned(sel)) + 1 downto to_integer(unsigned(sel))) <= (others => '1');
    end process;
end architecture arch;
```
Script:

```sh :image: ghdl/ghdl:bullseye-mcode
ghdl -a ent.vhd
ghdl -e ent
ghdl --synth --out=none ent
```

**Context**
- OS: Linux pop-os 6.2.6-76060206-generic
- Origin: Built from sources (SHA `6544b839fe6dceb56cb544bc4f69f0ab28100172`) with `./configure --enable-synth --default-pic`

```
******************** GHDL Bug occurred ***************************
Please report this bug on https://github.com/ghdl/ghdl/issues
GHDL release: 4.0.0-dev (3.0.0.r112.g6544b839f) [Dunoon edition]
Compiled with GNAT Version: 10.4.0
Target: x86_64-linux-gnu
[...]/minimal_bug/
Command line:
ghdl --synth --workdir=work --out=none top
Exception TYPES.INTERNAL_ERROR raised
Exception information:
raised TYPES.INTERNAL_ERROR : synth-vhdl_aggr.adb:725
Call stack traceback locations:
0x560df1f4b2c3 0x560df1f99601 0x560df1facba9 0x560df1fb0467 0x560df1fbdad2 0x560df1fbed8a 0x560df1fc07a2 0x560df1fc09df 0x560df1fa218e 0x560df1fa26c0 0x560df1fa2796 0x560df1f5c01b 0x560df2052937 0x560df2028adc 0x560df2106e30 0x560df1cbdec5 0x7f1137a29d8e 0x7f1137a29e3e 0x560df1cbc343 0xfffffffffffffffe
******************************************************************
```

**Additional context**
[...]

PR #2419, opened 2023-04-21
testsuite/synth: fixed the test for #2418

bourquenoud:
**Description**
Very small fix related to issue #2418. The test is based on the minimal reproducible example I made, which was wrong. This PR fixes it.

### DIFF
--- a/testsuite/synth/issue2418/ent.vhdl
+++ b/testsuite/synth/issue2418/ent.vhdl
@@ -16,6 +16,6 @@ begin
         reg <= (others => '0');
 
         -- This is the line that causes the error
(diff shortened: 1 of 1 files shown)


tgingold:
Sorry I forgot about his PR.
I'd like to keep the original test as it is, should I add a new test based on this PR ?

bourquenoud:
I think the current test or the corrected one both trigger the same error, but the correct version is more specific to the error. While I don't exactly know if this is the case in the current codebase, a vector of a width of one seems like a special case and could be processed in a way that wouldn't trigger the error with the current test.

But I am not familiar enough with this codebase to know if my correction is really relevant

PR #2439, opened 2023-05-14
Updated test cases

Paebbels:
# New Features

* Added CLI option `--library` to demo application for pyGHDL.dom.

# Updates

* Improved exception when opening a source file.

# Bug Fixes

* Fixed usage of symbols in unit tests and pretty-printing.


tgingold:
The test is trying to load `lib_Display/Display.pkg.vhdl` twice.

tgingold:
This should fix the crash:
```
--- a/testsuite/pyunit/dom/StopWatch.py
+++ b/testsuite/pyunit/dom/StopWatch.py
@@ -83,7 +83,7 @@ class Designs(TestCase):
     _encoderFiles = _encoderEntityFiles + (
         ("lib_Pretty",    Path("toplevel.Encoder.vhdl")),
     )
-    _displayFiles = _utilityCounterFiles + _displayEntityFiles + _displayPackageFiles + (
+    _displayFiles = _utilityCounterFiles + _displayEntityFiles + (
         ("lib_StopWatch", Path("toplevel.Display.vhdl")),
     )
     _stopwatchFiles = _utilityEntityFiles + _displayEntityFiles + _stopwatchEntityFiles + (
```

Paebbels:
While this removes the mistake, how can we make the code more save. Also a user using pyGHDL.dom could make this mistake.

Options I see:
* I could have an index of loaded files and raise an exception that file was already loaded.
* I could ask libghdl, if the file was already loaded / exists in libghdl's file mappings.
* Can we handle the exception?

Looking at the code, `name_table.Get_Identifier` registers the filename to libghdl. Should I have checked / can I check here for an error code?

Paebbels:
Oh I see, `name_table.Get_Identifier` returns an existing id, if it's already in the system, otherwise creates it. And then `files_map.Reserve_Source_File` fails because there is already a reserved memory region to it.

Could `files_map.Reserve_Source_File` return a zero or so in case of failure? Then it could be checked.

Paebbels:
The Python code is here: https://github.com/ghdl/ghdl/blob/master/pyGHDL/dom/NonStandard.py?ts=2#L185-L194

tgingold:
You can call Find_Source_File to get the index of a source file.

Paebbels:
`Find_Source_File` is not yet in the Python binding. I found `Find_Source_File` in `files_map` with 2 parameters. Currently I use `Reserve_Source_File` with first parameter (directory) set to `name_table.Null_Identifier`. Can I assume `Find_Source_File` also works well with directory set to `name_table.Null_Identifier`?

Is this ok, or do I need to setup directory structures as well? If so, how should it be done?

tgingold:
Yes, it is ok.  Directory set to Null_Identifier means local directory.

Paebbels:
What am I missing?
> AttributeError: function 'files_map__find_source_file' not found

```
@BindToLibGHDL("files_map__find_source_file")
def Find_Source_File(Directory: NameId, Name: NameId) -> SourceFileEntry:
    """
    Return an existing entry for a filename.

    :param Directory: ``Null_Identifier`` for :obj:`DirectoryId` means current directory.
    :param Name:      File name
    :return:          Return ``No_Source_File_Entry``, if the file does not exist.
    """
    return 0
```

Is there an export missing in the Ada code?

tgingold:
You need to make it public: copy the specification to the private part of the .ads file.

Paebbels:
Ah, sorry, now I see it. Search directed me to the body file adb.

With a check, I can now emit this exception in Python:
> pyGHDL.dom.DOMException: Source file 'C:\Git\GitHub\GHDL\ghdl\testsuite\pyunit\dom\examples\Display\lib_Display\Display.pkg.vhdl' already loaded.

Paebbels:
@Tcenova is interested in this PR.

PR #2446, opened 2023-05-28
Warning on incomplete sensitivity lists v3

### LINKED ISSUE #1425: Warning on incomplete process sensitivity lists

hipolitoguzman:
Close #1425 

- [x] DO indicate which issues are either fixed or closed by this PR. See [GitHub Help: Closing issues using keywords](https://help.github.com/articles/closing-issues-via-commit-messages/).

Hi! I saw that pull request https://github.com/ghdl/ghdl/pull/1865 was opened years ago but never merged, and I thought I could try to make it work with the current codebase since this would be a really cool feature to have.

> This pull request implements `-Wsynthesis` warning which is intended to check for good RTL design practices and reveal potential synthesis/simulation mismatches. Currently, the following checks are executed:
> 
>     1. Check for incomplete sensitivity lists
> 
>     2. Check for over-specified sensitivity lists.
>        Note that clocked processes are skipped in this dummy implementation (not checked). A process is classified as
>        clocked process if it contains 'event attribute or a call to one of "rising_edge" or "falling_edge" functions.
> 
> 
> The question of detecting what is a clocked process and how is the "expected" sensitivity list constructed might still be fine-tuned in the future (e.g. constructs within --pragma synthesis off should be ignored and not reported as false positively missing from sensitivity lists).

I have started from @Blebowski's branch and merged it into a recent fork of the project (from yesterday). I have also implemented @tgingold's feedback on https://github.com/ghdl/ghdl/pull/1865 as best as I have understood it.

The main functionality is working, but at this point I need a bit of guidance with three tests that are failing in the test suite (see "Further comments" for details)

- [x] DO make sure that the build is successful.

**Further comments**

In this version, all tests from the testsuite (testsuite/testsuite.sh) pass, with the exception of synthesis testcases ``issue1244``, ``issue1212`` and ``issue1005``. These testcases fail with a failing assert in vhdl-canon.adb:715, while extracting the sensitivity list from callees. Of course if we make ``-Wsynthesis`` disabled by default, all tests pass.

At this point I think I need some guidance here into how to discriminate the cases where we shouldn't do the checking of the sensitivity lists even if the ``-Wsynthesis`` option is activated, or some other way to avoid making these three testcases fail.

This is my first time contributing to GHDL so please let me know anything that could be improved.

:heart: Thank you!

Hipólito


tgingold:
Thank you for working on this improvement.
I have started the checks so that I can see the error.

tgingold:
The failure of the assertion in vhdl-canon.adb occurs because the subprogram body is not known when the sensitivity is extracted.

Unfortunately, there is no way to avoid that.

I would simply add a flag so that the assertion is not checked if the sensitivity is extracted for the purpose of -Wsensitivity.

hipolitoguzman:
Thanks a lot @tgingold for your help

With the new flag (``Allow_Unknown_Subprogram_Body``), the synth testcases now pass, so the full testsuite passes when compiling with mcode

But I am seeing now (thanks to the CI tests) that one of the tests, ``gna/bug0129``, fails for the llvm and gcc backends, but not for the mcode backend. (Actually, the test was also failing before, but I didn't look much into it because I thought it could be related to the other three.)

Interestingly, if I add a report that prints the value of ``tready`` to the ``p_main`` process in ``testsuite/gna/bug0129/repro1.vhdl``, the value checked by the third assert becomes correct again and the test doesn't fail:

```vhdl
  begin                                                                         
    report "execution of process p_main, cnt=" & natural'image(cnt);            
    report ("tready: " & std_logic'image(tready));  -- Adding this line makes the test pass
    count (cnt, v);                                                             
                                                                                
    cnt_next <= v;                                                              
  end process p_main;  
```

This makes kind of sense because ``tready`` does not directly appear inside the ``p_main`` process, but is read by the ``count`` procedure invoked inside the ``p_main`` process.

So it looks like an incomplete sensitivity list was created during the analysis step, when subprogram bodies are unknown, but wasn't re-created/updated during the elaboration step.

Is this correct? If so, is there a way to mark the generated sensitivity list as outdated or removing it so the elaboration step re-creates it with the full information?

Thanks!

P.S: By the way, I like ``-Wsensitivity`` more than ``-Wsynthesis``, so I'll change the name of the warning, and maybe ``-Wsynthesis`` can be added in the future when there are more checks that fall under its umbrella

tgingold:
Sorry for the delay.

I also agree that `-Wsensitivity` is a better name.

> This makes kind of sense because tready does not directly appear inside the p_main process, but is read by the count procedure invoked inside the p_main process.
> 
> So it looks like an incomplete sensitivity list was created during the analysis step, when subprogram bodies are unknown, but wasn't re-created/updated during the elaboration step.
> 
> Is this correct? If so, is there a way to mark the generated sensitivity list as outdated or removing it so the elaboration step re-creates it with the full information?

I haven't yet deeply looked, but you don't need to care about 'process (all)': by definition the sensitivity list is always correct.

hipolitoguzman:
Thank you @tgingold , I was able to make it work by skipping the checks if the process had "all" as their sensitivity list

I have also added a test for issue #1425 

I think this one is ready for code review :)

hipolitoguzman:
I'm not sure about how to fix the error in ``Test/Docs`` since ``error: externally-managed-environment``  seems to be related to the docker image used and not to the actual documentation being compiled

hipolitoguzman:
I have noticed the solution doesn't work well with records, so I'll work a bit more on it.

I'm not sure if I can mark this PR as draft in the meanwhile

tgingold:
I don't think the current approach will work well with records and arrays as soon as you have non locally static indexes.

Only synthesis can deal correctly with this warning.

However, I think such a warning could still be useful after analysis, but it requires some trade-offs.

hipolitoguzman:
Hi,

I have a solution that seems to work with (at least basic) records and arrays: if the golden element has a prefix, follow the prefixes until the top-most element, which is the one that should appear in the sensitivity list

I have updated the test I have added (in ``testsuite/gna/issue1425``) with three more examples: an array, a record, and a record-inside-a-record

I am getting a Codacy issue in ``testsuite/gna/issue1425/testsuite.sh``, since for some tests I set ``GHDL_STD_FLAGS`` to ``-Werror`` so the test fails if we get unexpected sensitivity list warnings. The variable is being used, so I'm not sure on how to solve the issue

Thank you!

hipolitoguzman:
Hi,

I managed to solve the Codacy issue so this PR is again ready to be reviewed. @tgingold, @umarcor, please let me know if I need to change anything

Thank you!

tgingold:
Looks in good shape.  Tests in progress.  I will probably rework it a little bit.
Thank you for this long work!

hipolitoguzman:
Thank you Tristan. Now that the tests seems to have correctly finished, let me know if you need me to merge master again in this branch to solve the conflicts

tgingold:
I have just merged a modified and rebased version of the branch.
The warning is not yet enabled by default. I first want to see how it behaves on real code.

PR #2468, opened 2023-07-17
Fix #1930: Vhdl Identifier Length Cap

### LINKED ISSUE #1930: Support identifiers longer than 1023 characters
Currently GHDL only supports identifiers (e.g. port names) up to 1023 characters long and (cleanly) terminates with a message like `test_long_log_msg/test.vhd:13:10: identifier is too long (> 1023)` otherwise. It would be nice to allow arbitrarily long identifiers (as long as allowed by the VHDL spec).

*(There's no pressing need to change anything. Of course, such long identifiers are unlikely in manually written code [unless you're using German words as identifiers], but are happening in auto-generated code. Filing this issue to have something to point to in cocotb where we disable a test for GHDL due to this limitation.)*

alpyen:
Handles/Fixes #1930 

Changes the scanning functions from fixed size strings to Unbounded_Strings as the LRM does allow for arbritrary identifier lengths rather than the current 1K/4K cap and does not meet standard that way.

Only the strings and their access have been changed, any code around it (such as keeping track of the Length of the string) has remained unmodified for now.

Since this change from fixed size strings to unbounded ones can have a bigger impact on performance, memory and so on I would highly appreciate taking a closer look before merging as this is my first contribution and Ada is not my strong suit.

The unlimited length works on my end, but it would be great if somebody could countercheck with a bigger project and see if everything works just fine.

### DIFF
--- a/src/vhdl/translate/trans.adb
+++ b/src/vhdl/translate/trans.adb
@@ -14,6 +14,7 @@
 --  You should have received a copy of the GNU General Public License
 --  along with this program.  If not, see <gnu.org/licenses>.
 
+with Ada.Strings.Unbounded;
 with Name_Table; -- use Name_Table;
 with Vhdl.Nodes_Priv;
 with Tables;
@@ -180,9 +181,11 @@ package body Trans is
    end Subprgs;
 
    package body Chap10 is
+      use Ada.Strings.Unbounded;
+
       --  Identifiers.
       --  The following functions are helpers to create ortho identifiers.
-      Identifier_Buffer : String (1 .. 4096);
+      Identifier_Buffer : Unbounded_String;
       Identifier_Len    : Natural := 0;
       Identifier_Start  : Natural := 1;
 
@@ -732,7 +735,7 @@ package body Trans is
 
       procedure Add_String (Len : in out Natural; Str : String) is
       begin
-         Identifier_Buffer (Len + 1 .. Len + Str'Length) := Str;
+         Insert (Identifier_Buffer, Len + 1, Str);
          Len := Len + Str'Length;
       end Add_String;
 
@@ -923,9 +926,9 @@ package body Trans is
          --Identifier_Buffer (L + Str'Length + 1) := Nul;
          if Is_Local then
             return Get_Identifier
-              (Identifier_Buffer (Identifier_Start .. L));
+              (Slice (Identifier_Buffer, Identifier_Start, L));
          else
-            return Get_Identifier (Identifier_Buffer (1 .. L));
+            return Get_Identifier (Slice (Identifier_Buffer, 1, L));
          end if;
       end Create_Id;
 
@@ -957,7 +960,7 @@ package body Trans is
             Add_Nat (Len, Natural (Val));
          end if;
          Add_String (Len, Str);
-         return Get_Identifier (Identifier_Buffer (1 .. Len));
+         return Get_Identifier (Slice (Identifier_Buffer, 1, Len));
       end Create_Identifier;
 
       function Create_Identifier (Str : String)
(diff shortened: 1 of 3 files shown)


tgingold:
The patch is correct, but I have some doubts about the performances.
An unbounded string is heavy, and this is an issue in the scanner.  An identifier is a very common token and each time an identifier is scanned, malloc/free will be called.

I suppose we need to create an ad-hoc type, more efficient than Unbounded_Strings.
(There is also grt-vstrings that could be used).

Did you do any performance evaluation ? (Just asking).

alpyen:
> The patch is correct, but I have some doubts about the performances. An unbounded string is heavy, and this is an issue in the scanner. An identifier is a very common token and each time an identifier is scanned, malloc/free will be called.

Yes, I was thinking about the same issue and wanted to see if this has any meaningful impact first.

> I suppose we need to create an ad-hoc type, more efficient than Unbounded_Strings. (There is also grt-vstrings that could be used).

Looking at the grt-vstrings it seems like you can allocate a a bigger chunk of memory instead of doing it bytewise to save on the mallocs/and frees, is that correct? If so, this would be a good substitution for unbounded strings. I didn't have it on my radar but I could try an implementation with that and see where it goes?

> Did you do any performance evaluation ? (Just asking).

Nothing of substance, the example file with a 1MB signal name and inverting it every few ns compiles and executes a few milliseconds measurably faster but it doesn't seem to scale with simulation runtime.
However I don't have any bigger project to compile against and see how big the performance hit actually is.

umarcor:
I believe we will need some specific test which is particularly "heavy" with regard to the number of identifiers used in the design. According to CI, the impact of this PR is none.

- This PR took 36min 26s: https://github.com/ghdl/ghdl/actions/runs/5578998055
- Previous executions of the master branch took: 37m 49s, 45m 32s, 43m 27s, 37m 41s...

Since this PR was executed once only, I would consider the fact it is 1 min faster just incidental. CI has a variability of +-5min over a 40min average.
Anyway, it seems to not have a negative impact when running all of the tests in the testsuites. I believe that's meaningful considering the dialogue above.

alpyen:
Agreed, I'm setting up the unisim compilation and will compare the three different versions (fixed size, unbounded, vstr) with one another once I have implemented the vstr one and will report back with benchmarks once that's done.

alpyen:
Assuming my implementation is correct (last commit), the results are quite surprising.
I let my testfile (this time with a 5M signal name) and unisim (Xilinx 2022.2) compile (as default with mcode), these are the results:

```´
averaged over 5 runs for each mode using "time"
mode: real, user, system time

== Fixed Size Strings ==

example.vhdl:
-i: 152.0ms, 138.0ms, 13.6ms
-a: 152.2ms, 135.4ms, 17.0ms
-r: 695.5ms, 643.2ms, 52.0ms

unisim:
run 1: 6m30s260ms, 6m22s830ms, 10s330ms
run 2: 6m29s530ms, 6m22s870ms, 09s870ms


== Unbounded Strings ==

example.vhdl:
-i: 196.4ms, 175.4ms, 20.8ms
-a: 201.0ms, 173.8ms, 27.4ms
-r: 849.4ms, 758.0ms, 91.2ms

unisim:
run 1: 7m04s410ms, 6m56s970ms, 10s630ms
run 2: 7m05s560ms, 6m58s400ms, 10s300ms


== VStrings ==

example.vhdl:
-i: 328.4ms, 310.0ms, 18.4ms
-a: 327.6ms, 308.2ms, 19.2ms
-r: 874.6ms, 811.2ms, 63.2ms

unisim:
run 1: 7m04s520ms, 6m54s620ms, 13s080ms
run 2: 7m05s010ms, 6m54s700ms, 13s510ms
```
Unbounded_Strings seem to be faster on my machine than the Vstrings implementation.
The chunked allocation resulting in lower mallocs/frees shows up in the lower system time,
however it looks like the additional software overhead(?) of the Vstrings results in much higher user time.

Unisim times however are almost identical, the identifiers used there are probably much shorter
and probably therefore don't need to be extended further than the initial 32 byte vstring length.

Seeing the much higher time for Vstrings I decided to change the backend to LLVM and see if a different compiler will yield different results and the fixed size version was the only test to complete (after around 7 minutes).
Both Unbounded_Strings and Vstrings did not finish. The Unbounded one was still 'running' after 45 minutes before I exited out, and the Vstrings version flat out crashed in the beginning.

tgingold:
(Back from being off).  I will investigate the llvm failure, this is weird.

tgingold:
I have merged the part for trans, using vstrings, after adjustments.

tgingold:
I have also modified and merge the change for the scanner. 

Closing this MR.

alpyen:
Thanks!

PR #2560, opened 2023-12-13
Fix a crash when called with incorrectly formatted "--fst" cmdline option

smaslovski:
Hi,

I noticed that ghdl crashed when called with an incorrectly formatted "--fst" cmdline option (e.g. without "=" between the option and the filename). The reported error was "CONSTRAINT_ERROR : simul-fst.adb:212 index check failed". This commit fixes this small problem.

BR,

Stanislav


tgingold:
Thanks!

PR #2574, opened 2024-01-06
Remove replacement characters since VHDL 2008

std-max:
After I had read the VHDL 93 LRM, I noticed the replacement characters were not supported anymore since VHDL 2008. Didn't know that!
So here are some minor fixes in the code to add a better support.

Changes:
- prevent from using '!' as a '|'
- update pragma for replacement char "%"

@tgingold I think we should also add a pragma for the use of colon ':' in string litteral but I dont know how you want to proceed. Feel free to edit this PR or give me some insights...

### DIFF
--- a/src/vhdl/vhdl-scanner.adb
+++ b/src/vhdl/vhdl-scanner.adb
@@ -2309,6 +2309,11 @@ package body Vhdl.Scanner is
                   Current_Token := Tok_Not_Equal;
                   Pos := Pos + 1;
                else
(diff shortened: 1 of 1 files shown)


tgingold:
It's ok.  I will slightly edit it: factorize the error messages and remove useless parentheses in the if-then condition.  That's minor!

For ':', you mean in numeric literals ?  Yes, it's not allowed anymore.  Feel free to create a new MR.

tgingold:
Thanks!

PR #2585, opened 2024-01-13
libghw: Speed up ghw_read_cycle_cont without null-type signals

### LINKED ISSUE #2548: libghw: handling large ghw files (performance)
**Description**
Use of libghw in gtkwave leads to long loadtimes when working with large ghw files.

**Expected behaviour**
Loadtime in the range of seconds (instead of minutes).

**How to reproduce?**
A suitable workload is linked here in [this specific comment in gtkwave #280](https://github.com/gtkwave/gtkwave/issues/280#issuecomment-1791391127)

**Additional context**
For discussion and analysis in a gtkwave context, see [gtkwave #280](https://github.com/gtkwave/gtkwave/issues/280) and this [Issue](https://gitlab.fel.cvut.cz/canbus/ctucanfd_ip_core/-/issues/284) and [Patch](https://gitlab.com/canfd/server-tools/-/commit/7d0e8b6d01b3406b8167f7fcbf44f4e714e7c7cf) from another project.

stefanlippuner:
**Description** 

This PR adds a flag if all the signals in the `h->sigs` array (other than the entry with index 0) have a non-NULL type.
In that case, the linear search in `ghw_read_hie` can be replaced by a simple addition.
This seems to be the regular case and provides a massive speedup (about 6x to 20x+ in a quick test) in gtkwave.
I've also added some checks for out-of-bounds accesses.

**Context**
I hope that this is the correct place to put these changes to libghw.
Otherwise, I can instead submit them to gtkwave first.

This PR is similar to the patch from #2548, in which the addition is used unconditionally.
Here's that patch without the white space changes: [output.diff.zip](https://github.com/ghdl/ghdl/files/13929182/output.diff.zip)
I'm not sure if it's possible to have a case where a non-NULL type signal exists, so I decided to add the flag to avoid breaking that case.

For the ghw file from https://github.com/gtkwave/gtkwave/issues/280#issuecomment-1791391127, the load time on my machine is 20s with these changes. Without them, I stopped it after 6min.

Closes #2548 

**Varia**

On a somewhat unrelated note: The indentation in these files was a mix of tabs and spaces. Is that intended? It certainly made it quite difficult to edit the code for me (using VS Code or Sublime Text)...

### DIFF
--- a/ghw/libghw.c
+++ b/ghw/libghw.c
@@ -1097,6 +1097,7 @@ ghw_read_hie (struct ghw_handler *h)
   h->nbr_sigs++;
   h->skip_sigs = NULL;
   h->flag_full_names = 0;
+  h->sigs_no_null = 0;
   h->sigs = (struct ghw_sig *) malloc (h->nbr_sigs * sizeof (struct ghw_sig));
   memset (h->sigs, 0, h->nbr_sigs * sizeof (struct ghw_sig));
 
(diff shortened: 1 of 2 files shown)


tgingold:
I think this is the right approach.  Thanks!

tgingold:
Is it OK ?
For the indentation, it is not intended.  I will fix that.

stefanlippuner:
:+1:  Thanks for having a look! 

Yes, I think the changes are ready to be merged. I only rebased my branch because the pipeline was failing and I thought your changes on the master might fix it. But now some other jobs failed, but I don't think it's related to my changes. How do you usually address that for GHDL?

tgingold:
@stefanlippuner A few tests are indeed randomly failing on windows with llvm be.  I was never able to reproduce that, but restarting the test is usually ok.  Thanks!

tgingold:
@Xiretza Yes, I am not sure that ghdl can currently generate such a file, but it's an implementation decision.

Issue #2589, opened 2024-01-16
Can no longer install pyGHDL using pip: Multiple .egg-info directories found


gralco:
**Description**
After having sucessfully installed pyGHDL (and thereby also ghdl-ls) using `pip install .` a few months back, I am no longer able to install a newer build as of today.

I have tested the following three commits, each with the same result: 71287d46036e45f2bcafbc06e4b1ca1b37d48b85 (previously installed on the Windows machine), 2135cbf1458bd1b8b8f42bda81222ab57bd66de6 (currently installed on the Fedora machine), d0693788ecfb77957d972d16ca07f27c0e76e17e (current head commit).

After seeing this issue on the Windows machine, I tested doing the `pip install .` on the Fedora machine without changing commits from what was used to install pyGHDL on it to begin with, which resulted in the same "Multiple .egg-info directories found" error.

On the Windows 10 machine I tried doing a `pip uninstall pyghdl` but that did not make it so that I am able to install a newer build. I reluctant to do the same the same on my Fedora machine out of fear of losing the existing pyGHDL install I have on it now (as what happened on Windows).

**Expected behavior**
After successfully updating GHDL via `configure --enable-synth`, `make`, `sudo make install`, I would expect that doing `pip install .` would update pyGHDL and ghdl-ls.

**How to reproduce?**
Run `pip install .`, this was done on two separate machines with different OSes that successfully installed pyGHDL using this method in the past.

Here's what I get when I attempt it (on both Fedora and Windows):
```
$ pip install . -v
Using pip 23.2.1 from /usr/lib/python3.12/site-packages/pip (python 3.12)
Defaulting to user installation because normal site-packages is not writeable
Processing /home/ekuzmenko/ghdl
  Running command pip subprocess to install build dependencies
  Collecting pyTooling>=4.0.1
    Obtaining dependency information for pyTooling>=4.0.1 from https://files.pythonhosted.org/packages/53/89/be6c6b36bb0d0fb71e915f85650f6bde40e3a27d31cd22a23ad4f4e08319/pyTooling-6.0.0-py3-none-any.whl.metadata
    Using cached pyTooling-6.0.0-py3-none-any.whl.metadata (21 kB)
  Collecting setuptools>=62.3.3
[...]

PR #2591, opened 2024-01-16
pyproject.toml: Specify pyTooling<5.0, fixes ghdl/ghdl#2589

gralco:
Since pyTooling 6.0.0 was released, attempting to install pyGHDL via `pip install .` fails with the following exception: "Multiple .egg-info directories found".

This corrects the problem.


tgingold:
Just merged a similar diff.  Thanks!

PR #2604, opened 2024-02-10
libghw - Add checks to prevent memory access issues when opening ghd files

stefanlippuner:
This MR adds checks to prevent memory access issues when opening ghw files. The type of fixed issues include:
-  Out-of-bounds accesses in arrays (with the array size and index read from the ghw file).
-  Casts to union members without checking the type.
-  Continued program execution after an error condition.
-  Overflows/underflows on malloc size variables.

I've added a function `ghw_error_exit_line` that's called when an invalid value is detected. I've chosen to call `exit(1)` in that case, after printing the . The original code had calls to `abort()` in some cases, I can also change it to do call `abort()`, if you prefer.

Note: codacy fails, but I did not change those lines.

### DIFF
--- a/ghw/libghw.c
+++ b/ghw/libghw.c
@@ -48,6 +48,19 @@ malloc_unwrap (size_t size)
   return ret;
 }
 
+// Call calloc ()
+void *
+calloc_unwrap (size_t nmemb, size_t size)
+{
+  void *ret = calloc(nmemb, size);
+  if (ret == NULL)
+  {
+    fprintf(stderr, "libghw could not allocate %zu elements of size %zu bytes. Terminating.\n", nmemb, size);
+    exit(12);
+  }
+  return ret;
+}
+
 /* Reopen H through decompressor DECOMP.  */
 
 static int
@@ -158,6 +171,15 @@ ghw_get_i32 (struct ghw_handler *h, unsigned char *b)
     return (b[0] << 24) | (b[1] << 16) | (b[2] << 8) | (b[3] << 0);
   else
     return (b[3] << 24) | (b[2] << 16) | (b[1] << 8) | (b[0] << 0);
+}
+
+// Read an i32, make sure it's positive, cast to u32
+uint32_t
+ghw_get_i32_positive (struct ghw_handler *h, unsigned char *b)
+{
+  int32_t val_i = ghw_get_i32 (h, b);
+  if (val_i < 0) ghw_error_exit ();
+  return (uint32_t) val_i;
 }
 
 int64_t
@@ -423,17 +445,14 @@ ghw_read_str (struct ghw_handler *h)
 
   // Read number of strings and total string size. Make sure that nbr_string
   // is not too large, as it's incremented afterwards.
-  h->nbr_str = (uint32_t) ghw_get_i32 (h, &hdr[4]);
-  if (h->nbr_str >= UINT32_MAX-8) ghw_error_exit();
+  h->nbr_str = ghw_get_i32_positive (h, &hdr[4]);
   h->nbr_str++;
   h->str_size = (uint32_t) ghw_get_i32 (h, &hdr[8]);
 
   // Allocate str_table, str_content. Avoid exceeding size_t limits.
-  uint64_t alloc_size = ((uint64_t)h->nbr_str) * sizeof (char *);
-  if (alloc_size > SIZE_MAX) ghw_error_exit();
-  h->str_table = (char **) malloc_unwrap (alloc_size);
+  h->str_table = (char **) calloc_unwrap(h->nbr_str, sizeof (char *));
 
-  alloc_size = h->str_size + h->nbr_str + 1;
+  uint64_t alloc_size = h->str_size + h->nbr_str + 1;
   if (alloc_size > SIZE_MAX) ghw_error_exit();
   h->str_content = (char *) malloc_unwrap (alloc_size);
   char *p_end = h->str_content + alloc_size;
@@ -458,7 +477,7 @@ ghw_read_str (struct ghw_handler *h)
       prev = h->str_table[i - 1];
       for (j = 0; j < prev_len; j++)
       {
-	if (p >= p_end) goto ghw_read_str_err;
+	if (p >= p_end) goto err_ghw_read_str;
 	*p++ = prev[j];
       }
 
@@ -469,11 +488,11 @@ ghw_read_str (struct ghw_handler *h)
 	    return -1;
 	  if ((c >= 0 && c <= 31) || (c >= 128 && c <= 159))
 	    break;
-	  if (p >= p_end) goto ghw_read_str_err;
+	  if (p >= p_end) goto err_ghw_read_str;
 	  *p++ = c;
 	}
 
-      if (p >= p_end) goto ghw_read_str_err;
+      if (p >= p_end) goto err_ghw_read_str;
       *p++ = 0;
 
       if (h->flag_verbose > 1)
@@ -496,8 +515,8 @@ ghw_read_str (struct ghw_handler *h)
     return -1;
   return 0;
 
-ghw_read_str_err:
-  printf("Invalid string entry in GHW file.\n");
+err_ghw_read_str:
+  fprintf(stderr, "Invalid string entry in GHW file.\n");
   ghw_error_exit();
 
 
@@ -623,7 +642,7 @@ ghw_read_array_subtype (struct ghw_handler *h, union ghw_type *base)
   sa->base = base;
   nbr_els = get_nbr_elements (arr->el);
   nbr_scalars = 1;
-  sa->rngs = malloc_unwrap (arr->nbr_dim * sizeof (union ghw_range *));
+  sa->rngs = calloc_unwrap (arr->nbr_dim, sizeof (union ghw_range *));
   for (j = 0; j < arr->nbr_dim; j++)
     {
       sa->rngs[j] = ghw_read_range (h);
@@ -666,7 +685,7 @@ ghw_read_record_subtype (struct ghw_handler *h, struct ghw_type_record *base)
       int nbr_scalars;
 
       sr->els =
-	malloc_unwrap (base->nbr_fields * sizeof (struct ghw_record_element));
+	calloc_unwrap (base->nbr_fields, sizeof (struct ghw_record_element));
       nbr_scalars = 0;
       for (j = 0; j < base->nbr_fields; j++)
 	{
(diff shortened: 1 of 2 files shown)


ghw/libghw.c
```diff
@@ -367,11 +420,23 @@ ghw_read_str (struct ghw_handler *h)
...
   h->nbr_str++;
-  h->str_size = ghw_get_i32 (h, &hdr[8]);
-  h->str_table = (char **) malloc ((h->nbr_str + 1) * sizeof (char *));
-  h->str_content = (char *) malloc (h->str_size + h->nbr_str + 1);
+  h->str_size = (uint32_t) ghw_get_i32 (h, &hdr[8]);
+
+  // Allocate str_table, str_content. Avoid exceeding size_t limits.
+  uint64_t alloc_size = ((uint64_t)h->nbr_str) * sizeof (char *);
+  if (alloc_size > SIZE_MAX) ghw_error_exit();
+  h->str_table = (char **) malloc_unwrap (alloc_size);
```
tgingold, line 434: I think it's clearer/safer to use `calloc`.

ghw/libghw.c
```diff
@@ -367,11 +420,23 @@ ghw_read_str (struct ghw_handler *h)
 
   if (hdr[0] != 0 || hdr[1] != 0 || hdr[2] != 0 || hdr[3] != 0)
     return -1;
-  h->nbr_str = ghw_get_i32 (h, &hdr[4]);
+
+  // Read number of strings and total string size. Make sure that nbr_string
+  // is not too large, as it's incremented afterwards.
+  h->nbr_str = (uint32_t) ghw_get_i32 (h, &hdr[4]);
```
tgingold, line 426: Simply test that the value of returned by `ghw_get_i32` is not negative.

ghw/libghw.c
```diff
@@ -392,7 +457,10 @@ ghw_read_str (struct ghw_handler *h)
       h->str_table[i] = p;
       prev = h->str_table[i - 1];
       for (j = 0; j < prev_len; j++)
+      {
+	if (p >= p_end) goto ghw_read_str_err;
```
tgingold, line 461: Please use `err_ghw_read_str` for the label.

ghw/libghw.c
```diff
@@ -424,6 +495,12 @@ ghw_read_str (struct ghw_handler *h)
   if (memcmp (hdr, "EOS", 4) != 0)
     return -1;
   return 0;
+
+ghw_read_str_err:
+  printf("Invalid string entry in GHW file.\n");
```
tgingold, line 500: Errors should be reported on `stderr`.

ghw/libghw.c
```diff
@@ -531,18 +610,20 @@ ghw_read_array_subtype (struct ghw_handler *h, union ghw_type *base)
...
 
-  sa = malloc (sizeof (struct ghw_subtype_array));
+  sa = malloc_unwrap (sizeof (struct ghw_subtype_array));
   sa->kind = ghdl_rtik_subtype_array;
   sa->name = NULL;
   sa->base = base;
   nbr_els = get_nbr_elements (arr->el);
   nbr_scalars = 1;
-  sa->rngs = malloc (arr->nbr_dim * sizeof (union ghw_range *));
+  sa->rngs = malloc_unwrap (arr->nbr_dim * sizeof (union ghw_range *));
```
tgingold, line 626: Use `calloc` to avoid overflow ?

ghw/libghw.c
```diff
@@ -585,7 +666,7 @@ ghw_read_record_subtype (struct ghw_handler *h, struct ghw_type_record *base)
       int nbr_scalars;
 
       sr->els =
-	malloc (base->nbr_fields * sizeof (struct ghw_record_element));
+	malloc_unwrap (base->nbr_fields * sizeof (struct ghw_record_element));
```
tgingold, line 669: likewise ?

ghw/libghw.c
```diff
@@ -628,24 +709,28 @@ ghw_read_type_bounds (struct ghw_handler *h, union ghw_type *base)
...
 
   if (hdr[0] != 0 || hdr[1] != 0 || hdr[2] != 0 || hdr[3] != 0)
     return -1;
-  h->nbr_types = ghw_get_i32 (h, &hdr[4]);
-  h->types =
-    (union ghw_type **) malloc (h->nbr_types * sizeof (union ghw_type *));
+  h->nbr_types = (uint32_t) ghw_get_i32 (h, &hdr[4]);
+
+  // Allocate type storage, initialize it to 0 to catch invalid accesses.
+  uint64_t h_types_size = (uint64_t)h->nbr_types * sizeof (union ghw_type *);
```
tgingold, line 730: Check non negative result, use `calloc`

ghw/libghw.c
```diff
@@ -683,9 +771,9 @@ ghw_read_type (struct ghw_handler *h)
 	      printf ("\n");
 	    h->types[i] = (union ghw_type *) e;
 	    break;
-	  err_b2:
-	    free (e->lits);
-	    free (e);
+	  err_rt_b2:
+	    if (e->lits != NULL) free (e->lits);
```
tgingold, line 775: (There is no need to check for NULL when calling `free`)

ghw/libghw.c
```diff
@@ -1078,13 +1171,14 @@ ghw_read_hie (struct ghw_handler *h)
   /* Number of declared signals (which may be composite).  */
   nbr_sigs = ghw_get_i32 (h, &hdr[8]);
   /* Number of basic signals.  */
-  h->nbr_sigs = ghw_get_i32 (h, &hdr[12]);
+  h->nbr_sigs = (uint32_t) ghw_get_i32 (h, &hdr[12]);
+  if (h->nbr_sigs > UINT32_MAX - 10) ghw_error_exit();
```
tgingold, line 1175: Check for non-negative value instead (maybe we should create `ghw_get_i32`.

ghw/libghw.c
```diff
@@ -1098,8 +1192,10 @@ ghw_read_hie (struct ghw_handler *h)
   h->skip_sigs = NULL;
   h->flag_full_names = 0;
   h->sigs_no_null = 0;
-  h->sigs = (struct ghw_sig *) malloc (h->nbr_sigs * sizeof (struct ghw_sig));
-  memset (h->sigs, 0, h->nbr_sigs * sizeof (struct ghw_sig));
+  uint64_t sigs_size = (uint64_t)h->nbr_sigs * sizeof (struct ghw_sig);
+  if (sigs_size > SIZE_MAX) ghw_error_exit();
+  h->sigs = (struct ghw_sig *) malloc_unwrap (sigs_size);
```
tgingold, line 1197: `calloc`.

ghw/libghw.c
```diff
@@ -1115,6 +1211,7 @@ ghw_read_hie (struct ghw_handler *h)
 
       if (t == ghw_hie_eos)
 	{
+	  if (blk->parent == NULL) ghw_error_exit();
```
tgingold, line 1214: I doubt this is possible.  Maybe use assert.

ghw/libghw.c
```diff
@@ -1149,7 +1247,7 @@ ghw_read_hie (struct ghw_handler *h)
 	case ghw_hie_design:
 	case ghw_hie_eos:
 	  /* Should not be here.  */
-	  abort ();
+	  ghw_error_exit();
```
tgingold, line 1250: `abort` is ok in that case (which shouldn't happen).

ghw/libghw.c
```diff
@@ -1189,7 +1287,7 @@ ghw_read_hie (struct ghw_handler *h)
 	    if (nbr_el < 0)
 	      return -1;
 	    sigs =
-	      (unsigned int *) malloc ((nbr_el + 1) * sizeof (unsigned int));
+	      (unsigned int *) malloc_unwrap ((nbr_el + 1) * sizeof (unsigned int));
```
tgingold, line 1290: `calloc`.

ghw/libghw.c
```diff
@@ -1299,7 +1397,7 @@ print_name (struct ghw_hie *hie, int full_names)
       p = p->parent;
       ++depth;
     }
-  buf = (struct ghw_hie **) malloc (depth * sizeof (struct ghw_hie *));
+  buf = (struct ghw_hie **) malloc_unwrap (depth * sizeof (struct ghw_hie *));
```
tgingold, line 1400: `calloc`.

ghw/libghw.c
```diff
@@ -1369,7 +1467,7 @@ ghw_disp_hie (struct ghw_handler *h, struct ghw_hie *top)
 	  break;
 	case ghw_hie_generic:
 	case ghw_hie_eos:
-	  abort ();
+	  ghw_error_exit();
```
tgingold, line 1470: `abort` is fine here.

ghw/libghw.c
```diff
@@ -1400,7 +1498,7 @@ ghw_disp_hie (struct ghw_handler *h, struct ghw_hie *top)
 	  }
 	  break;
 	default:
-	  abort ();
+	  ghw_error_exit();
```
tgingold, line 1501: `abort` is fine here.

tgingold (changes requested):

ghw/libghw.c
stefanlippuner: Agreed, fixed.

ghw/libghw.c
stefanlippuner: Sure, but don't blame me if someone wants to trace 2^31 signals :wink:

ghw/libghw.c
stefanlippuner: Changed.

ghw/libghw.c
stefanlippuner: Indeed, fixed.

ghw/libghw.c
stefanlippuner: Yes, changed.

ghw/libghw.c
stefanlippuner: Done.

ghw/libghw.c
stefanlippuner: Done. (I don't think the negative value check is actually necessary here)

ghw/libghw.c
stefanlippuner: Done, I've added a function `ghw_get_i32_positive`.

ghw/libghw.c
stefanlippuner: Indeed, removed.

ghw/libghw.c
stefanlippuner: Changed.

ghw/libghw.c
stefanlippuner: Sure, changed back to abort. My thinking in changing these was that it would be more consistent with using `ghw_error_exit` everywhere.

ghw/libghw.c
stefanlippuner: Changed.

ghw/libghw.c
stefanlippuner: Changed.

ghw/libghw.c
stefanlippuner: Reverted.

ghw/libghw.c
stefanlippuner: Reverted.

stefanlippuner:
Thanks for reviewing the changes! I think using calloc is a great suggestion!

tgingold:
Thanks!

PR #2609, opened 2024-02-16
update for GCC 14

sharkcz:
**Description**
update Makefile to accept GCC 14, which is the system compiler in Fedora 40

But
````
touch gcc/vhdl/lang.opt.urls
````
workaround is still required to overcome
````
...
make[2]: *** No rule to make target '../../gcc/vhdl/lang.opt.urls', needed by 's-options'.  Stop.
make[2]: *** Waiting for unfinished jobs....
make[2]: Leaving directory '/builddir/build/BUILD/gcc-14.0.1-20240208/obj-ppc64le-fedora_ghdl-linux/gcc'
make[1]: *** [Makefile:4678: all-gcc] Error 2
make[1]: *** Waiting for unfinished jobs....

````

Seems the `vhdl` backend in GCC will need a little update in the buildsystem to not require the missing file. I believe there is a way, because not all existing backends provide the `lang.opt.urls` file.


tgingold:
Thanks.  The urls file has to be investigated.

PR #2648, opened 2024-04-15
Partial merge of #762: "VPI: Adding vpiPort detection (Corrects regression introduced by #753)".

gatk555:
This is the result of a partial merge and user resolve of #762.  Changes to all files except grt-vpi.adb have been dropped,
as that is enough to fix the problem identified in the title.  Test suite passes (on Debian Linux) and continued use has
shown no problems,


gatk555:
Included in merged #2757, so closed.

PR #2674, opened 2024-06-19
Display the list of all drivers of a signal in case an unresolved signal has multiple drivers.

cderrien:
Whenever a source is added to a signal, check the number of sources after adding it. Upon detecting the presence of multiple drivers, display the drivers. Port sources are not displayed because I could not figure out how to exercise this path, though I report their presence if found.

This is an enhancement, and is not related to an issue.

### DIFF
--- a/src/grt/grt-processes.adb
+++ b/src/grt/grt-processes.adb
@@ -36,7 +36,6 @@ with Grt.Disp_Signals;
 with Grt.Stats;
 with Grt.Threads; use Grt.Threads;
 pragma Elaborate_All (Grt.Table);
-with Grt.Stdio;
 with Grt.Analog_Solver;
 
 package body Grt.Processes is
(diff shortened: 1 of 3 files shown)


tgingold:
There are some style issues that should be fixed in order for the builds to be ok.

Do you have a testcase ?  I think with the mcode backend, you can even have better messages.

cderrien:
Am I missing an option regarding gnat warnings ? I used the --enable-werror option in the configure script, but my attempts to build ghdl locally succeeded without error, only to then fail in the automatic builds.

I have no real testcase, the point of this was to give more information in case of errror.
As a demonstration, the following file triggers the multiple drivers error.
```
entity muldriver is
end entity;

architecture behaviour of muldriver is
  signal sig : bit;
begin
  sig <= '0';
  sig <= '1';
  sig <= '0';
end architecture;
```
Which gives a more detailed error message.
```
ghdl -a --std=08 -Wall muldriver.vhdl
ghdl -e --std=08 -Wall muldriver
./muldriver
Signal .muldriver(behaviour).sig is driven by :
.muldriver(behaviour).P0
.muldriver(behaviour).P1
./muldriver:error: several sources for unresolved signal
./muldriver:error: error during elaboration
```

It only displays two drivers, since the check is done upon adding a new driver, but at least the user can see what processes drive the signal.

tgingold:
Thanks!

PR #2757, opened 2024-09-21
VPI enhancements

gatk555:
Implement some additional parts of VPI: vpiVectorVal, vpiScaledRealTime and vpiSuppressTime; setting time and value in value change callbacks; and handle cbAfterDelay in vpi_remove_cb().  Tests included. It also includes PR #2648: "Partial merge of #762: VPI: Adding vpiPort detection (Corrects regression introduced by #753)" as these changes work together.

The use case is integration with Ngspice: with these changes a shared library built by GHDL can define the behaviour of a circuit element in a mixed-signal simulation.

### DIFF
--- a/src/grt/grt-vpi.adb
+++ b/src/grt/grt-vpi.adb
@@ -355,7 +355,7 @@ package body Grt.Vpi is
       Res : Std_Time;
    begin
       case V.mType is
-         when vpiSCaledRealTime =>
+         when vpiScaledRealTime =>
             Res := Std_Time (Unsigned_64 (V.mReal / Time_Scale_Unit));
          when vpiSimTime =>
             Res := Std_Time (Unsigned_64 (V.mHigh) * 2 ** 32);
@@ -1173,7 +1173,7 @@ package body Grt.Vpi is
       Len, Chunks, Bits, Base : Ghdl_Index_Type;
       A, Aall, B, Ball : Unsigned_32;
       Pointer : p_vpi_vecval;
-    begin
+   begin
       Info := Get_Value_Obj (Obj);
 
       case Info.Vtype is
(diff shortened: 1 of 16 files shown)


tgingold:
Thank you for the contribution.  There are still a few style issues; can you fix them ?

gatk555:
The Codacy report?  It seems it wants to move the declaration of variable "name" down into the "while" block.  Ugly in my opinion, but I am happy to do it if wanted.  The rest seem to be complaints about shell variable substitutions.  Indeed, quotes would add robustness, but there are hundreds of instances in all the other tests.

I plan to force-push in any case, as I realised that there is a memory leak.  If you can confirm that those are the fixes wanted, I will add them.  There is a related style question: I fixed the leak by reworking Free() in grt-vpi.adb to add explicit de-allocation.  Would Finalize be preferred style?

Thanks, G

tgingold:
No, not the codacy reports but the build failures.  They are due to style checks.

src/grt/grt-vpi.adb
```diff
@@ -1695,6 +1729,17 @@ package body Grt.Vpi is
       --  There is one reference to the callback as it is registered.
       Res.Cb_Refcnt := 1;
 
+      --  Copy caller's Time and Value structs.
+
+      if Res.Cb.Time /= null then
```
tgingold, line 1877: Shouldn't you test `Data` instead ?

src/grt/grt-vpi.adb
```diff
@@ -1695,6 +1729,17 @@ package body Grt.Vpi is
...
+         Res.Cb.Time := new s_vpi_time;
+         Res.Cb.Time.All := Data.Time.All;
+      end if;
+      if Res.Cb.Value /= null then
```
tgingold, line 1881: Likewise ?

gatk555:
Many thanks! G,

gatk555:
The co-simulation components for ngspice that depend on this PR have been merged and should be included in the next release, ngspice-45, expected summer 2025.

PR #2803, opened 2024-11-21
Add guix as a package manager providing ghdl binaries

csantosb:
-------

**Description**

It is possible now to use Guix to install ghdl binaries (clang backend only).

An example log of its use: https://builds.sr.ht/~csantosb/job/1371825


tgingold:
Thanks!

Issue #2728, opened 2024-08-20
Elaborated design available through pyGHDL


nrhint:
**Is your feature request related to a problem? Please describe.**
Summary:
Is there a plan to allow elaboration through the LSP? (https://github.com/ghdl/ghdl/issues/1753#issuecomment-834634597)

Background:
I am working on a program that parses VHDL and then converts it to other formats and manipulates it. In the past we wrote code that parses VHDL. Recently I have been working on integrating GHDL into the program to use GHDL to parse and analyze the code. I have been using the LSP for this and it has been very excellent. However I ran into a snag today where I needed to get the value of a constant from a package. Here are minimum examples:

my_pkg.vhd:
`
library IEEE;
use IEEE.math_real.all;
use work.my_pkg.all;

package my_other_pkg is

  constant FOO : real := 100;
  constant BAR : natural := my_func(integer(FOO));

end package my_other_pkg;
`

my_other_pkg.vhd:
`
library IEEE;
use IEEE.math_real.all;
use IEEE.numeric_std.all;

package my_pkg is

  function my_func (num : natural) return natural is
  begin
    return num / 3 + 1;
  end function my_func;

end package my_pkg;
`

Looking at what I can of the tree there is no elaborated value for the constant `BAR`. Other times there are values such as: `constant BAR : natural := FOO/3'` does provide a value (accessed through `nodes.Get_Default_Value`) From what I can tell if a function is builtin to the IEEE library the value is elaborated/resolved however if it is a user function it is not. I know that GHDL can actually resolve this because if I simulate the code it does have the correct value in either case.

**Describe the solution you'd like**
I would like to be able to get the elaborated values of constants and other declarations through the LSP

**Describe alternatives you've considered**
The alternative would be to return the function to the other program and then run the function there however that is what we are trying to avoid and remove from that program.

**Additional context**
I did go through an try to create an endpoint in the LSP that would perform this action and it works for many constants and signals (haven't finished all the declarations yet) however it fails in this case (which is my problem) and I can not find the actual value anywhere in the LSP (or XML) tree

Paebbels:
I'm not sure LSP (`pyGHDL.LSP`) is the the right interface. There is `pyGHDL.libghdl` as a low-level binding to `libghdl...{so|dll|dylib}`. Both, LSP and DOM are using it to access the provided features (parse, analyze, ...).

[...]

PR #2807, opened 2024-11-23
WIP Added binding to elaboration of the toplevel design #2728

nrhint:
Pull request for feedback on issue #2728 . I added a binding to libghdl that when I poke at appears to work however I have not yet managed to actually get the design to elaborate. When I pass a node from the tree to be elaborated I typically get the error `no field Block_Configuration`. I believe from looking at the source that I am basing my changes on (`src/ghdldrv/ghdlsynth.adb`) that the design needs to be configured first. There is not a binding for that either however I can see that things are flags for configuration status when inspecting the tree. In my testing I also attempted to bind the `ghdl_synth` function for verifying these changes could work however I was unable to get that working either. I would love feedback and suggestions for how to move forward.

### DIFF
--- a/pyGHDL/libghdl/utils.py
+++ b/pyGHDL/libghdl/utils.py
@@ -36,6 +36,7 @@ from typing import List, Any, Generator
 
 from pyTooling.Decorators import export
 
+from pyGHDL.libghdl._decorator import BindToLibGHDL
 from pyGHDL.libghdl._decorator import EnumLookupTable
 from pyGHDL.libghdl._types import NameId
 import pyGHDL.libghdl.name_table as name_table
@@ -455,3 +456,8 @@ def sequential_iter(n) -> Generator[Any, None, None]:
         return
     else:
         raise Exception("Unknown node of kind {}".format(kind_image(k)))
+
+@export
+@BindToLibGHDL("elab__vhdl_insts__elab_top_unit")
+def elab_top_unit(config: int) -> int:
+    return 0    


nrhint:
I also noticed that there are 2 flags that are similar. There is an `Elab_Flag` and an `Elaborated_Flag`. I am curious what the difference is as they seem to be mutually exclusive to each other.

tgingold:
I agree having two flags with a similar name is confusing.
Elab_Flag is a local flag: it is used in some functions to detect if a node has already been processed.
At the end of the function, it has to be cleared.

Elaborated_Flag is present only on a few nodes to detect use before elaboration errors (eg: a user read the value of a constant before the full definition).

pyGHDL/libghdl/utils.py
```diff
@@ -455,3 +456,8 @@ def sequential_iter(n) -> Generator[Any, None, None]:
         return
     else:
         raise Exception("Unknown node of kind {}".format(kind_image(k)))
+
+@export
+@BindToLibGHDL("elab__vhdl_insts__elab_top_unit")
+def elab_top_unit(config: int) -> int:
+    return 0
```
Paebbels, line 463: Please add a doc-string to this function.  
See this example: https://github.com/ghdl/ghdl/blob/master/pyGHDL/libghdl/files_map.py#L258

```suggestion
    return 0  # pragma: no cover
```

pyGHDL/libghdl/utils.py
```diff
@@ -36,6 +36,7 @@
 
 from pyTooling.Decorators import export
 
+from pyGHDL.libghdl._decorator import BindToLibGHDL
 from pyGHDL.libghdl._decorator import EnumLookupTable
```
Paebbels, line 40:
```suggestion
from pyGHDL.libghdl._decorator import BindToLibGHDL, EnumLookupTable
```

nrhint:
Excellent, I will make those changes you have suggested. When actually trying to use the function I get the ada error `no field Block_Configuration`. Is this because the nodes need to be configured before elaboration?

Paebbels:
I think that's a question for @tgingold.

nrhint:
Coming back to this after a while and I have a few questions. 

For https://github.com/ghdl/ghdl/blob/master/src/ghdldrv/ghdlsynth.adb line 554 I see that the elab argument is only taking the configuration in as an input. (my code is old and I will pull the latest tomorrow). Before it was needing a primary and secondary argument and I was wondering if there is only a primary what to do for the secondary. 

The second question is if I am elaborating multiple files do I need to do them in order or if they are all already in the work library will GHDL elaborate them in the correct order for the LSP?

tgingold:
@nrhint `Elab_Top_Unit` argument is a configuration declaration.  If you have an architecture, take the default configuration from the architecture (`Get_Default_Configuration_Declaration`). If you have an entity, get the latest architecture for it (`Libraries.Get_Latest_Architecture`).

I am not sure what do you mean by order for multiple elaborations.  A full design tree is elaborated, which includes all the dependencies.

PR #2815, opened 2024-11-28
Activate GCC backend in CI

Paebbels:
This PR is for investigation and discussion on how to reactivate GHDL + GCC backend compilation in GitHub Actions.


Paebbels:
@tgingold could you take over this PR and finish what ever is needed to get GCC working again?

I tried to extract the correct command order from the old `run-ci.sh`, but I couldn't figure out whats missing. You're free to modify this PR as you need.

PR #2817, opened 2024-11-29
Pipeline finetuning

Paebbels:
Currently, TAR/GZ archives have `bin`, `lib`, ... as root directories in the tarball.

As I learned today, a convention enforces archive creates to add a single root-directory into the tarball, which has the same name as the archive, so at decompression, users don't need to create a directory.

> I my opinion this is a still and almost dumb convention, we need to follow this convention ...

Therefore, the job template used in the pipeline supports a new mode, so artifacts packages as an archive can be packaged with such a single root-directory.

**Modifications:**
* Renamed `*.tgz` to `*.tar.gz` (this will match the files from MSYS2/Pacman `*.tar.zst`.
* Package macOS and Ubuntu build results in TAR/GZ archives with a single root-directory.

### DIFF
--- a/.github/workflows/Pipeline.yml
+++ b/.github/workflows/Pipeline.yml
@@ -298,12 +298,8 @@ jobs:
       - Ubuntu
       - Windows-fast
       - Windows
-      - PublishToGitHubPages
       - Python
-      - StaticTypeCheck
-      - DocCoverage
-      - PublishCoverageResults
-      - PublishTestResults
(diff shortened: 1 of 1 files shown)


Paebbels:
Assets for Ubuntu and macOS have no permissions set. These get lost by transporting build results via artifacts within GitHub Action workflows.

See also: https://github.com/actions/upload-artifact/issues/38

PR #2833, opened 2024-12-15
fix time overflow(issue#2832)

### LINKED ISSUE #2832: "wait until" timeout over flow
**Description**
Created an infinite loop. "wait until" syntax timeout overflow

**Expected behaviour**
Like integer overflow error instead of  bug
 
**How to reproduce?**
```vhd :file: ent.vhd
entity ent is
end entity;

architecture a of ent is
    signal sig : integer := 0;
    begin
        process is
        begin
            while 1 = 1 loop
                report "expect overflow";
                wait on sig for 1000 ms;
            end loop;
        end process;
end architecture;
```

```sh :image: ghdl/ghdl:bullseye-mcode
ghdl -a ent.vhd
ghdl --elab-run ent
```

- OS: void-linux 

```
GHDL release: 4.1.0 (tarball) [Dunoon edition]
Compiled with GNAT Version: 13.2.0
Target: x86_64-unknown-linux-gnu
/home/user/src/vhdl/
Command line:
ghdl --elab-run ent
Exception CONSTRAINT_ERROR raised
Exception information:
raised CONSTRAINT_ERROR : grt-processes.adb:414 overflow check failed
```

Hojun-Cho:
Check overflow before add.
Fixes #2832.

### DIFF
--- a/src/grt/grt-processes.adb
+++ b/src/grt/grt-processes.adb
@@ -410,6 +410,9 @@ package body Grt.Processes is
          --  LRM93 8.1
          Error ("negative timeout clause", Filename, Line);
       end if;
+      if Current_Time > Std_Time'Last - Time then
+         Error ("timeout overflow", Filename, Line);
+      end if;
       Proc.Timeout := Current_Time + Time;
       Update_Process_First_Timeout (Proc);
    end Ghdl_Process_Wait_Set_Timeout;


tgingold:
It should be OK, but I am not sure if the overflow is an error.  Shouldn't the process be suspended forever instead ?
I will try to find the expected behaviour in the LRM.

PR #2840, opened 2024-12-19
Add WinGet information to README

### LINKED ISSUE #2771: Add to winget
The current installation workflow for windows is not ideal (download an archive, unpack and add bin to path).

Since we already have a scoop version (see #2680), can we get a winget version? ([how to submit a package to winget](https://learn.microsoft.com/en-us/windows/package-manager/package/))

Xcodo:
**Description**
Add information that GHDL is available on WinGet.

Following the existing style, no specific details of the package are given, but its details are

```
Name Id               Version Source
-------------------------------------
GHDL ghdl.ghdl.ucrt64 4.1.0   winget
```

Closes #2771.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Together, these third-party integrations make GHDL a comprehensive tool for digi
 # Getting GHDL
 
 - Pre-built packages:
-  - GHDL is available through the default package manager on most major distributions: Debian/Ubuntu, Fedora, Arch Linux, Guix, and MSYS2 for Windows. You can install GHDL directly through these package managers, making the installation process straightforward for most users.
+  - GHDL is available through the default package manager on most major Linux distributions (Debian/Ubuntu, Fedora, Arch Linux, Guix), as well as MSYS2 and WinGet for Windows. You can install GHDL directly through these package managers, making the installation process straightforward for most users.
   - After each successful CI run, [nightly](https://github.com/ghdl/ghdl/releases/tag/nightly) tarballs/zipfiles for Ubuntu and Windows (MSYS2) are updated. These nightly builds provide access to the latest features and updates, which may not yet be available in the package manager repositories.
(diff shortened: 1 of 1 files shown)


Paebbels:
> `ghdl.ghdl....`

Why is it 2 times `ghdl` in the name?

Is it possible to add mcode to the package name, in case we later offer an llvm or llvm-jit variant?

Xcodo:
The format is <organisation>.<application>.<specifics> - for open source software without a company behind it it's common to have the name twice (Python, WinMerge, InkScape, Cppcheck).

I requested the package, but the implementer put the `ucrt64` tag on the end after querying the two download options. The package name in the releases folder doesn't specify the backend at the moment, which is why the package ended up not having this info.

- `ghdl-MINGW32.zip`
- `ghdl-UCRT64.zip`

A change to the package availability can be requested in the WinGet repository if that's what's desired.

Paebbels:
>  for open source software without a company behind it it's common to have the name twice (Python, WinMerge, InkScape, Cppcheck).

Ah, understood.

-----
I would prefer these variants:
* ghdl-UCRT64-mcode
* ghdl-UCRT64-llvm
* ghdl-UCRT64-llvm-jit

MinGW32 isn't needed anymore and moreover, the Ada compiler was removed from MSYS2, thus we can't build for MinGW32 anymore.

MinGW64 is currently replaced by UCRT64. UCRT64 has more packages then MinGW64. Maintainers begin dropping support for it.

Xcodo:
My bad for not asking about the names first.

Perhaps the first step would be to change the package names in the GHDL GitHub Releases folder to be similar to the Linux packages, then once they are stable match them in WinGet. We can put publicizing this on hold until the next GHDL release.

Paebbels:
You can find the asset names for the next release here: https://github.com/ghdl/ghdl/blob/master/.github/workflows/Pipeline.yml#L343

Such a release will look like this: https://github.com/paebbels/ghdl/releases/tag/testing

Please note, there are asset names (download file) and display names are a bit more descriptive.

---
Any suggestions on these names what you would change?

parker-research:
I think this is good to merge and conclude? Pretty minor change.

PR #2899, opened 2025-02-17
Disable macOS-15 builds due to cstdio.h issue.

Paebbels:
By accident, macOS-15 was enabled in the pipeline. This PR is deactivating macOS-15 builds (for now).

The current investigation shows 2 options:
* The GNAT compiler from https://github.com/alire-project/GNAT-FSF-builds doesn't contain a definition of `FILE*` in `cstdio.h`.
* The GNAT compiler's `cstdio.h` don't contain `FILE*` itself and references a `_stdio.h`. Maybe this file is provided by the operating system, which has changed from macOS-14 to macOS-15.


tgingold:
Already fixed.

Paebbels:
So we can also close #2900 ?

PR #2901, opened 2025-02-18
Add macOS-15 assets to the asset list of the nightly release

Paebbels:
This PR adds macOS-15 assets to the asset list of the nightly release. It was tested on `Paebbels/ghdl` using tag `testing`. The pipeline failed, because the `inventory.json` file for `setup-ghdl` is loaded from `ghdl/ghdl` and not from `Paebbels/ghdl`.


tgingold:
Thanks!

PR #2910, opened 2025-02-25
pyGHDL: Record constraints

Paebbels:
This PR fixes some issues related to pyGHDL:
* Changed naming pattern introduced by PyPA for all new wheel files: changed from `pyGHDL...whl` to `pyghdl...whl`.
  * Not yet applied to MSYS2, because it uses an older wheel version.
* Bumped dependencies:
  * Supports latest pyVHDLModel v0.31.0.
* Improved `GHDLBaseException`.
* Use `pyTooling.Warning.WarningCollector` to raise and collect warnings.
* Fixed missing init parameter for `parent`.
* Handle `ConstrainedArraySubtypeSymbol` and `ConstrainedRecordSubtypeSymbol`.

It also resolves all exceptions when running the pyunit testsuite. Thus, `pyunit.dom` is now reactivated when running Python unit tests.


Paebbels:
@tgingold I think it's ready for review and merge. Besides the normal pipeline, which was successful, I also triggered a tagged run (testing: https://github.com/Paebbels/ghdl/actions/runs/13557385428), because PyPA changed the pattern for wheel files. So the release page needs to be reverified by that pipeline.

It's not fully perfect, but all pyunit testcases are working again, which was not the case since a few month.

-------------
/cc @DrPiWork, @fisherdog1

Paebbels:
OK. I modified too many pyGHDL file names. 2 variants are now reverted and a new tagged pipeline is here: https://github.com/Paebbels/ghdl/actions/runs/13557802425

Paebbels:
The pipeline finished and created again 33 assets (expected).

I would summarize, all is OK.

PR #2936, opened 2025-03-18
Updates LSP testcases

Paebbels:
# Changes

* Bumped dependencies.
* Set release to draft mode (instead of prerelease).
* Updated LSP testcases.
  * More type hints
  * Use more f-strings
  * Set encoding when opening files.


Paebbels:
@tgingold  
`Test004_Error_Project` is currently marked as xfail.  
Can we both have a look at it?

On Windows I see this results:
```json
[
  {
    "jsonrpc": "2.0",
    "method": "window/showMessage",
    "params": {
      "type": 1,
      "message": "json error in project file C:/Git/GitHub/GHDL/ghdl/testsuite/pyunit/lsp/004errprj\\hdl-prj.json:1:3"
    }
  }
]
```

The expected answer is:
```json
[
  {
    "jsonrpc": "2.0",
    "method": "window/showMessage",
    "params": {
      "type": 1,
      "message": "json error in project file @ROOT@/004errprj/hdl-prj.json:1:3"
    }
  }
]
```

So it looks like a relative path to `@ROOT@` is expected as well as path separators in POSIX style.

tgingold:
@Paebbels Probably we need to deal with the path separator.

Paebbels:
Was LSP called already with a wrong path separator, because it expects forward slashes?  

I'll try to find where this paths comes from.

Maybe we should have a check for a wrong "path style".

Issue #2973, opened 2025-04-13
[setup-ghdl] "nightly" tag gone?


stnolting:
I am using the CI action to get GHDL:

```
- name: '📦 Install GHDL'
  uses: ghdl/setup-ghdl@v1
  with:
    version: nightly
    backend: mcode
```

However, this fails.

![Image](https://github.com/user-attachments/assets/fb9dc5a9-c525-4721-8cca-3adc4d8493e6)

I'm no expert, but it seems like the "nightly" tag is gone. Hardcoding the latest version works:

- name: '📦 Install GHDL'
  uses: ghdl/setup-ghdl@v1
  with:
    version: v5.0.1
    backend: mcode
```

Paebbels:
This is related to #2972. I pushed/moved the `nightly` tag. I'll try to provide a PR today (stupid PyPA ...).

Please note due to another incompatibility of MSYS2 and LLVM, not all GHDL variants are supported by `setup-ghdl` (we needed to delete MSYS2 + LLVM combination).

Paebbels:
@stnolting, @jimlewis

A patch is running at Paebbels/GHDL (35 minutes). If this is ok, we can merge (+ review + 35 minutes). Then we can move the nightly tag (+ trigger time + 35 minutes).

Paebbels:
Nightly has been moved. (35 minutes)

stnolting:
That was fast! Thank you very much!

Paebbels:
> That was fast! Thank you very much!

Because we had the same issue already in https://github.com/ghdl/ghdl/commit/5c97ac274926173727a6eaa3cdc0a5ca6b0999a9 for Ubuntu :).

Issue #2972, opened 2025-04-13
Failing pipeline due to Python Packaging Authority


Paebbels:
The pipeline are failing to install Python packages, because PyPA (Python Packaging Authority) changed how to normalize package names. This stupid change is rolled out in different speed to our various platforms. Thus the hit faster on rolling releases like MSYS2 and might reach very late at Ubuntu or macOS. Moreover, we might have fun in Ubuntu because some GitHub CI images will use the old normalization and others will use the new normalization.

Even more crazy, why I have no appreciation, Python is case sensitive and allows for mixed case package names. Now we have the situation to deliver Python packages in an installation file with a different name ... In addition, the package name is still mixed case, just the installer file is all-lower case.

Paebbels:
@tgingold do you fix this in one of you next pushes or should I prepare a small PR?

tgingold:
Please, do the PR.

Paebbels:
I'm on it. I got many private notes and other issues here on GitHub about this issue.

PR #2974, opened 2025-04-13
Fix issues with pyGHDL package name

Paebbels:
# Changes

* Bumped dependencies
* Changed pyGHDL package name, because PyPA change the normalization rules for package names.
  * #2972 
  * #2973


Paebbels:
Should be finished in a few minutes. I killed 2 of your pipelines after they finished basic compilation and testing, because pyGHDL will fail in these pipeline. This gives more resources (macOS and Windows instances) for this fix.

PR #3036, opened 2025-08-11
update for GCC 15

sharkcz:
The usual version bump for GCC 15. But this time we need a change to `ortho-lang-9.c` caused by https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=55947b32c38a. The "ortho-lang" change itself look like https://src.fedoraproject.org/rpms/ghdl/blob/rawhide/f/ghdl-gcc15-ortho.patch

I suppose we want to fork the `ortho-lang-9.c` to `ortho-lang-15.c`, because it's not compatible with older gcc. And when doing that, we probably want to switch to the C++ filename as well, so `ortho-lang-15.cc` will be added and the `Makefile.in` updated appropriately. What do you think?


tgingold:
Yes, we can fork.
Or maybe add #if/#endif if possible, as this is quite a small change.

PR #3066, opened 2025-09-30
[libghw/performance] Improves the performance of loading big ghw files

simonebri:
**Description** 

This pr adds a signal index cache in libghw to drastically speed-up (especially in gtkwave) loading of big ghw files when null signals are present. 
The current algorithm parses again the list at every request, while this method improves performances.
The cache is then freed upon closure of the ghw file being loaded.

### DIFF
--- a/ghw/libghw.c
+++ b/ghw/libghw.c
@@ -1668,9 +1668,8 @@ ghw_read_cycle_cont (struct ghw_handler *h, int *list)
 	{
 	  /* build the cache if not existing yet */
 	  if(h->no_null_sig_cache == NULL) {
-	    h->no_null_sig_cache = malloc_unwrap(sizeof(uint32_t)*h->nbr_sigs);
-	    memset(h->no_null_sig_cache,0xFF,sizeof(uint32_t)*h->nbr_sigs);
-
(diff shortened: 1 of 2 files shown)


simonebri:
@tgingold apparently ghwdump was not setting ghw_handler to zero, thus my new cache pointer was not null. I've added an initialization of the cache array pointer in the gwh_open function, and now I get all test passing locally.

ghw/libghw.c
```diff
@@ -1653,33 +1662,43 @@ ghw_read_cycle_cont (struct ghw_handler *h, int *list)
...
-	    {
-	      i++;
-	      if (i >= h->nbr_sigs)
-		goto err;
-	      if (h->sigs[i].type != NULL)
-		d--;
+	  /* build the cache if not existing yet */
+	  if(h->no_null_sig_cache == NULL) {
+	    h->no_null_sig_cache = malloc_unwrap(sizeof(uint32_t)*h->nbr_sigs);
+	    memset(h->no_null_sig_cache,0xFF,sizeof(uint32_t)*h->nbr_sigs);
```
tgingold, line 1672: Looks ok for me.
I was about to merge it, but re-reading I don't understand why it is initialized to 0xff ?
Later, (line 1689), the value read from the cache is compared to 0.

tgingold:
Thanks!

PR #3161, opened 2026-01-14
Updates for MSYS2

Paebbels:
# Changes

* Updated Python 3.12 to 3.13 in MSYS2/MinGW64
* Updated Python 3.12 to 3.13 in MSYS2/UCRT64.

Somehow it looks like GitHub also partially updated some of the OS images from Python 3.9 to 3.12, thus I needed to add compatibility code depending on what runner / cached OS is selected.


Paebbels:
@tgingold it has been tested in `paebbels/ghdl` as normal commit on a branch and via tag `testing`.

Issue #3232, opened 2026-04-16
Synth: Invalid VHDL type names generated when block RAM signal names contain escaped characters


mvonbuelow:
**Description**
When block RAM signals contain escaped characters in their names, the VHDL synthesis output generates invalid type names for the associated arrays. Although VHDL supports (or GHDL generates) escaped identifiers (e.g., `\:\` or `\sig[123]\`), the synthesis process incorrectly reuses these names when deriving type identifiers, producing results such as `\sig[123]\_type`, which is not a valid VHDL identifier.

**Expected behaviour**
The expected type name should be something syntactically correct. I think it doesn't matter what exactly. I think the easiest is to use a static name `bram` or so, as there should not be any collision in the scope. https://github.com/ghdl/ghdl/blob/3dcaf42a57ef55edbcecee8a8bb1ce970c9333a1/src/synth/netlists-disp_vhdl.adb#L735

```vhd
-- [...]
    type bram_type is array (0 to 1) -- this is syntactically incorrect
      of std_logic_vector (31 downto 0);
    variable \:\ : bram_type :=
      (1 => "11111111111111111111111111111111",
       0 => "11111111111111111111111111111111");
-- [...]
```

If you agree with that solution, I can also make a PR myself.

**How to reproduce?**
```vhd :file: ent.vhd
library ieee;
use ieee.std_logic_1164.all;


entity tb is
  port (
    clk : in  std_logic;
    a   : in  std_logic_vector(31 downto 0);
    b   : out std_logic_vector(31 downto 0)
    );
end entity;

architecture synthesis of tb is
  type xyz is array (0 to 1) of std_logic_vector (31 downto 0);
  signal \:\ : xyz := (others => (others => '1'));
  signal idx : integer;
begin
  p : process (clk)
  begin
    if rising_edge(clk) then
      \:\(idx) <= a;
      b        <= \:\(idx);
    end if;
  end process;
end architecture;
```

```sh
ghdl synth --out=vhdl test.vhd -e tb
```

outputs:

```vhd
library ieee;
use ieee.std_logic_1164.all;
entity tb is
  port (
    clk: in std_logic;
    a: in std_logic_vector (31 downto 0);
    b: out std_logic_vector (31 downto 0)
  );
end entity;

library ieee;
use ieee.std_logic_1164.all;
use ieee.numeric_std.all;

architecture rtl of tb is
  signal wrap_clk: std_logic;
[...]

PR #3234, opened 2026-04-17
Synth: Fix invalid VHDL type names generated when block RAM signal names contain escaped characters

mvonbuelow:
**Description**
See #3232

Hope the test is okay. I will adapt if not :)

### DIFF
--- a/testsuite/synth/issue3232/testsuite.sh
+++ b/testsuite/synth/issue3232/testsuite.sh
@@ -1,0 +1,7 @@
+#! /bin/sh
+
+. ../../testenv.sh
+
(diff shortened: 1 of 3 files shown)


testsuite/synth/issue3232/testsuite.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/sh
+
+. ../../testenv.sh
+
+synth_only tb
+fgrep -q 'mem_type' syn_tb.vhdl
```
tgingold: Why not `synth_analyze` which also analyse the result ?

PR #3236, opened 2026-04-18
Dropped ISO-8859-1 characters from ieee2008 files.

petterreinholdtsen:
This ensure they are readable as UTF-8 and ASCII alike, and bring the Copyright lines in line with the ones in the libraries/ieee/ directory.

An alternative might be to switch to UTF-8, but from PR #3174 I suspect this also would require changing the
‎pyGHDL/libghdl/vhdl/utils.py processing and risk break user written code currently using ISO-8859-1 files because of the use of decode("iso-8859-1") in that file.

### DIFF
--- a/libraries/ieee2008/numeric_bit.vhdl
+++ b/libraries/ieee2008/numeric_bit.vhdl
@@ -58,7 +58,7 @@ use STD.TEXTIO.all;
 
 package NUMERIC_BIT is
   constant CopyRightNotice : STRING
-    := "Copyright © 2008 IEEE. All rights reserved.";
+    := "Copyright 2008 IEEE. All rights reserved.";
 
   --============================================================================
   -- Numeric Array Type Definitions
--- a/libraries/ieee2008/numeric_std.vhdl
+++ b/libraries/ieee2008/numeric_std.vhdl
@@ -68,7 +68,7 @@ use IEEE.STD_LOGIC_1164.all;
 
 package NUMERIC_STD is
   constant CopyRightNotice : STRING
-    := "Copyright © 2008 IEEE. All rights reserved.";
+    := "Copyright 2008 IEEE. All rights reserved.";
 
 
   --============================================================================


Paebbels:
The VHDL standard says files are `ISO-8859-1` (a.k.a `Latin-1`) encoded.

Also, you can't modify the copyright protected files from IEEE.

petterreinholdtsen:
[Patrick Lehmann]
> The VHDL standard says files are `ISO-8859-1` (a.k.a `Latin-1`)
> encoded.

Yes, I saw this mentioned in PR #3174.  I even saw comments stating how
UTF-8 is used more and more.  This make the files readable also in this
case.

What is your point?  That the files should stay unreadable as UTF-8?

-- 
Happy hacking
Petter Reinholdtsen

Paebbels:
These files are released by IEEE and the encoding of all VHDL files is specified by the VHDL LRM (IEEE Std. 1076-2019). I created this issue: [vasg/Packages#4: LEGAL: Wrong Copyright notice in files](https://opensource.ieee.org/vasg/Packages/-/issues/4), because even in VHDL 2019, it was wrongly updated.

See also: [LRM-Issues#290: ISO-8859-1 vs. UTF-8 Encoding](https://gitlab.com/IEEE-P1076/VHDL-Issues/-/work_items/290)

I'm open to discuss a change in the IEEE working group for either removing the critical characters (this depends on IEEE legal department) or changing the encoding to UTF-8. I would support that change to future proof VHDL.

In any good programming language, you can specify how to read and interpret text inputs from files by setting the encoding. It's not a big deal in e.g. Python to read these files as `Latin-1`.

petterreinholdtsen:
[Patrick Lehmann]
> I'm open to discuss a change in the IEEE working group for either
> removing the critical characters (this depends on IEEE legal
> department) or changing the encoding to UTF-8. I would support that
> change to future proof VHDL.

Great.  I hope you succeed.

Note, replacing © with (C) might be considered too.  I do not know the
legal consequences of using (C) in USA, but know it is commonly used
elsewhere in the code base.  As far as I know, the legal effect is the
same when using either of © and "Copyright", so using "Copyright ©" is
more using both belts and suspenders, but then again I do not see the
removal of © as a big deal as the copyright status of the file is still
clearly visible in its content.

-- 
Happy hacking
Petter Reinholdtsen

tgingold:
Although that's a very minor point, I don't think it is a good idea to slightly change the definition of those files.

Let's wait until IEEE changes those files.

PR #3258, opened 2026-05-12
Implement option --expect-finish for std.env.finish.

bezborodow:
Simulations may timeout before they are complete. This option will return a non-zero error code unless std.env.finish is explicitly called to terminate the simulator. This will allow automated testbenches to distinguish between a timeout and a successfully completed test.

Looking for feedback and review before merging.

### DIFF
--- a/doc/using/Simulation.rst
+++ b/doc/using/Simulation.rst
@@ -110,6 +110,15 @@ Here is the list of the most useful options. For further info, see :ref:`DEV:Deb
 
   .. index:: display time
 
+.. option:: --expect-finish
+
+  Simulations may timeout before they are complete. This option will return a
+  non-zero exit code status (indicating an error) unless ``std.env.finish`` is
+  explicitly called to terminate the simulation. This option will allow
+  automated testbenches to distinguish between a timeout and a successfully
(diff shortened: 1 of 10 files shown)


tgingold:
Why not.
It would be nice to add a small testbench and even nicer to document the new option in `doc/using/Simulation.rst`!

bezborodow:
I have added documentation and a testbench.

JimLewis:
What mechanism are you using to stop the code when it times out?   

What you are asking for is a feature that is easily addressed in VHDL code.   I will elaborate in #3259

bezborodow:
> What mechanism are you using to stop the code when it times out?

The idea is to use in combination with `--stop-time`. If the simulation ends without `std.env.finish`, then the new change with `--expect-finish` will return non-zero to the shell.

> What you are asking for is a feature that is easily addressed in VHDL code.

Currently, I use a watchdog process in VHDL to fail the simulation after a timeout.

JimLewis:
When you use `--stop-time` you split the knowledge of the test and make it so the scripts must know how long a particular test case runs.   Most test cases will benefit from the timeout/Watchdog in VHDL.

bezborodow:
Closing pull request 'wontfix' based on feedback.

PR #3291, opened 2026-06-19
grt: add extended VCD (EVCD) waveform writer (--evcd)

wm2015email:
## Summary

Adds a new signal dumper that writes **IEEE 1364 (Clause 18) extended VCD (EVCD)** files, selected at run time with `--evcd=FILENAME` (plus `--evcd-nodate`). Extended VCD currently has no support in GHDL.

Unlike the standard VCD writer, EVCD is **port-oriented** (`$dumpports` semantics): only ports are dumped, declared as `$var port`, and each is written as a `p` value record carrying **direction-aware state + drive-strength** information. Ports across the whole elaborated hierarchy are dumped, each under its own `$scope`.

Example output for a `std_logic_vector(3 downto 0)` input port holding `"0001"`:

```
$var port [3:0] <2 data_in $end
...
pDDDU 6660 0006 <2
```

## Implementation

- New `Grt.Evcd` module (`src/grt/grt-evcd.{ads,adb}`), modeled on `Grt.Vcd` and reusing its value-extraction helpers; it additionally records each port's `mode` to pick the EVCD state-character set (input `D/U/N/Z`, output `L/H/X/T`, inout unknown-direction `0/1/F/?`).
- `std_ulogic` has no Verilog drive strength, so driven values use `strong` (6) and high-impedance uses `highz` (0), matching the format's reference example bytes.
- Registered in both the GRT runtime (`grt-modules`) and the integrated `ghdl -r` path (`ghdlrun`).

## Testing

- New testsuite case `gna/evcd01` exercising `in`/`out`/`inout`, scalar and vector ports, asserting the `$var port` declarations, the `$dumpports` block, per-bit `p`-records, the inout three-state record, and `$vcdclose`.
- Built with the mcode backend and verified end-to-end.

Docs (`doc/using/Simulation.rst`) and `NEWS.md` updated.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,9 @@
 
 - New signal dumper for the extended VCD (EVCD) format, selected with the
   `--evcd=FILENAME` run option (IEEE 1364 `$dumpports`).  It dumps the design
-  ports with direction-aware state and drive-strength information.
+  ports with direction-aware state and drive-strength information.  Optional,
+  non-standard annotations (port direction, VHDL type, full nine-state
+  `std_logic`) can be added as `$comment` blocks with `--evcd=FILENAME:CODES`.
 
 ## [2026-03-07] 6.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+- New signal dumper for the extended VCD (EVCD) format, selected with the
+  `--evcd=FILENAME` run option (IEEE 1364 `$dumpports`).  It dumps the design
+  ports with direction-aware state and drive-strength information.
+
 ## [2026-03-07] 6.0
 
 - Initial support of vhdl2019 (with --std=19)
--- a/doc/using/Simulation.rst
+++ b/doc/using/Simulation.rst
@@ -293,6 +293,33 @@ Export waveforms
   VCD reader.  The default is to write ``std_ulogic`` as they are (so
   keeping states ``U/W/L/H/-``), which is supported by several VCD readers.
 
+.. option:: --evcd=<FILENAME>
+
+  .. index:: evcd
+
+  .. index:: extended value change dump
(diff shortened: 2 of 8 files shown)


wm2015email:
A couple of design notes to help review — happy to change any of these if you'd prefer different conventions:

**1. Port direction → EVCD state set.** Each port's `mode` selects the state-character group from IEEE 1364 Table 18: `in` → input chars (`D/U/N/Z`), `out`/`buffer` → output chars (`L/H/X/T`), and `inout`/`linkage` → the unknown-direction chars (`0/1/F/?`). Since `std_ulogic` has no Verilog drive strength, driven values use `strong` (6) and high-impedance uses `highz` (0); this reproduces the spec's own example bytes (`pU 0 6`, `pD 6 0`, `pX 6 6`, etc.).

**2. Ports are dumped throughout the whole hierarchy, not just the top entity.** Strict `$dumpports` dumps a single module's ports. For a VHDL flow the top unit is usually a port-less testbench, so I walk the elaborated hierarchy and emit every instance's ports (each under its own `$scope`), which is the practical analog of `$dumpports(testbench.DUT, …)`. Internal signals are still excluded — it remains ports-only. If you'd rather restrict it to the root instance's ports, that's a small change.

**3. Vector ports.** The spec example only shows scalar `p` records; for an N-bit port I concatenate per-bit state and per-bit strength fields (so width-1 reduces exactly to the scalar form), e.g. `data_in = "0001"` → `pDDDU 6660 0006 <2`. This matches what Synopsys/Verdi emit.

Value extraction reuses the existing `Grt.Vcd` helpers; the new module only adds the port table (carrying each port's `mode`) and the EVCD formatting. Verified end-to-end with the mcode backend, and `testsuite/gna/evcd01` covers in/out/inout scalar+vector ports.

PR #3292, opened 2026-06-19
Add native --flow dataflow database exporter

wm2015email:
## What

Adds a `-r --flow[=FILE]` run option (and a dedicated `ghdl --flow` command) that dumps a JSON **dataflow database** (extension `.flow`) directly from GHDL's elaborated model, in a single native pass.

```
ghdl -r tb --flow                 # writes ghdl.flow
ghdl -r tb --flow=design.flow     # writes design.flow
ghdl --flow [--out=FILE] tb       # elaborate + dump, no simulation
```

## Why

Existing external dataflow tracers need three passes (source scan + VPI + merge) only because VPI exposes no dataflow. GHDL already has it all elaborated and resolved, so one native pass replaces them.

## What's in the file

The new `Simul.Flow` package walks the analyzed AST and the elaborated `Simul.Vhdl_Elab` tables (Processes/Drivers/Sensitivity/Connect/Signals), right after `Compute_Sources`, and emits:

- **`modules`** — per entity/architecture: ports (dir/type, incl. constraints like `std_logic_vector(3 downto 0)`), generics, signals, constants, and the read-set / drive-set / sensitivity of every process and concurrent assignment, with source spans.
- **`hierarchy`** — the elaborated instance tree (generates unrolled with the real loop index, component instantiations resolved), with the **real** port map and per-instance resolved generic values.
- **`nets`** — one entry per canonical net, with the signals that collapse into it (port↔net unification) and its fan-in/fan-out cells.
- **`cells`** — one entry per elaborated process/assignment, classified combinational or clocked (with the clock net), and its driven/read nets.
- **`packages`** — package constants with values; plus `ghdl` provenance (version, hash, std).

## Notes

- The option is handled in the front-end (`ghdlrun`): captured in the post-top argument loop, dumped in `Compile_Elab_Setup`, and filtered out of the argv handed to grt (the dump happens at elaboration, before grt decodes run options). The `ghdl --flow` command mirrors `ghdl --synth` (elaborate, dump, stop).
- Only the interpreted / JIT-elaboration run path builds the required model; the compiled-elaboration path is unaffected.

## Tests / docs

- `testsuite/gna/flow01`–`flow07`: base dataflow, resolved generics, for-generate, component instantiation, if-generate pruning, package/architecture constants + type constraints, and the `ghdl --flow` command.
- Documented in `doc/using/Simulation.rst` and `NEWS.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,8 @@
   (ports/generics/signals and per-process read/drive sets), the elaborated
   hierarchy with real port maps, canonical nets with fan-in/fan-out and
   port/net collapsing, and combinational-vs-clocked cells.  Bare `--flow`
-  writes `ghdl.flow`.
+  writes `ghdl.flow`.  There is also a dedicated `ghdl --flow [--out=FILE]
+  UNIT` command that elaborates and dumps without running a simulation.
 
 ## [2026-03-07] 6.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,12 @@
+## Unreleased
+
+- New `--flow[=FILE]` run option: dump a native dataflow database (JSON, with
+  the `.flow` extension) of the elaborated design.  It records modules
+  (ports/generics/signals and per-process read/drive sets), the elaborated
+  hierarchy with real port maps, canonical nets with fan-in/fan-out and
+  port/net collapsing, and combinational-vs-clocked cells.  Bare `--flow`
+  writes `ghdl.flow`.
+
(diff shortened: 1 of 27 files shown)


tgingold:
Do we need to support both command and the option to generate the flow file at the start of a simulation ?

tgingold:
The CI is currently failing when running with this PR.  Please adjust!

PR #3299, opened 2026-06-28
Add vpi_get/put of vpiReal and vpiScaled values.

gatk555:
As described, but also corrects an error in a previous PR of mine that added support for vpiVectorVal.


tgingold:
Thanks!

PR #3313, opened 2026-07-22
pyGHDL.dom updates

Paebbels:
> [!NOTE]
> This branch accumulated 22 pull-requests, so the notes below are grouped by theme rather than
> enumerating individual classes. The linked pull-requests carry the per-construct detail.

**Requires:**
* pyVHDLModel ≥0.39.0
* pyTooling ≥8.19.0

# New Features

* **VHDL-2019 mode views.** Mode view declarations and elements, and ports declared `view v`, which
  no longer have to be forced through the subtype position. (Paebbels/ghdl#6, Paebbels/ghdl#15)
* **VHDL-2008 subprogram instantiation and aliases.** Generic subprograms can be instantiated, and
  alias declarations are translated with and without a subtype indication. (Paebbels/ghdl#7)
* **Configurations.** Configuration declarations, block and component configurations, binding
  indications and the entity/configuration aspects — the last design unit that reached the model as
  a bare shell. (Paebbels/ghdl#9)
* **Complete assignment statements.** Conditional and selected assignments, concurrent and
  sequential, plus signal force and release. (Paebbels/ghdl#12)
* **Block headers.** A block statement's generic clause, generic map, port clause and port map. (Paebbels/ghdl#17)
* **Discrete ranges and aggregate choices.** A discrete range is translated as a subtype indication
  where the source writes one, so `for i in integer range 0 to 7 loop` keeps its type mark. (Paebbels/ghdl#14)
* **Documentation from comments.** Doc-comments are extracted for design units, declarations, record
  elements, mode view elements and the PSL default clock. (Paebbels/ghdl#19)
* **Type conversions.** A parenthesis name may be a type conversion, not only an indexed or sliced
  name. (Paebbels/ghdl#18)

# Bug Fixes

* **Subprogram declarations without a body were dropped**, so a package's subprograms were missing
  from the model. (Paebbels/ghdl#11)
* **Constrained scalar subtypes lost their range constraint.** `ConstrainedScalarSubtypeSymbol` read
  the range and discarded it. A range written as a range attribute (`natural range v'range`) was
  discarded too, and a subtype indication that adds only a resolution function is now a
  `SimpleSubtypeSymbol` rather than a constrained subtype with no constraint. (Paebbels/ghdl#8, Paebbels/ghdl#21)
* **Generate-statement alternative labels** were not read, and bit-string literals lost their
  signedness and length. (Paebbels/ghdl#10)
* **Package instantiations** did not receive their context items or their generic map. (Paebbels/ghdl#5)
[...]

PR #3327, opened 2026-08-11
pyGHDL documentation

Paebbels:
> [!NOTE]
> This branch is split off `paebbels/dom-updates` (ghdl/ghdl#3313) and only the changes **on top of
> it** are described here. The diff against `master` also carries #3313's content; please review that
> one there.

Documentation for the `pyGHDL` Python package. No behaviour changes: doc-strings, three type
annotations, and one deleted file.

Doc-string coverage measured with `interrogate` (`fail-under = 59` in `pyproject.toml`):

| | before | after |
|---|---|---|
| `pyGHDL.dom` | 14.7 % | **100.0 %** |
| `pyGHDL.libghdl` | 8.3 % | 11.8 % |
| **`pyGHDL`** | 11.1 % | **39.2 %** |

# `pyGHDL.dom` — complete

Every module, class, field, `__init__`, dunder method, property, function and method now has a
doc-string whose `:param:` list matches its signature, verified by an AST pass.

* **238 classes.** 227 mirror a pyVHDLModel class and every undocumented one had a base carrying
  real text, so they reuse it with `@InheritDocString(VHDLModel_X, merge=True)` and add the
  pyGHDL-specific sentence. `BaseFirst` order is deliberate: the autoapi module template lists
  classes by their first line, so the pyGHDL note must come second or every entry in the module
  index reads alike.
* **228 `__init__` methods.** The 28 that already carried `@InheritDocString` are *fixed*, not
  extended: a `pyGHDL.dom` mirror takes the IIR node first and no `parent`, so the inherited
  parameter list described a different signature — all 28 disagreed. `Design` published
  `:param allowBlackbox:`, which it does not accept, and omitted `vhdlVersion`, which it does.
* **146 methods**, of which 113 are the `parse` classmethods. The IIR node parameter has 45 distinct
  names across them, so each names the construct it holds rather than sharing one sentence.
* **19 class fields**, **7 properties**, the module level, and the last dunder methods.

Wording conventions applied throughout: a keyword is marked with a new `:vhdlkw:` role (see below),
a construct is named rather than called "a model entity", and whether a statement's label is optional
was **checked against GHDL** rather than assumed — it is mandatory for the three instantiations, the
block statement and the three generate statements, optional everywhere else.

# `pyGHDL.libghdl` — the hand-written half

The six generated files (`nodes.py`, `nodes_meta.py`, `tokens.py`, `elocations.py`, `errorout.py`,
`std_names.py`) are **untouched**: they are written by `scripts/pnodespy.py` and a doc-string added
here would be lost on the next regeneration. Filling those is ghdl/ghdl#3328.

The 28 hand-written modules are complete — 27 module doc-strings naming the Ada package each binds,
3 classes and 17 functions and closures. The private raw bindings (`_Get_Name_Ptr`,
`_String8_Address`, …) now say which public function to use instead.

Three legacy patterns were cleared while sweeping: **61** `:return:` became `:returns:`, **31**
`Type: ``X``` suffixes were dropped (they name the *Ada* type, which the Python signature already
gives), and **22** `undocumented` placeholders were filled.

# Infrastructure

* **A `:vhdlkw:` role** in `doc/prolog.inc`, beside the existing `:pycode:` and colour roles, marking
  a VHDL keyword in prose so it is not mistaken for ordinary English — `others`, `for`, `report` and
  `process` all read as both. Rendered as emphasis with a CSS class, so the styling can change
  without touching a doc-string. It is deliberately **not** called `keyword`: Sphinx's standard
  domain owns that name, and registering another role under it is silently ignored.
* **`setup.py`** declares `pythonVersions`, and `pyGHDL/lsp/version.py` is deleted — nothing imported
  it and its `0.1dev` contradicted `pyGHDL.__version__`.
* **A malformed grid table** in `pyGHDL/lib/__init__.py` emitted `ERROR: Malformed table.` on every
  documentation build; it is a `list-table` now.

# Defects found and recorded

Several of these are behaviour, not documentation, so they are described in the doc-strings and left
for a separate change:

* **`Document.LibGHDLProcessingTime` raises `AttributeError`** rather than returning `None` when the
  document was constructed with `dontParse=True` — the backing field is only assigned when parsing
  ran, and `ExtendedType` uses slots. `Design`'s sibling fields are `Nullable[float] = None`, so the
  two classes disagree, and `pyGHDL/cli/dom.py:309` multiplies one of them unguarded.
* **`FunctionCall.__init__` discards its `operand`** — `super().__init__()` takes no arguments
  because `pyVHDLModel.Expression.FunctionCall` is a `pass` stub. Recorded as a `.. todo::`; the class
  is never constructed, so nothing is broken today.
* **208 parameters default to `None` while annotated with the bare type**, including all 70 `label`
  parameters.

# Verification

* `pytest testsuite/pyunit` — **190 passed, 17 xfailed**, unchanged throughout.
* `black --check pyGHDL` — 71 files unchanged.
* AST checks: 0 classes, methods or functions without a doc-string in `pyGHDL.dom`; 0 `:param:` lists
  disagreeing with a signature.
* A local HTML build: the only `pyGHDL.dom` messages are 188 pre-existing
  `sphinx_autodoc_typehints` forward-reference warnings about type annotations.

### DIFF
--- a/doc/prolog.inc
+++ b/doc/prolog.inc
@@ -30,6 +30,7 @@
      span.colorgreen {color: #009933; }
      span.colorblue {color: #0066FF; }
      span.colorpurple {color: #9900CC; }
+     span.vhdlkw {font-style: italic; }
    </style>
 
 .. role:: bolditalic
@@ -57,6 +58,9 @@
    :class: colorred strike
 .. role:: addition
    :class: colorgreen
+
+.. role:: vhdlkw
+   :class: vhdlkw
 
 .. role:: pycode(code)
    :language: python
--- a/pyGHDL/cli/dom.py
+++ b/pyGHDL/cli/dom.py
@@ -57,6 +57,15 @@ __license__ = ""
 
 class SourceAttribute(Attribute):
     def __call__(self, func):
+        """
+        Attaches the source selection command line arguments to the decorated handler method.
+
+        The attribute bundles the four arguments that select what to analyze: ``-f``/``--file``,
+        ``-F``/``--files``, ``-D``/``--directory`` and ``-L``/``--library``.
+
+        :param func: The handler method to attach the command line arguments to.
+        :returns:    The same handler method, with the arguments attached.
+        """
         self._AppendAttribute(
             func,
             CommandLineArgument(
--- a/pyGHDL/cli/dom.py
+++ b/pyGHDL/cli/dom.py
@@ -60,8 +60,12 @@ class SourceAttribute(Attribute):
         """
         Attaches the source selection command line arguments to the decorated handler method.
 
-        The attribute bundles the four arguments that select what to analyze: ``-f``/``--file``,
-        ``-F``/``--files``, ``-D``/``--directory`` and ``-L``/``--library``.
+        The attribute bundles the four arguments that select what to analyze:
+
+        * ``-f`` / ``--file`` - the filename to parse, can be used multiple times.
+        * ``-F`` / ``--files`` - a list of filenames to parse.
+        * ``-D`` / ``--directory`` - the directory to parse.
+        * ``-L`` / ``--library`` - the default library for files in the root directory.
 
         :param func: The handler method to attach the command line arguments to.
         :returns:    The same handler method, with the arguments attached.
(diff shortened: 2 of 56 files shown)


Paebbels:
@tgingold pipeline is green for this pyGHDL documentation PR.

Paebbels:
@tgingold thanks for merging.

PR #3328, opened 2026-08-11
GHDL's AST documentation

Paebbels:
> [!NOTE]
> This branch is split off `paebbels/dom-updates` (ghdl/ghdl#3313) and only the changes **on top of
> it** are described here. The diff against `master` also carries #3313's content; please review that
> one there.

Documentation for GHDL's AST, and the generator work behind it. **15 commits, 12 files,
+8308 / -1253** — of which 8100 lines are regenerated bindings.

This finishes the narrative ot AST.rst, adds a generated reference for all **336**
node kinds, and fills the empty doc-strings in the generated Python bindings from the same source.

# `scripts/pnodes.py` keeps the comments

`pnodes.py` already parsed the whole meta-model out of `vhdl-nodes.ads` — kinds, formats, fields,
accessors, `Alias` and `Only for` qualifiers, and the access attributes — but discarded the prose.
Since the file documents its own comment grammar in its header, keeping it is a small addition to a
parser the Ada build already validates.

| | count |
|---|---|
| accessor descriptions | **219** of 395 |
| node descriptions | **99** |
| per-node field notes | **453** |

Two details of the format matter, both found by rendering the result rather than reading the parser:

* **A field's description directly precedes its `Get/Set_` line**, so the block after the *last* blank
  line under a kind header belongs to the first field, not to the node. Without that split,
  `Iir_Kind_Design_File` claimed *"The library containing this design file."* as its own description.
* **Only the standard two-space prefix is stripped**, so a line indented further keeps its relative
  indentation — otherwise the continuation lines of a multi-line grammar production collapse onto the
  margin and the production can no longer be recognised.

A field disabled by commenting out its `Get/Set_` line is recorded separately; otherwise its
description is attributed to the field that follows it, as happens today with
`Iir_Kind_Signal_Declaration`'s `Signal_Driver`.

`parse_files()` is the entry point for other tools, and `pnodes.py` itself is now fully documented
(11 of 51 entities → 51 of 51). The `get_format` action, which raised `TypeError` because
`do_get_format` called `gen_get_format` without `kinds`, is fixed.

# The empty doc-strings are filled

`pnodespy.py` emitted `""""""` for every generated accessor — a placeholder waiting for text that
`pnodes.py` was throwing away:

```py
def Get_Library_Declaration(obj: Iir) -> Iir:
    """
    Library declaration of a library clause.  This is Forward_Ref as the dependency of the unit on the
    library is not tracked.

    :param obj: The node to read the ``Library_Declaration`` field of.
    :returns:   The node's ``Library_Declaration`` field.
    """
```

All six generated files now have **0** entities without a doc-string, which is where the bulk of the
diff comes from. Things with no comment to draw on are written in the generator: the ten
enumerations, `Iir_Kinds`, `Get_Location`, the `Has_*` predicates, the `Get_<Type>` meta-model
accessors, `get_field_type`, `get_field_attribute`, `get_field_by_index`, `Enable_Warning`, and a
module doc-string for each file naming the Ada package it binds.

`interrogate` on `pyGHDL/libghdl`: **8.3 % → 96.5 %**; on `pyGHDL`: 11.1 % → **66.5 %**.

# The documentation

**Hand-written**, completing the narrative:

* **Node Type** — a node is a 32-bit table index, not a pointer. The two formats as a table: `Short`
  is 28 slots, `Medium` 37, and they differ *only* in the number of `Field` and `State` slots.
* **Chains, lists and flists** — the three ways nodes are linked, why a chain is an ownership
  relation (the link lives in the element, so a node is in at most one chain), and the matching
  iterators in `pyGHDL.libghdl.utils`.
* **Asking the meta-model** — the `Has_*` predicates, with `pyGHDL.dom._Utils.GetLabelOfNode` as the
  worked example of choosing between `Label` and `Alternative_Label` by asking rather than
  enumerating kinds.

**Generated** by `doc/ASTReference.py` when Sphinx starts, into a gitignored `doc/internals/ast/`, so
the reference cannot describe a node that no longer exists. 51 pages, 336 kinds, 331 diagrams.

Two things shaped it:

* **`Iir_Kinds_*` ranges are not a partition.** They are contiguous intervals that nest: 58 kinds are
  in no range at all and some are in six. The 29 *maximal* ranges become page boundaries, each kind
  is assigned to the first containing it, and kinds between two ranges keep their place as their own
  page — so the `Iir_Kind` order is preserved and every kind appears exactly once.
* **A diagram cannot name its targets.** 237 of the 395 accessors return a bare `Iir`, so the *kind* a
  field points at is not knowable from the source. The diagrams show which fields are links and of
  what nature: owned (black), reference (blue), forward reference resolved during semantic analysis
  (orange), chain (green). That last one is also the honest answer to "which fields are filled
  later" — the prose says so for about ten fields, but `Forward_Ref` marks it structurally for 23.

Comment prose is reflowed except runs containing `::=`, which stay literal so the grammar productions
survive, and the ReST-significant characters are escaped: an Ada comment is prose, not ReST, and
`Iir_Kind_Entity_Class` writes `` `<>' `` while `Iir_Kind_Constant_Declaration` draws a table with
`|`.

# Verification

* **No generated output changed by accident.** Every file both generators produce — the six Python
  and the six Ada ones — was regenerated and diffed; a full regeneration of the branch leaves the
  tree clean, so the *Check if Python bindings changed* step passes.
* `pytest testsuite/pyunit` — **190 passed, 17 xfailed** against the regenerated bindings, so they
  behave identically.
* `black --check pyGHDL` — 71 files unchanged; the regenerated files are formatted exactly as the
  Makefile does it.
* A local HTML build produces the 51 pages and 331 diagrams with no message from
  `doc/internals/ast/`.


Paebbels:
@tgingold can you please have a look at

1. [Internals &rarr; AST](https://paebbels.github.io/ghdl/internals/AST.html)  
   This continued your description how GHDL's AST internally works.
2. [Internals &rarr; AST &rarr; Subprogram Body](https://paebbels.github.io/ghdl/internals/ast/24-subprogram-body.html)  
   This is an example how Claude.ai proposes to document GHDL's IIR nodes.  
   The content is taken from `nodes.ads` via `pnodes.py`.

I'm not confident enough to judge if this is good and correct. You can comment and review in here and I can hand it over to Claude.ai for refinements.

My idea for visualizing the nodes was more like c-structs:

| My idea | Claude.ai | New rendering |
| --- | --- | --- |
| image | image | image |

Also I think we need a legend explaining arrow styles and arrow colors.

Paebbels:
The new graphviz rendering is better.

PR #3330, opened 2026-08-11
Documentation for pyGHDL.cli and pyGHDL.lsp

Paebbels:
Doc-strings for `pyGHDL.cli` and `pyGHDL.lsp`, the two packages of `pyGHDL` that had none worth speaking of.

This branch was split off `master`, so it is independent of #3327 and #3328 and touches none of the same files. It changes doc-strings and comments only — no signature, no behaviour, no dependency.

| Package | Before | After |
|---------|-------:|------:|
| `pyGHDL.cli` | 23.8 % | **100 %** |
| `pyGHDL.lsp` | 17.6 % | **100 %** |
| `pyGHDL` overall | 39.2 % | 44.6 % |

(`interrogate`; the overall figure moves only this far because `pyGHDL.dom` and `pyGHDL.libghdl` are documented in #3327 and #3328.)

## What is documented

**`pyGHDL.cli`** — the two console scripts registered by `setup.py`. `dom.py` is the one worth reading: its handlers say what each output is for, and the difference between analyzing a single file and a directory is stated where it matters.

**`pyGHDL.lsp`** — the language server, in the order a message travels through it:

- **`lsp.py`**, the transport. Messages are framed with a `Content-Length` header, which is what separates this from a plain JSON stream — the reader has to consume the header to know how many bytes the body has. Dispatch, the error replies and the URI normalization are described here; the last exists because clients disagree about capitalizing a Windows drive letter, and the same document arriving under two spellings would be looked up as two files.
- **`vhdl_ls.py`**, one method per request. Mostly one-line delegations, so the doc-strings carry what the delegation does not: `didChange` does not lint because `apply_changes` has already published, `didSave` compares the client's text against the server's buffer because a divergence there makes every position reported from then on wrong, and `rangeFormatting` re-lints because its own edit moves everything that was reported.
- **`document.py`**, the source buffer. Not a string: *libghdl* analyzes bytes it owns, so the buffer is over-allocated by a gap and edits are written into it in place. Extending it doubles the gap and moves the text, which invalidates the source file entry — and that is why `apply_changes` repairs the `sfe -> document` map afterwards. The conversion between the protocol's 0-based positions and *libghdl*'s 1-based lines is noted at each crossing.
- **`workspace.py`**, the project and the re-analysis. Documented below.
- **`references.py`** and **`symbols.py`**, resolution and the outline.

## The part worth reading twice

Editing one file cannot simply drop that file's units. Another file that used a package from it was analyzed against the tree being freed and holds references into it. Three methods handle this and now say so:

- `compute_anti_dependences` reverses the dependency graph. A unit records what it depends on; invalidation needs the opposite question answered.
- `obsolete_dependent_units` walks that reverse graph transitively and puts each unit back to `Disk` state. It writes the unit's position in the source into the node **before** freeing anything, because the tree is exactly what is about to disappear. The recursion is broken by clearing each entry as it is taken, so a cycle cannot loop forever.
- `obsolete_doc` ties the two together and purges the design file from its library.

`Workspace.__init__` describes the analysis flags rather than restating them: locations and comments are kept because the server answers questions about them, and analysis is forced past errors because a file being edited is incomplete most of the time.

In `references.py`, `find_def` matches a name by its own location plus the length of its identifier, because a node records where it *starts* but not where it ends. That is what the otherwise-puzzling operator groupings are — the width of the symbol that wrote them, one character for `+`, two for `**`, three for `and`, four for `nand`. Its descent reads the node metadata instead of walking children (much faster) and skips fields holding a reference, so a node is reached only through the field that owns it; that is also what keeps a cycle from looping forever.

## Incidental corrections

- Three doc-strings described an `lsp_*` naming convention for the request handlers. There is no such prefix, and the name is not what dispatch uses — `LanguageProtocolServer.handle` looks the incoming method up in `VhdlLanguageServer.dispatcher`, an explicit table. Corrected.
- Several doc-strings used `:param x:` inside prose (`Get a document from :param doc_uri:`), which is not what that role means. They have real field lists now.
- The class doc-string of `Document` sat between the encoding comment and the `encoding` attribute the comment describes. The comment is back on its attribute.
- Four class doc-strings ended in a blank line, which renders as a stray paragraph break.

## Verification

`pytest testsuite/pyunit` passes unchanged (190 passed, 17 xfailed, 12 subtests). `black --check` is clean over `pyGHDL`. The CI's `Python`, `DocCoverage` and `StaticTypeCheck` jobs are green.

Comparing the AST of every touched file against `master` with doc-strings stripped gives an identical dump for all of them, so no executable code changed. (`pyGHDL/cli/__init__.py` is the one apparent exception, and only because it had no module body at all before this branch gave it a doc-string.)

## Not in this PR

Four bugs turned up while reading the code. None is fixed here, since this PR is documentation:

- `Document.flatten_symbols` puts the parent *symbol* in `containerName`, which the protocol specifies as a `string`. The golden replies record the violation — `"containerName": {"kind": 2, "name": "rtl", ...}` where `"rtl"` belongs. Fixing it also updates four `replies.json` files.
- `Workspace.gather_diagnostics` reuses its `doc` parameter as a loop variable, so the "clear the diagnostics of a document that no longer has any" test at the end checks the wrong document. Fix the last error in one file while any other file still has one and the editor keeps showing the stale diagnostics.
- `Workspace._last_linted_doc` is set to `None` in `__init__` and never assigned again, so its branch in `obsolete_doc` is dead and `compute_anti_dependences` — which walks every unit of every library — runs on every keystroke.
- `m_workspace__did_change_configuration` and `m_workspace__did_change_watched_files` use the `m_<area>__<request>` naming of *python-language-server*, and neither request is a key of `dispatcher`. Both are unreachable, so changing a setting or touching a file outside the editor does not re-lint.

The last three are pre-existing on `master`; `_last_linted_doc` goes back to `bf74a0983` ("rework 'python', rename to 'pyGHDL'"), so the assignment was most likely lost in that move. The last two interact — wiring up the handlers makes two notifications live that each re-lint every document — so they are best done together. Happy to open them as issues or as a follow-up PR, whichever you prefer.

### DIFF
--- a/pyGHDL/cli/__init__.py
+++ b/pyGHDL/cli/__init__.py
@@ -34,6 +34,6 @@
 """
 Command line applications shipped with pyGHDL.
 
-:mod:`~pyGHDL.cli.dom` exercises :mod:`pyGHDL.dom` from a terminal, and :mod:`~pyGHDL.cli.lsp` starts the language
+:mod:`pyGHDL.cli.dom` exercises :mod:`pyGHDL.dom` from a terminal, and :mod:`pyGHDL.cli.lsp` starts the language
 server. Both are registered as console scripts by :file:`setup.py`.
 """
--- a/pyGHDL/cli/__init__.py
+++ b/pyGHDL/cli/__init__.py
@@ -31,3 +31,9 @@
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ============================================================================
+"""
+Command line applications shipped with pyGHDL.
+
+:mod:`~pyGHDL.cli.dom` exercises :mod:`pyGHDL.dom` from a terminal, and :mod:`~pyGHDL.cli.lsp` starts the language
+server. Both are registered as console scripts by :file:`setup.py`.
+"""
--- a/pyGHDL/cli/dom.py
+++ b/pyGHDL/cli/dom.py
@@ -36,7 +36,8 @@
 A terminal application to exercise :mod:`pyGHDL.dom`.
 
 It analyzes VHDL source files with *libghdl*, translates the result to the document object model and pretty-prints it,
-which makes it the quickest way to see what the model makes of a given file. Installed as ``pyghdl-dom``.
+which makes it the quickest way to see what the model makes of a given file. Installed as ``pyghdl-dom`` by
+:file:`setup.py`.
 """
 
 from argparse import RawDescriptionHelpFormatter
@@ -148,7 +149,7 @@ class Application(TerminalApplication, ArgParseHelperMixin):
 
     def __init__(self):
         """
-        Initializes the application with an empty design and the argument parser.
+        Initializes the application.
         """
         super().__init__()
 
--- a/pyGHDL/cli/dom.py
+++ b/pyGHDL/cli/dom.py
@@ -32,6 +32,13 @@
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ============================================================================
+"""
+A terminal application to exercise :mod:`pyGHDL.dom`.
+
+It analyzes VHDL source files with *libghdl*, translates the result to the document object model and pretty-prints it,
+which makes it the quickest way to see what the model makes of a given file. Installed as ``pyghdl-dom``.
+"""
+
 from argparse import RawDescriptionHelpFormatter
 from pathlib import Path
 from platform import system as platform_system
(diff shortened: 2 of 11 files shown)


Paebbels:
@tgingold this is also ready for documenting the LSP and CLI parts.

PR #3331, opened 2026-08-15
Bug fixes for pyGHDL.lsp

Paebbels:
# Breaking Changes

* ⚠️ `pyGHDL.lsp.document`: `containerName` in a `textDocument/documentSymbol` reply is a **string** - the name of
  the enclosing symbol - where it used to be the whole parent symbol object. A client that read `containerName.name`
  or any other member of that object has to read the value itself instead. See *Bug Fixes*: the old shape never
  conformed to the protocol.

# New Features

* `pyGHDL.lsp.vhdl_ls`:
  * `VhdlLanguageServer.initialize` returns `serverInfo` beside the capabilities:
    `{"name": "ghdl-ls", "version": <pyGHDL version>}`. The reply carried the capabilities alone, so a client
    could neither log which server it had started nor vary its behaviour by server version. `serverInfo` is part
    of the protocol since LSP 3.15.

# Changes

* `pyGHDL.lsp.workspace`:
  * `Workspace.obsolete_doc` skips computing the anti-dependencies when the document being obsoleted is also the
    one analyzed last. That fast path already existed and never ran, because `Workspace._last_linted_doc` was
    never assigned - `git log -S` finds one commit touching it, the rename of `python/` to `pyGHDL/`. It is
    assigned at the end of `Workspace.lint` and `Workspace.apply_changes` again, which removes a walk over every
    unit from the hot path while a file is being typed in.

    Why this is safe is written into the doc-string of `obsolete_doc`: semantic analysis happens in `lint` and
    `apply_changes` alone, so a dependent obsoleted by the previous round can only have come back through an
    analysis of its own - and that would have made *it* the last analyzed document, leaving the guard false.

    **This is the one behavioral change here rather than a repair, and it is a separate commit (`b14b9ff`), so
    it can be dropped on its own if you would rather keep the unconditional walk.**

# Bug Fixes

* `pyGHDL.lsp.document`:
  * ⚠️ `Document.flatten_symbols` sent the whole parent symbol as `containerName` - a dictionary with `kind`,
    `name` and `location` - while
    [`SymbolInformation.containerName`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#symbolInformation)
    is declared `string`. The reply never conformed to the protocol, so a client rendering the flat symbol shape
    displayed an object where a name belongs. It is the enclosing symbol's name now. Three golden reply files had
    the object baked in.
* `pyGHDL.lsp.workspace`:
  * `Workspace.gather_diagnostics` never withdrew the diagnostics of a file whose errors had been fixed, as long
    as any other file still reported one. The method reassigned its own `doc` parameter twice - once when
    attaching related information to a message, and again in the publishing loop:

    ```python
    for sfe, diag in diags.items():
        doc = self.sfe_to_document(sfe)          # the parameter is overwritten
        self.publish_diagnostics(doc.uri, diag)
    if doc is not None and doc._fe not in diags: # so this tests the wrong document
        self.publish_diagnostics(doc.uri, [])
    ```

    By the time the trailing check runs, `doc` is the last file that *had* diagnostics, whose `_fe` is by
    construction in `diags`. The condition is therefore false whenever anything was reported, and no empty list is
    published. A client keeps diagnostics until it is sent new ones, so the corrected errors stay on screen.

    The documents diagnostics were published for are now tracked in `Workspace._diags_set` - a field that was
    created for this and never filled - and every document in it that reports nothing in the current round is sent
    an empty list. That also covers a case the original code did not attempt: a diagnostic disappearing from a file
    other than the one that was analyzed. `Workspace.rm_document` discards its URI from that set, since closing a
    document withdraws its diagnostics there already.

# Known Issues

* `pyGHDL.lsp.vhdl_ls`:
  * `VhdlLanguageServer.dispatcher` still has no entry for two implemented handlers, so those notifications remain
    unreachable. Wiring them changes what the server acts on, which is a decision rather than a repair, so it is
    left out of this pull-request.
* `pyGHDL.lsp`:
  * The classes in this package derive from `object` and carry no type hints, unlike the rest of `pyGHDL`. Not
    touched here - it would bury the four fixes above in a formatting diff.

# Documentation

* The doc-strings of `Workspace.gather_diagnostics`, `Workspace.obsolete_doc` and `Document.flatten_symbols` state
  the rules the code now follows: which documents get an empty list and why, the invariant behind the fast path,
  and that `containerName` is a name. No user-facing documentation covers these internals.

# Unit Tests

* `testsuite/pyunit/lsp/Diagnostics.py` is new: six testcases for what `gather_diagnostics` publishes and
  withdraws. They script the error memory instead of driving *libghdl*, which is what makes the multi-file cases
  reachable at all - the JSON harness can only express what one server run over one directory produces.
  **Three of the six fail without this pull-request**: the analyzed document not being cleared while another file
  has diagnostics, a document that stopped reporting never being cleared, and the related-information
  reassignment.
* The 13 `testsuite/pyunit/lsp/*/replies.json` files gain `serverInfo`; three of them carry the corrected
  `containerName`. They were edited in place rather than regenerated from a server run, so the diff shows the
  behavioural change and not a reordering of keys.

----------
# Related Issues and Pull-Requests

* Reviewed and merged in the fork as [Paebbels/ghdl#48](https://github.com/Paebbels/ghdl/pull/48).
* Independent of [#2936](https://github.com/ghdl/ghdl/pull/2936) (`paebbels/lsp`), which touches the testsuite
  examples and the LSP test driver, not `pyGHDL/lsp/*` and not the golden replies.

### DIFF
--- a/pyGHDL/lsp/document.py
+++ b/pyGHDL/lsp/document.py
@@ -309,6 +309,8 @@ class Document(object):
         The two shapes differ in more than nesting: a location replaces the range, ``detail`` has no counterpart,
         and the nesting is expressed by naming the container. A client that asked for the flat form gets it here.
 
+        ``containerName`` is the *name* of the enclosing symbol, as ``SymbolInformation`` declares it a string.
+
         :param syms:   The symbols to flatten. They are modified in place.
         :param parent: The symbol the given ones are nested in, or ``None`` at the top level.
         :returns:      The symbols and all their children, in one list.
@@ -319,7 +321,7 @@ class Document(object):
             del s["range"]
             s.pop("detail", None)
             if parent is not None:
-                s["containerName"] = parent
+                s["containerName"] = parent["name"]
             res.append(s)
             children = s.pop("children", None)
             if children is not None:
--- a/pyGHDL/lsp/vhdl_ls.py
+++ b/pyGHDL/lsp/vhdl_ls.py
@@ -41,6 +41,8 @@ itself.
 
 import logging
 
+from pyGHDL import __version__ as pyGHDLVersion
+
 from . import lsp
 from .workspace import Workspace
 
@@ -178,7 +180,7 @@ class VhdlLanguageServer(object):
         :param rootUri:               The root as a URI.
         :param initializationOptions: Options from the client. They are logged, not read.
         :param _:                     Further members of the request, ignored.
-        :returns:                     The capabilities of this server.
+        :returns:                     The capabilities of this server, and the name and version it runs under.
         :raises InitError:            If *libghdl* could not be initialized.
         """
         log.debug(
@@ -193,7 +195,10 @@ class VhdlLanguageServer(object):
         self.workspace = Workspace(rootUri, self.lsp)
 
         # Get our capabilities
-        return {"capabilities": self.capabilities()}
+        return {
+            "capabilities": self.capabilities(),
+            "serverInfo": {"name": "ghdl-ls", "version": pyGHDLVersion},
+        }
 
     def initialized(self):
         """
--- a/pyGHDL/lsp/workspace.py
+++ b/pyGHDL/lsp/workspace.py
@@ -502,6 +502,11 @@ class Workspace(object):
         into the tree being freed, so those units are obsoleted too. The design file is then purged from its
         library, which leaves the document ready to be parsed again.
 
+        Obsoleting the dependent units is skipped when this document was also the one analyzed last, as it is while
+        a file is being edited: they were obsoleted by that previous round and only an analysis of their own could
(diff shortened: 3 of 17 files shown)


Paebbels:
@tgingold when documenting `pyGHDL.lsp` these issues where identified and proposed as fixes.

I'm not into LSP, so I let you judge. If you have remarks just write (code-review) comments and Claude.ai can pick it up.

tgingold:
it should be split: adding the version in the baselines means the tests would fail when the version is updated.
I would prefer not to add the version.

Paebbels:
So I should find a way in testing, so the version isn't hard coded in the expected data, right?

Issue #3332, opened 2026-08-16
GHDL VPI: array of unconstrained `std_logic_vector` port: `vpi_handle_by_index` silently returns NULL


dalex78:
**Description**
I want to retrieve the value of a port which is an array of the `std_logic_vector` in C through VPI.

`mwe.vhd` declares one array-of-unconstrained-vector type in a package:

```vhdl
type t_slv_vector is array (natural range <>) of std_logic_vector;
```

Entity `mwe` has a single port `arr_i : in t_slv_vector(0 to 1)(3 downto 0)`.
`mwe_vpi.cpp` resolves the whole-array port by name, then calls
`vpi_handle_by_index(arr, i)` for `i in 0, 1`, probing `vpi_get(vpiSize, ...)`
and `vpi_get_value` on any handle it gets back.

**Expected behaviour**
Get whatever the initial value of the port is:

```
VPI module loaded!
[MWE] arr_i(0): handle OK, vpiSize=4, read = "0101"
[MWE] arr_i(1): handle OK, vpiSize=4, read = "1010"
```

**Failure signature**
```
VPI module loaded!
[MWE] arr_i(0): vpi_handle_by_index returned NULL
[MWE] arr_i(1): vpi_handle_by_index returned NULL
```

**How to reproduce?**
Here is the Minimal Working Example (MWE). You can also download the files provided including the run.sh scrip to run it.

[mwe.vhd.txt](https://github.com/user-attachments/files/31117279/mwe.vhd.txt)
[mwe_vpi.cpp](https://github.com/user-attachments/files/31117272/mwe_vpi.cpp)
[run.sh](https://github.com/user-attachments/files/31117273/run.sh)

```vhd :file: mwe.vhd
library ieee;
use ieee.std_logic_1164.all;

package mwe_pkg is
    -- Unresolved element, unconstrained at the type-declaration site -- only
    -- the object's own elaborated instance gives it a range.
    type t_sulv_vector is array (natural range <>) of std_ulogic_vector;
end package mwe_pkg;

library ieee;
use ieee.std_logic_1164.all;
use work.mwe_pkg.all;

entity mwe is
    port (
        arr_i : in t_sulv_vector(0 to 1)(3 downto 0) := ("0101", "1010")
    );
end entity mwe;

architecture rtl of mwe is
begin
end architecture rtl;
```

```vhd :file: mwe_vpi.cpp
#include <stdio.h>

#include "vpi_user.h"

[...]

PR #3333, opened 2026-08-16
grt-avhpi: fix VPI silent NULL on array of unconstrained std_logic_ve…

dalex78:
Fixes issue https://github.com/ghdl/ghdl/issues/3332

The patch was generated by an AI agent. I cannot provide any guarantee that it is the correct fix. I'll let the maintainers be the judge of that. Feel free to edit, copy or change anything that you deem necessary.

I used a AI agent to investigate the bug, here is his analysis:

> GHDL's VHPI layer (`src/grt/grt-avhpi.adb`) has a lookup table it uses to turn an internal type descriptor into a VPI handle. For an array whose element type is itself an unconstrained, resolved vector (`std_logic_vector` with no range, only given one per elaborated instance), that lookup table is missing an entry. So when `vpi_handle_by_index` tries to resolve the element's subtype, the lookup fails, and the function gives up early and returns NULL — silently, with no error message. Code that doesn't check for NULL just ends up reading or driving nothing.

### DIFF
--- a/src/grt/grt-avhpi.adb
+++ b/src/grt/grt-avhpi.adb
@@ -131,6 +131,7 @@ package body Grt.Avhpi is
                           N_Addr => Avhpi_Get_Address (Ref),
                           N_Type => Ref.Obj.Obj_Type,
                           N_Idx => 0,
+                          N_Bounds => Null_Address,
                           N_Obj => Ref.Obj);
                when VhpiIndexedNameK =>
                   Res := (Kind => AvhpiNameIteratorK,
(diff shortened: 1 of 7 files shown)


dalex78:
One issue with that MR is that it will conflict with https://github.com/ghdl/ghdl/pull/3335 on several points (comments, some similar part of the code added, etc).

To avoid those issue I made a third MR combining the two MRs. I also removed most of the comments that the AI agent adds.

https://github.com/ghdl/ghdl/pull/3336

dalex78:
Not sure what the failing job has to do with this MR.

PR #3335, opened 2026-08-16
grt-avhpi: fix VPI crash on array of unconstrained std_ulogic_vector

dalex78:
Fixes issue https://github.com/ghdl/ghdl/issues/3334

The patch was generated by an AI agent. I cannot provide any guarantee that it is the correct fix. I'll let the maintainers be the judge of that. Feel free to edit, copy or change anything that you deem necessary.

I used a AI agent to investigate the bug, here is his analysis:

> Once GHDL's VHPI layer (`src/grt/grt-avhpi.adb`) has resolved the array element's type, it computes the byte offset of a given index by looking at the element's type kind. It knows how to do this for a handful of type kinds (plain scalars, and elements with their own fixed range), but `std_ulogic_vector` used as an element with no range at all is a kind it has never seen before — `std_logic_vector` in the same position at least gets its own distinct type record, but an unresolved `std_ulogic_vector` doesn't even get that. Since none of the known cases match, it falls into a catch-all branch that treats this as an internal inconsistency and aborts the whole simulator process, rather than just failing that one lookup.

### DIFF
--- a/src/grt/grt-avhpi.adb
+++ b/src/grt/grt-avhpi.adb
@@ -131,6 +131,7 @@ package body Grt.Avhpi is
                           N_Addr => Avhpi_Get_Address (Ref),
                           N_Type => Ref.Obj.Obj_Type,
                           N_Idx => 0,
+                          N_Bounds => Null_Address,
                           N_Obj => Ref.Obj);
                when VhpiIndexedNameK =>
                   Res := (Kind => AvhpiNameIteratorK,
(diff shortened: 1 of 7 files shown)


dalex78:
One issue with that MR is that it will conflict with https://github.com/ghdl/ghdl/pull/3333 on several points (comments, some similar part of the code added, etc).

To avoid those issue I made a third MR combining the two MRs. I also removed most of the comments that the AI agent adds.

https://github.com/ghdl/ghdl/pull/3336

Issue #1395, opened 2020-07-16
GHDL able to define a string with a range outside type "string"  as in  (4 downto 0)


sckoarn:
**Description**
GHDL is able to process a string definition which is outside the predefined range of string  (natural range <>)  which does not include 0

**Expected behaviour**
In the case of the EDA Playground example below, it should fail at run time when a dynamic string with range that includes 0 is attempted.

**How to reproduce?**
https://www.edaplayground.com/x/5smV

> NOTE: `:file:` and `:image:` identifiers are specific to [issue-runner](https://github.com/1138-4EB/issue-runner). We suggest to use these, since it allows continuous integration workflows to automatically test the MWE. Using `ghdl/ghdl:*` docker images to run the MWEs ensures that the latest available GHDL is used.

> NOTE: Large files can be uploaded one-by-one or in a tarball/zipfile.

**Context**
Please, provide the following information:

- OS:
- Origin:
  - [ ] Package manager: `version`
  - [ ] Released binaries: `tarball_url`
  - [ ] Built from sources: `commit SHA`
 - *  EDA Playground

If a `GHDL Bug occurred` block is shown in the log, please paste it here:

```
******************** GHDL Bug occurred ***************************
Please report this bug on https://github.com/ghdl/ghdl/issues
...
******************************************************************
```

**Additional context**
Add any other context about the problem here. If applicable, add screenshots to help explain your problem.

eine:
FTR:

```vhd
entity ent is
end;

architecture arch of ent is
  signal strt: string(0 to 4);
  signal strd: string(4 downto 0);
begin
process
  variable strvt: string(0 to 4);
  variable strvd: string(4 downto 0);
begin
report "HELLO" severity note;
wait;
end process;
end;
```

```vhd
# ghdl -a test.vhd 
test.vhd:5:23:warning: static range violates bounds [-Wruntime-error]
test.vhd:5:23:warning: static range violates bounds [-Wruntime-error]
test.vhd:6:23:warning: static range violates bounds [-Wruntime-error]
test.vhd:6:23:warning: static range violates bounds [-Wruntime-error]
test.vhd:9:26:warning: static range violates bounds [-Wruntime-error]
test.vhd:9:26:warning: static range violates bounds [-Wruntime-error]
test.vhd:10:26:warning: static range violates bounds [-Wruntime-error]
test.vhd:10:26:warning: static range violates bounds [-Wruntime-error]
```

---

```vhd
entity ent is
end;

[...]

Issue #562, opened 2018-04-26
Misleading elab error


kraigher:
**ent.vhd**
```vhdl
entity ent is
  generic (
    max : natural);
  port (
    p : out natural range 1 to max := 3);
end entity;

architecture a of ent is
begin
end;
```

**ent2.vhd**
```vhdl
entity ent2 is
end entity;

architecture a of ent2 is
  constant max : natural := 2;
  signal p : natural range 1 to max;
begin

  inst : entity work.ent
    generic map (max => max)
    port map (p => p);
end;

```

**shell**
```
+ ghdl --version
GHDL 0.36-dev (v0.35-139-g0edf0a10) [Dunoon edition]
 Compiled with GNAT Version: 7.3.1 20180303 (Red Hat 7.3.1-5)
 mcode code generator
Written by Tristan Gingold.

Copyright (C) 2003 - 2015 Tristan Gingold.
GHDL is free software, covered by the GNU General Public License.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ ghdl -a ent.vhd
+ ghdl -a ent2.vhd
+ ghdl --elab-run ent2
ghdl:error: bound check failure at ent2.vhd:5
  from: work.ent2(a).inst.CMP_ELAB at ent2.vhd:5
ghdl:error: error during elaboration
```

The error is reported as occuring on ent2.vhd:5 but should be reported on ent.vhd:5

tgingold:
I can reproduce the issue.

kraigher:
I do not agree on the classification as "enhancement". The reproducer can be misleading regarding the error since ent2.vhd:5 seems related to ent.vhd:5. This is however just coincidence. If I insert 10 newlines above line 5 in ent2.vhd it will still complain about ent2.vhd:5 which is then only a empty line. Clearly the code has mixed up the file name and line numbers which I would classify as a bug.

OmarElfouly:
hello @tgingold and @Paebbels 
I tried to replicate this issue and it seems to have resolve itself as the error it reports is now:

```vhd
ent.vhd:5:39:error: value out of range
    p : out natural range 1 to max := 3);
                                      ^
```

Full terminal:
```
root@8df63971cf2d:/usr/testingElab# ghdl --version
GHDL 5.0.0-dev (4.1.0.r271.gec95e9130) [Dunoon edition]
 Compiled with GNAT Version: 13.2.0
[...]

PR #3346, opened 2026-08-17
fix llvm/gcc backend accepting 0-inclusive STRING index bounds

dalex78:
https://github.com/ghdl/ghdl/issues/1395


dalex78:
The fail of `issue562` is unrelated to that MR: see https://github.com/ghdl/ghdl/issues/562#issuecomment-5320525397

However, one of the tests added by that MR fails. I'll investiguate.

dalex78:
@tgingold is this one good enough ?

tgingold:
It's OK but I would like to rework it a little bit: create a separate subprogram for the check, try to reproduce the error in a different case.  i will come back on it later.

dalex78:
Another follow up PR which might address what you wanted to do @tgingold : https://github.com/ghdl/ghdl/pull/3380

PR #3348, opened 2026-08-17
vhdl-sem_names: guard Name_Staticness lookup on ELEMENTs prefix

### LINKED ISSUE #2245: ghdl Crash:  access type and  'base
**Description**
While testing  attribute  'base it was found that the code included causes  ghdl to  crash on analysis.
This code runs on  edaplayground under Aldec with no issues.

**Expected behaviour**
Messages should be output.   -(int_max)   message  with  22  and message with 23

**How to reproduce?**
Run the code provided.

```use std.env.all;

entity unc_tbase3 is
end entity;

architecture test of unc_tbase3 is

  type mx_ivec is array (natural range <>) of integer;
  type mx_ivec_p is access mx_ivec;
  subtype ints8 is mx_ivec(7 downto 0);
  
begin

  fill_check: process
    variable ivec : mx_ivec_p;
  begin
    
    ivec := new ints8;
    
    report "" & integer'image(ints8'base'element'left);
    report "" & integer'image(ints8'base'element'pos(22));
    report "" & integer'image(ints8'base'element'succ(22));
    
    finish(0);    
  end process fill_check;

end test;
```

```sh :image: ghdl/ghdl:bullseye-mcode
ghdl -a ent.vhd
ghdl --elab-run ent
```

> NOTE: `:file:` and `:image:` identifiers are specific to [issue-runner](https://github.com/1138-4EB/issue-runner). We suggest to use these, since it allows continuous integration workflows to automatically test the MWE. Using `ghdl/ghdl:*` docker images to run the MWEs ensures that the latest available GHDL is used.

> NOTE: Large files can be uploaded one-by-one or in a tarball/zipfile.

**Context**

- OS:
- Origin:

If a `GHDL Bug occurred` block is shown in the log, please paste it here:

```
predef_16_08/ghdl_tmp$ ghdl -a --std=08 test.vhdl 

GHDL release: 3.0.0-dev (2.0.0.r483.g1668538d0) [Dunoon edition]
Compiled with GNAT Version: 8.3.0
Target: x86_64-linux-gnu
/project/dsim/kcampbell/superp/test/dsim/vh_tests/predef_16_08/ghdl_tmp/
Command line:
ghdl -a --std=08 test.vhdl
Exception SYSTEM.ASSERTIONS.ASSERT_FAILURE raised
Exception information:
raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : no field Base_Name
Call stack traceback locations:
0x7f6f222094f1 0x55a978d5c7d3 0x55a978e46b63 0x55a978e485ad 0x55a978e487f6 0x55a978e48244 0x55a978e487f6 0x55a978e44415 0x55a978e487d8 0x55a978e74792 0x55a978e6a0cb 0x55a978e6a2a2 0x55a978e6a8fd 0x55a978e73e28 0x55a978e748a1 0x55a978e75b41 0x55a978e39484 0x55a978e394c7 0x55a978e3b0c1 0x55a978e3b245 0x55a978e3be80 0x55a978e3c6da 0x55a978e3c930 0x55a978e3ca8f 0x55a978e5ea5b 0x55a978e65f48 0x55a978e3244c 0x55a978f15e9d 0x55a978e25698 0x55a978fdd818 0x55a978bbece3 0x7f6f21df2099 0x55a978bbd298 0xfffffffffffffffe
```

**Additional context**
Add any other context about the problem here. If applicable, add screenshots to help explain your problem.

dalex78:
Fixes https://github.com/ghdl/ghdl/issues/2245

### DIFF
--- a/src/synth/elab-vhdl_expr.adb
+++ b/src/synth/elab-vhdl_expr.adb
@@ -1065,12 +1065,18 @@ package body Elab.Vhdl_Expr is
          when Iir_Kind_Subtype_Attribute =>
             return Exec_Name_Subtype (Syn_Inst, Get_Prefix (Name));
 
+         when Iir_Kind_Base_Attribute =>
+            --  Unlike 'SUBTYPE, this is the base type of the prefix and not
+            --  the prefix's own (possibly constrained) subtype.
+            return Get_Subtype_Object (Syn_Inst, Get_Type (Name));
+
          when Iir_Kind_Element_Attribute =>
             declare
                Pfx : Type_Acc;
             begin
                Pfx := Exec_Name_Subtype (Syn_Inst, Get_Prefix (Name));
-               return Pfx.Arr_El;
+               --  The prefix is unbounded for T'BASE'ELEMENT.
+               return Get_Array_Element (Pfx);
             end;
 
          when others => Error_Kind ("exec_name_subtype", Name);
--- a/src/synth/elab-vhdl_types.adb
+++ b/src/synth/elab-vhdl_types.adb
@@ -902,8 +902,8 @@ package body Elab.Vhdl_Types is
             begin
                T := Synth_Array_Attribute_Prefix (Syn_Inst, Atype2);
                pragma Assert (T.Is_Global);
-               --  Always a bounded array/vector.
-               return T.Arr_El;
+               --  A bounded array/vector, except for T'BASE'ELEMENT.
+               return Get_Array_Element (T);
             end;
          when Iir_Kind_Enumeration_Type_Definition
            | Iir_Kind_Integer_Type_Definition
--- a/src/vhdl/vhdl-nodes.ads
+++ b/src/vhdl/vhdl-nodes.ads
@@ -4912,6 +4912,12 @@ package Vhdl.Nodes is
    --   Get/Set_Prefix (Field0)
    --
    --   Get/Set_Type (Field1)
+   --
+   --   Get/Set_Base_Name (Field5)
+   --
+   --   Get/Set_Type_Staticness (State1)
+   --
+   --   Get/Set_Name_Staticness (State2)
 
    -- Iir_Kind_Across_Attribute (Short)
    -- Iir_Kind_Through_Attribute (Short)
--- a/src/vhdl/vhdl-nodes_meta.adb
+++ b/src/vhdl/vhdl-nodes_meta.adb
@@ -5233,8 +5233,11 @@ package body Vhdl.Nodes_Meta is
       Field_Pathname_Expression,
       --  Iir_Kind_Box_Name
       --  Iir_Kind_Base_Attribute
+      Field_Type_Staticness,
+      Field_Name_Staticness,
       Field_Prefix,
       Field_Type,
+      Field_Base_Name,
       --  Iir_Kind_Subtype_Attribute
       Field_Type_Staticness,
       Field_Name_Staticness,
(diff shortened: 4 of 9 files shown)


dalex78:
The fail of `issue562` in the CI is unrelated to that MR: see https://github.com/ghdl/ghdl/issues/562#issuecomment-5320525397

dalex78:
Rebased on main where the failing test where deactivated. 
See: https://github.com/ghdl/ghdl/commit/230525db38c21580747ca453081164b032a415ab

tgingold:
I don't think this is the correct fix.  Name_Staticness field should be added on the nodes.

dalex78:
# Issue #2245 — `'BASE'ELEMENT` attribute chain

Reworked fix, replacing the one tgingold turned down on PR #3348.

## What the maintainer asked for

> I don't think this is the correct fix. Name_Staticness field should be added on the nodes.

## What is actually wrong in GHDL

Two independent defects sit on the same path, and the design in the report needs both fixed.

**1. `Iir_Kind_Base_Attribute` was missing fields it needs.** It carried only `Prefix` and `Type`, while its two siblings — `Iir_Kind_Subtype_Attribute` and `Iir_Kind_Element_Attribute`, which likewise denote a type and can be the prefix of another name — carry `Base_Name (Field5)`, `Type_Staticness (State1)` and `Name_Staticness (State2)`. `Finish_Sem_Name_1`, finishing `ints8'base'element`, does `Set_Name_Staticness (Res, Get_Name_Staticness (Get_Prefix (Res)))`, and with a `'BASE` node as that prefix this raised `ASSERT_FAILURE : no field Name_Staticness`.

**2. `T'BASE` was routed through `Sem_Type_Mark` when it is an array attribute's prefix.** LRM08 16.2 says of `T'BASE`: *"This attribute is allowed only as the prefix of the name of another attribute; for example, T'BASE'LEFT."* So `ints8'base'element` is legal, and `'Base attribute cannot be used as a type mark` was simply the wrong diagnostic. `Sem_Predefined_Type_Attribute` already knows this — it special-cases `Iir_Kind_Base_Attribute` and skips `Sem_Type_Mark`, which is why `integer'base'left` has always worked. `Finish_Sem_Array_Attribute_Prefix`, the equivalent entry point for every array-kind attribute (`'LEFT`, `'RIGHT`, `'HIGH`, `'LOW`, `'RANGE`, `'REVERSE_RANGE`, `'LENGTH`, `'ASCENDING`, `'ELEMENT`), had no such case and called `Sem_Type_Mark` on the `'BASE` prefix.

## The fix

`src/vhdl/vhdl-nodes.ads` — add `Base_Name (Field5)`, `Type_Staticness (State1)` and `Name_Staticness (State2)` to `Iir_Kind_Base_Attribute`, matching `Iir_Kind_Subtype_Attribute` / `Iir_Kind_Element_Attribute`. `src/vhdl/vhdl-nodes_meta.adb` is regenerated by `make -C src/vhdl` (the Python bindings under `pyGHDL/` do not encode per-kind field lists and are unchanged).

`Sem_Base_Attribute` — set the three new fields, mirroring `Sem_Subtype_Attribute`.

`Finish_Sem_Array_Attribute_Prefix` — when the prefix denotes a `'BASE` attribute, finish it with `Finish_Sem_Name` instead of `Sem_Type_Mark`, exactly as `Sem_Predefined_Type_Attribute` already does.

## A second, pre-existing crash found on the way

`Sem_Type_Mark` reported the `'Base attribute cannot be used as a type mark` error and then let the `Base_Attribute` node flow on as a normal result, so a genuine misuse such as

```vhdl
signal s : ints8'base;
```

printed the correct error and then died with `get_type_of_subtype_indication: cannot handle IIR_KIND_BASE_ATTRIBUTE`. Reproduced unchanged on GHDL 1.0.0 (Debian) and on a plain master build, so it long predates this issue; it is unrelated to the `'ELEMENT` path but one line away from it, so it is fixed here too by returning `Create_Error_Type` after the diagnostic. Say so at review time — it is easy to split out if tgingold would rather have it separate.

## Result

`testsuite/gna/issue2245/test.vhdl` is the report's design, unchanged. It now analyzes, elaborates and runs:

```
test.vhdl:20:5:@0ms:(report note): -2147483648
test.vhdl:21:5:@0ms:(report note): 22
test.vhdl:22:5:@0ms:(report note): 23
simulation finished @0ms with status 0
```

which matches what the reporter got from Aldec. `err.vhdl` checks the other direction: `'BASE` in a real type mark position is still rejected, and no longer crashes afterwards.

Diagnostics that must stay were checked by hand and all still hold, with no internal error: `ints8'base'left` → `prefix type is not constrained` (the base type of an array subtype is unconstrained, so `'LEFT` really is undefined on it — the LRM's `T'BASE'LEFT` example is about scalar `T`); `s'base'…`, `work'base'…`, `f'base'…` → `a type mark must denote a type or a subtype`; `nosuch'base'…` → `no declaration for "nosuch"`; `integer'base` alone → `'base attribute not allowed in an expression`. `integer'base'base'left` and `s8'base'element'element` resolve correctly.

dalex78:
If that fix is not good enough. I'll guess I'll just drop that attempt.

dalex78:
# Issue #2245 — `'BASE'ELEMENT` attribute chain

Reworked fix, replacing the one tgingold turned down on PR #3348.

## What the maintainer asked for

> I don't think this is the correct fix. Name_Staticness field should be added on the nodes.

## What is actually wrong in GHDL

Two independent defects sit on the same path, and the design in the report needs both fixed.

**1. `Iir_Kind_Base_Attribute` was missing fields it needs.** It carried only `Prefix` and `Type`, while its two siblings — `Iir_Kind_Subtype_Attribute` and `Iir_Kind_Element_Attribute`, which likewise denote a type and can be the prefix of another name — carry `Base_Name (Field5)`, `Type_Staticness (State1)` and `Name_Staticness (State2)`. `Finish_Sem_Name_1`, finishing `ints8'base'element`, does `Set_Name_Staticness (Res, Get_Name_Staticness (Get_Prefix (Res)))`, and with a `'BASE` node as that prefix this raised `ASSERT_FAILURE : no field Name_Staticness`.

**2. `T'BASE` was routed through `Sem_Type_Mark` when it is an array attribute's prefix.** LRM08 16.2 says of `T'BASE`: *"This attribute is allowed only as the prefix of the name of another attribute; for example, T'BASE'LEFT."* So `ints8'base'element` is legal, and `'Base attribute cannot be used as a type mark` was simply the wrong diagnostic. `Sem_Predefined_Type_Attribute` already knows this — it special-cases `Iir_Kind_Base_Attribute` and skips `Sem_Type_Mark`, which is why `integer'base'left` has always worked. `Finish_Sem_Array_Attribute_Prefix`, the equivalent entry point for every array-kind attribute (`'LEFT`, `'RIGHT`, `'HIGH`, `'LOW`, `'RANGE`, `'REVERSE_RANGE`, `'LENGTH`, `'ASCENDING`, `'ELEMENT`), had no such case and called `Sem_Type_Mark` on the `'BASE` prefix.

## The fix

`src/vhdl/vhdl-nodes.ads` — add `Base_Name (Field5)`, `Type_Staticness (State1)` and `Name_Staticness (State2)` to `Iir_Kind_Base_Attribute`, matching `Iir_Kind_Subtype_Attribute` / `Iir_Kind_Element_Attribute`. `src/vhdl/vhdl-nodes_meta.adb` is regenerated by `make -C src/vhdl` (the Python bindings under `pyGHDL/` do not encode per-kind field lists and are unchanged).

`Sem_Base_Attribute` — set the three new fields, mirroring `Sem_Subtype_Attribute`.

`Finish_Sem_Array_Attribute_Prefix` — when the prefix denotes a `'BASE` attribute, finish it with `Finish_Sem_Name` instead of `Sem_Type_Mark`, exactly as `Sem_Predefined_Type_Attribute` already does.

## A third crash, exposed by the fix itself

Once `ints8'base'element` analyzes, it is a legal type mark, so

```vhdl
variable v : ints8'base'element;
```

is legal too — and it crashed the **mcode** elaborator with `exec_name_subtype: cannot handle IIR_KIND_BASE_ATTRIBUTE`. The gcc and llvm backends ran it correctly, which is exactly the kind of silent divergence that makes checking all three worthwhile: testing only those two would have shipped the crash.

`Exec_Name_Subtype` (`src/synth/elab-vhdl_expr.adb`, shared by the mcode elaborator and by synthesis) had no `Iir_Kind_Base_Attribute` case. Adding one exposed a second layer: `T'BASE` yields an *unbounded* array, and both `'ELEMENT` consumers read `.Arr_El`, which only exists on the bounded variant of `Type_Acc`, so it turned into `discriminant check failed`. Both now use the existing `Get_Array_Element` helper, which handles either. `ghdl --synth` was checked on the same construct, since that code is shared.

## A second, pre-existing crash found on the way

`Sem_Type_Mark` reported the `'Base attribute cannot be used as a type mark` error and then let the `Base_Attribute` node flow on as a normal result, so a genuine misuse such as

```vhdl
signal s : ints8'base;
```

printed the correct error and then died with `get_type_of_subtype_indication: cannot handle IIR_KIND_BASE_ATTRIBUTE`. Reproduced unchanged on GHDL 1.0.0 (Debian) and on a plain master build, so it long predates this issue; it is unrelated to the `'ELEMENT` path but one line away from it, so it is fixed here too by returning `Create_Error_Type` after the diagnostic. Say so at review time — it is easy to split out if tgingold would rather have it separate.

## Result

`testsuite/gna/issue2245/test.vhdl` is the report's design, unchanged. It now analyzes, elaborates and runs:

```
test.vhdl:20:5:@0ms:(report note): -2147483648
test.vhdl:21:5:@0ms:(report note): 22
test.vhdl:22:5:@0ms:(report note): 23
simulation finished @0ms with status 0
```

which matches what the reporter got from Aldec.

`pos.vhdl` covers the neighbouring legal forms, each asserting a distinct value: `integer'base'high` (the LRM's own scalar example), `ints8'element` as a no-`'BASE` control, `ints8'base'element`, nested `mat2'base'element'element`, chained `integer'base'base`, and a variable declared as `ints8'base'element` — the case that crashed the mcode elaborator.

`err.vhdl` covers the other direction, and asserts that none of it produces an internal error: `'BASE` in a real type mark position; `ints8'base'left` → `prefix type is not constrained` (this used to give the wrong "cannot be used as a type mark" message); a non-type prefix; and `'BASE` alone in an expression.

`sanity gna vests synth` pass on all three backends.

Diagnostics that must stay were checked by hand and all still hold, with no internal error: `ints8'base'left` → `prefix type is not constrained` (the base type of an array subtype is unconstrained, so `'LEFT` really is undefined on it — the LRM's `T'BASE'LEFT` example is about scalar `T`); `s'base'…`, `work'base'…`, `f'base'…` → `a type mark must denote a type or a subtype`; `nosuch'base'…` → `no declaration for "nosuch"`; `integer'base` alone → `'base attribute not allowed in an expression`. `integer'base'base'left` and `s8'base'element'element` resolve correctly.

tgingold:
This looks ok (but still a draft).

dalex78:
I wanted to be sure the CI was running fine. Thank you !

Issue #2955, opened 2025-03-29
VHDL Synthesis: error with "nested array" port fully or partially unconstrained


sideprojectslab:
**Description**
VHDL-2008, when an entity has an input port of "nested array type" with more than one dimension, such as "signed_vector", the analyze/elaborate step fails if any of the port dimensions are unconstrained.

```
data_in : in  signed_vector;  -- not OK
data_in : in  signed_vector(0 to 7) ; -- not OK
data_in : in  signed_vector(0 to 7)(31 downto 0); -- OK
```

the reported error is:

```
******************** GHDL Bug occurred ***************************
Please report this bug on https://github.com/ghdl/ghdl/issues
GHDL release: 5.0.1 (4.1.0.r602.g37ad91899.dirty) [Dunoon edition]
Compiled with GNAT Version: 14.2.0
Target: x86_64-w64-mingw32
C:\Local\Data\eas\projects\eas\eas_ip_lib\svn\branches\release\v2\development\eas_ip_lib\fpga\vhdl\dsp\misc\test\test_sw_soft_mute\ghdl\_make\
Command line:
C:\Local\Development\compilers\GHDL 5.0\bin\ghdl.exe -r --std=08 --workdir=work -frelaxed aaa_test_tle --fst=waves.fst --max-stack-alloc=1024000
Exception ADA.ASSERTIONS.ASSERTION_ERROR raised
Exception information:
raised ADA.ASSERTIONS.ASSERTION_ERROR : synth-vhdl_expr.adb:635
******************************************************************
```

**Expected behaviour**
all the three port definitions above should be valid VHDL-2008

**How to reproduce?**

define an unconstrained "nested array" type in some package:

`type signed_vector   is array (natural range <>) of signed;`

make a dummy entity that uses the type as a port, only partially constrained (it would normally be constrained when the module is instantiated and a signal is connected to it)

```
entity dummy is
	generic
	(
		g_bit_width         : positive;
	);
	port
	(
		i_data : in  signed_vector(0 to g_num_chnl - 1);
		o_data : out signed_vector(0 to g_num_chnl - 1)(g_bit_width - 1 downto 0)
	);
end entity;

architecture behavioral of dummy is
begin
	o_data <= i_data;
end architecture;
```

Finally, instantiate the dummy entity in a top-level, assign a fully constrained signal to both i_data and o_data and and observe that the ananlysis/elaboration step fails with the error reported above

Paebbels:
[...]

PR #3375, opened 2026-08-20
testsuite: add regression test for issue2955, already fixed on master

dalex78:
Fixes issue https://github.com/ghdl/ghdl/issues/2955

## What the fix does

No source fix in this commit. Added `testsuite/gna/issue2955/` (a faithful reconstruction of the report's repro, covering both the fully-unconstrained and outer-only-constrained port shapes in one entity) whose `testsuite.sh` analyzes, elaborates, and simulates the design (with `--fst`, matching the report's wave-dump usage) and asserts it completes normally with no `GHDL Bug occurred`. This locks in the current, correct behavior as a regression guard.

### DIFF
--- a/testsuite/gna/issue2955/dummy.vhd
+++ b/testsuite/gna/issue2955/dummy.vhd
@@ -1,0 +1,23 @@
+library ieee;
+use ieee.numeric_std.all;
+use work.types_pkg.all;
+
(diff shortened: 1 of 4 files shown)


dalex78:
Wrong fix for the ticket.

PR #3377, opened 2026-08-20
vhdl-sem_assocs: constrain the element of an individual array associa…

### LINKED ISSUE #2955: VHDL Synthesis: error with "nested array" port fully or partially unconstrained

dalex78:
Fixes https://github.com/ghdl/ghdl/issues/2955

## What the fix does

Gives the array case the same treatment as the record case: when the element of the interface is not fully constrained, take the element subtype from the individual associations — they all share one — and set it on the synthesized subtype, which makes the `Fully_Constrained` marking true. `Element_Subtype` is a `Ref` field, so this refers to the actual's subtype without transferring ownership.

## Testing

`testsuite/gna/issue2955/repro.vhd` is the reporter's own design, elaborated and simulated. `chk.vhd` drives values across the element-wise association and checks them, because not crashing does not show the element bounds are right:

```
PASS b=9876 len=16
```

Identical on mcode, which accepts the design without the fix and so serves as a reference for the expected values.

The trigger is the unconstrained element, not the generic: with the element constrained in the port declaration the design elaborates on every backend, with or without this fix, whether the index range comes from a generic or a literal.

Full `sanity gna vests synth` pass on mcode, llvm and gcc.

### DIFF
--- a/src/vhdl/vhdl-sem_assocs.adb
+++ b/src/vhdl/vhdl-sem_assocs.adb
@@ -1070,6 +1070,39 @@ package body Vhdl.Sem_Assocs is
                  (Get_Associated_Expr (El), El_Type);
                El := Get_Chain (El);
             end loop;
+
+            --  If the element is not constrained by the interface, it is
+            --  defined by the individual associations - like the elements of
+            --  a record, cf. Finish_Individual_Assoc_Record.  Take it from
+            --  the first association; they all have the same subtype.
+            --  Without this, ACTUAL_TYPE keeps the unconstrained element of
+            --  the base type although it has just been marked as fully
+            --  constrained, so the bounds of the element are never
+            --  elaborated.
+            if Get_Kind (El_Type) in Iir_Kinds_Composite_Type_Definition
+              and then Get_Constraint_State (El_Type) /= Fully_Constrained
+            then
+               declare
+                  Assoc_Expr : constant Iir := Get_Associated_Expr (Chain);
+                  Assoc_Type : Iir;
+               begin
+                  if (Get_Kind (Assoc_Expr)
+                        = Iir_Kind_Association_Element_By_Individual)
+                  then
+                     Assoc_Type := Get_Actual_Type (Assoc_Expr);
+                  else
+                     Assoc_Type := Get_Type (Get_Actual (Assoc_Expr));
+                  end if;
+                  if Assoc_Type /= Null_Iir
+                    and then (Get_Kind (Assoc_Type)
+                                in Iir_Kinds_Composite_Type_Definition)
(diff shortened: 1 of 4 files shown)


dalex78:
@tgingold this is one of the remaining two currently open.
(I don't want to be pushy at all, tell me how you want me to operate, I don't mind at all.)
(I'll probably send a new batch of AI proposed fixes tomorrow. Same, if you don't want to be drowned in PRs, or you don't like the way certain AI related things are done, do tell me that's absolutely fine. Your project, your rules.)

tgingold:
I think this is OK but there might be missing checks for other elements to have matching bounds.

Issue #1622, opened 2021-01-26
Multiple definitions linkage errors on elaboration after analyzing two different files with the same name


machitgarha:
**Description**
When you analyze two different entity files located in two different places but with exactly the same name, after elaboration, you will get a bunch of annoying linkage errors representing something is defined twice.

**Expected behaviour**
Elaboration (and running) without any errors. Instead, a bunch of multiple definition errors.

**How to reproduce?**
1. Create two directories, `src` and `tests`.
2. Create two files with the same name inside them, e.g. `src/test.vhd` and `tests/test.vhd`.
3. Create two distinct entities in the files, in which you create an instance of the first entity inside second one's architecture. A minimal working example:

    ```vhd :file: src/test.vhd
    -- src/test.vhd
    library ieee;
    use ieee.std_logic_1164.all;

    entity something is
        port(
            clock: out std_logic
        );
    end entity;

    architecture structural of something is
    begin
    end architecture;
    ```

    ```vhd :file: tests/test.vhd
    -- tests/test.vhd
    library ieee;
    use ieee.std_logic_1164.all;

    entity other_thing is
    end entity;

    architecture structural of other_thing is
        component something is
            port(
                clock: out std_logic
            );
        end component;

        signal clock: std_logic;
    begin
        something_instance: something port map(clock);
    end architecture;
    ```
4. Analyze the files (no problems should happen):

    ```sh :image: ghdl/ghdl:buster-mcode
    ghdl -a --workdir=build/ {src,tests}/test.vhd
    ```

5. Elaborate the files:

    ```sh :image: ghdl/ghdl:buster-mcode
    # You will get errors after this
    ghdl --elab-run --workdir=build/ other_thing
    ```

    Wow, now look at your terminal:

    ```text
    /usr/bin/ld: build/test.o:(.rodata+0x40): multiple definition of `work__other_thing__RTI'; build/test.o:(.rodata+0x40): first defined here
    /usr/bin/ld: build/test.o: in function `work__other_thing__PKG_ELAB':
    test.vhd:(.text+0x0): multiple definition of `work__other_thing__PKG_ELAB'; build/test.o:test.vhd:(.text+0x0): first defined here
    /usr/bin/ld: build/test.o: in function `work__other_thing__DECL_ELAB':
    test.vhd:(.text+0x19): multiple definition of `work__other_thing__DECL_ELAB'; build/test.o:test.vhd:(.text+0x19): first defined here
    ...
[...]

Issue #539, opened 2018-03-07
Packages with same file name in different subdirectories results in compilation error


Zagitta:
I've recently run into an issue where the following folder structure results in a GHDL compilation error:
```
.
├── pkg1
│   └── pkg.vhd
├── pkg2
│   └── pkg.vhd
└── top.vhd

```
The two pkg.vhd files each contain an empty package declaration with different package names and top.vhd contains two use statements for the two packages and an empty top entity.

The ghdl commands issued and output are as following:
```
$ ghdl -i pkg1/pkg.vhd pkg2/pkg.vhd top.vhd 
$ ghdl -m top
analyze pkg1/pkg.vhd
analyze pkg2/pkg.vhd
analyze top.vhd
elaborate top
pkg.o:(.bss+0x0): multiple definition of `work__pkg2__ELABORATED'
pkg.o:(.bss+0x0): first defined here
pkg.o:(.data.rel.ro+0x0): multiple definition of `work__pkg2__RTI'
pkg.o:(.data.rel.ro+0x0): first defined here
pkg.o: In function `work__pkg2__ELAB_SPEC':
pkg.vhd:(.text+0x0): multiple definition of `work__pkg2__ELAB_SPEC'
pkg.o:pkg.vhd:(.text+0x0): first defined here
e~top.o: In function `work__pkg1__ELAB_BODY':
e~top:(.text+0x12): undefined reference to `work__pkg1__ELABORATED'
e~top:(.text+0x1c): undefined reference to `work__pkg1__ELABORATED'
e~top:(.text+0x22): undefined reference to `work__pkg1__ELAB_SPEC'
top.o: In function `work__top__PKG_ELAB':
top.vhd:(.text+0x1a): undefined reference to `work__pkg1__ELABORATED'
collect2: error: ld returned 1 exit status
ghdl:error: compilation error
```
```
$ ghdl -v
GHDL 0.35 (v0.35) [Dunoon edition]
 Compiled with GNAT Version: 7.3.0
 GCC back-end code generator
Written by Tristan Gingold.

Copyright (C) 2003 - 2015 Tristan Gingold.
GHDL is free software, covered by the GNU General Public License.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

A full reproduction of the issue can be found here: https://github.com/Zagitta/ghdl-bug
Simply execute `./run.sh`

Mercotui:
[...]

PR #3383, opened 2026-08-25
ghdldrv: report two design files compiled to the same object file

dalex78:
The object file of a design file is named after the base name of its source file, so two design files with the same base name in different directories are compiled to the same object file: the second analysis overwrites the first one and the link fails with "multiple definition" errors about symbols the user never wrote.

Detect the collision where the link list is built and report it, as suggested on both issues, instead of leaving the user with the linker errors.  This does not change how object files are named, which is what closing the issues needs.

Fixes part of #539 and #1622.

### DIFF
--- a/src/ghdldrv/ghdldrv.adb
+++ b/src/ghdldrv/ghdldrv.adb
@@ -336,6 +336,47 @@ package body Ghdldrv is
       Table_Low_Bound => 1,
       Table_Initial => 16);
 
+   --  Object files built from a source file, and the source file each one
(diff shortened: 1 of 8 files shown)


tgingold:
I think it should grep for 'JIT' instead of 'mcode'.

dalex78:
Investigating the fail

dalex78:
> should grep for 'JIT' instead of 'mcode'.

I'll look into that. Thank you

Issue #2052, opened 2022-05-06
GHDL crashes when using a full "hacked" 2^32 byte array.


tristanitschner:
**Description**

So recently I have been trying to work around the fact, that the maximum integer size in VHDL is 2^32 - 1. Given the amount of ram available nowadays, this shouldn't be a problem from a hardware perspective. This is what I came up with:

```vhdl
library ieee;
use ieee.numeric_std.all;
use ieee.std_logic_1164.all;

entity no_more_greater_than_pow_2_31_fuckery is 
  generic(
    width : integer := 8
         );
end;
architecture a_no_more_greater_than_pow_2_31_fuckery of no_more_greater_than_pow_2_31_fuckery is 
  subtype byte is std_logic_vector(width - 1 downto 0);
  type ram is array(integer range <>) of byte;
  type bank is array(integer range<>) of ram;

  signal mem : bank
  (2**16-1 downto 0)
  (2**16-1 downto 0);
  signal index : unsigned(32-1 downto 0) := (others => '0');

  function read(memx : bank; indexx : unsigned) return byte is
  begin
    return memx
    (to_integer(indexx(31 downto 16)))
    (to_integer(indexx(15 downto 0)));
  end function;

  procedure write(signal memx : inout bank; indexx : unsigned; value : byte) is
  begin
    memx
    (to_integer(indexx(31 downto 16)))
    (to_integer(indexx(15 downto 0))) 
    <= value;
  end procedure;

  signal clk,rst : std_logic := '1';
  constant clk_period : time := 10 ns;
begin

  clk <= not clk after clk_period/2;
  rst <= '1', '0' after clk_period*5;

  process (all) is
  begin
    if rising_edge(clk) then
      index <= index + "1";
      write(mem, index, std_logic_vector(index(byte'range)));
      if index > "0" then
        report "Wrote to memory the value " & to_hstring(read(mem, index - "1"));
      end if;
    end if;
  end process;

end;
```

Now when running
```sh
	ghdl -a --std=08 no_more_greater_than_pow_2_31_fuckery.vhd
	ghdl -e --std=08 no_more_greater_than_pow_2_31_fuckery
	ghdl -r --std=08 no_more_greater_than_pow_2_31_fuckery
```

this is what I get:

```sh
ghdl(32540,0x1182a1dc0) malloc: Incorrect checksum for freed object 0x7ff98ed37cd8: probably modified after being freed.
Corrupt value: 0x0
[...]

Issue #1111, opened 2020-01-23
generated executable crashes erroneously trying to allocate an extremely large amount of memory


roeekalinsky:
**Description**
When the width of a std_logic_vector signal exceeds a certain size, the ghdl executable crashes.  The crash is caused by the ghdl executable trying to allocate an amount of memory for that signal that is computed incorrectly, resulting in a requested size that is extremely large.  Observing using strace we see ghdl attempt to allocate a memory size of 0xFFFFFFFF80001000 when we would expect 0x0000000080001000 based on the memory allocation patterns observed in non-failing configurations.  This is highly suggestive of incorrect handling of conversion of a 32-bit value to a 64-bit value, possibly that it a 32-bit value is treated as signed when it should be treated as unsigned (looks like a 32-bit value that was sign-extended to 64-bit).

**Expected behaviour**
The ghdl executable should not crash, and should allocate the correct amount of memory in all cases.

**How to reproduce?**
A minimal working example is provided in the form of a trivial VHDL entity and a shell script that demonstrates configurations that work correctly and then a configuration where ghdl crashes as described above.  The shell script also contains further comments about the clues found in the strace logs that point to the suspected cause of the problem as described above.

example.vhd:
```vhd :file: example.vhd
library ieee;
use ieee.std_logic_1164.all;

entity example is
  generic (
    VEC_WIDTH : positive);
end example;

architecture behavioral of example is
  signal vec : std_logic_vector(VEC_WIDTH-1 downto 0);
begin
end behavioral;
```

example.sh:
```sh :file: example.sh
rm -rf ghdl_worklibs
mkdir  ghdl_worklibs

rm -rf ghdl_worklibs/example_lib
mkdir  ghdl_worklibs/example_lib

rm -rf ghdl_bin
mkdir  ghdl_bin

rm -rf strace_logs
mkdir  strace_logs

ghdl -a -Wno-hide  -Pghdl_worklibs/example_lib --workdir=ghdl_worklibs/example_lib --work=example_lib ./example.vhd

ghdl -e -Pghdl_worklibs/example_lib --workdir=ghdl_worklibs/example_lib --work=example_lib -o ghdl_bin/example example

# These configurations are ok.
#
# Note: Looking at strace, it looks like an amount of memory allocated equals
# round_up_to_power_of_two(8*VEC_WIDTH) + 4096
#
#                                                                                          # memory size allocated:
[...]

Issue #812, opened 2019-05-02
Large array crashes ghdl


wsneijers:
Hello,

I've been running into an issue with GHDL using large arrays in vhdl (declared as type). To start I already saw previous issues about the same sort of subject: #342, #471 and  #611. In those issues the problem is marked as solved. However I still have the same sort of issue, even with the latest master sources. 

In my case I'm using a test setup combining cocotb and GHDL. Where cocotb creates the test stimuli and monitors. Don't know if this can make a difference, but for completeness I thought id mention it. I'm trying to simulate the following file:

``` vhdl 
library IEEE;
use IEEE.std_logic_1164.all;
use IEEE.numeric_std.all;
use IEEE.std_logic_unsigned.all;

entity cosim_test is
end cosim_test;

architecture rtl of cosim_test is

type ram_type is array(0 to (2**28)-1) of std_logic_vector(127 downto 0);
signal ram : ram_type := (others => (others => '0'));

begin

end rtl;
```

Which for me causes the following output in the testbench:

```
loading VPI module '/home/docker/cocotb/build/libs/x86_64/libvpi.so'
     -.--ns INFO     cocotb.gpi                                  gpi_embed.c:114  in embed_init_python               Did not detect virtual environment. Using system-wide Python interpreter.
     -.--ns INFO     cocotb.gpi                                GpiCommon.cpp:91   in gpi_print_registered_impl       VPI registered
VPI module loaded!
/home/docker/cocotb/makefiles/simulators/Makefile.ghdl:63: recipe for target 'results.xml' failed
make[4]: *** [results.xml] Error 255
/home/docker/cocotb/makefiles/Makefile.sim:71: recipe for target 'sim' failed
make[3]: *** [sim] Error 2
CMakeFiles/Simulation.dir/build.make:68: recipe for target 'run_simulation' failed
make[2]: *** [run_simulation] Error 2
CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/Simulation.dir/all' failed
make[1]: *** [CMakeFiles/Simulation.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2

```

When I decrease the size of the array I get to a point where it starts working again:

``` vhdl
type ram_type is array(0 to (2**15)-1) of std_logic_vector(127 downto 0);
signal ram : ram_type := (others => (others => '0'));
```

[...]

PR #3385, opened 2026-08-25
report running out of memory instead of an internal error or a segfault

dalex78:
A design that needs more memory than the machine has did not say so. With mcode, Storage_Error escaped into the "GHDL Bug occurred" box; with gcc and llvm, the simulation stored through the null pointer a failed allocation returned and died on SIGSEGV, which grt reported as "NULL access dereferenced" -- blaming the user's design.

Add Grt.Errors.Out_Of_Memory (exported as __ghdl_out_of_memory) and call it from the grt allocation entry points: __gnat_malloc and __gnat_realloc, which every "new" in grt goes through, and __ghdl_malloc / __ghdl_malloc0, which the generated code calls.  Catch Storage_Error in the driver, the synthesis entry point and the translation front end for the mcode side.

Fixes part of #812, #1111 and #2052.

### DIFF
--- a/src/ghdldrv/ghdlmain.adb
+++ b/src/ghdldrv/ghdlmain.adb
@@ -701,6 +701,12 @@ package body Ghdlmain is
          Set_Exit_Status (Failure);
       when Exec_Error =>
          Set_Exit_Status (3);
+      when Storage_Error =>
+         --  Ran out of memory (eg elaborating an overly large design).
+         Simple_IO.Put_Line_Err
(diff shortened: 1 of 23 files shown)


tgingold:
Do we need index_add/index_mul ?  At least for mcode, the overflow should be caught.
(I am not sure it is worth investigating too much non-jit backends).

dalex78:
@tgingold I updated the description in the ticket: https://github.com/ghdl/ghdl/issues/2052#issuecomment-5418066338.

Issue #2765, opened 2024-09-27
GHDL Bug after "slice direction doesn't match index direction" error


Araneidae:
**Description**
I have the following transcript:
```
top.vhd:17:15:error: slice direction doesn't match index direction
        data_i(count'RANGE) => std_ulogic_vector(count_i),
              ^

******************** GHDL Bug occurred ***************************
Please report this bug on https://github.com/ghdl/ghdl/issues
GHDL release: 5.0.0-dev (tarball) [Dunoon edition]
Compiled with GNAT Version: 12.4.0
Target: x86_64-unknown-linux-gnu
/home/mga83/targetOS/IFC_1412/gddr6/sim/setup/do/ghdl-bug/
Command line:
/home/hir12111/.nix-profile/bin/ghdl -r --std=08 top
Exception CONSTRAINT_ERROR raised
Exception information:
raised CONSTRAINT_ERROR : elab-vhdl_objtypes.adb:1389 access check failed
******************************************************************
```
Note that the reported error is in fact incorrect as should be evident from the code below.

A MWE follows.

**How to reproduce?**

```vhd :file: handle.vhd
library ieee;
use ieee.std_logic_1164.all;

entity handle is
    port (
        data_i : in std_ulogic_vector;
        data_o : out std_ulogic_vector
    );
end;

architecture arch of handle is
begin
end;
```

```vhd :file: top.vhd
library ieee;
use ieee.std_logic_1164.all;
use ieee.numeric_std.all;

entity top is
    port (
        count_i : in unsigned(5 downto 0)
    );
end;

architecture arch of top is
    signal count : count_i'SUBTYPE;

begin
    handle : entity work.handle port map (
        data_i(count'RANGE) => std_ulogic_vector(count_i),
        unsigned(data_o(count'RANGE)) => count
    );
end;
```

```sh :image: ghdl/ghdl:bullseye-mcode
ghdl -a --std=08 handle.vhd top.vhd &&
ghdl -r --std=08 top
```

**Context**
The ghdl instance used was installed by a colleague @EmilioPeJu who can hopefully provide the correct git SHA for the installation.  This is running on RHEL8 and installed via NIX.

Edit: The git SHA is 4c45092.

Araneidae:
Turns out that I can remove quite a few red herrings: this issue is nothing to do with VHDL 2008 or type conversion and can be written as just:
```vhd :file: handle.vhd
library ieee;
use ieee.std_logic_1164.all;

entity handle is
[...]

PR #3388, opened 2026-08-25
vhdl-sem_assocs: direction of a formal sliced in several parts

### LINKED ISSUE #2688: IIR_KIND_CONCATENATION_OPERATOR thrown when instantiating an entity with unconstrained array type
This issue was present in 3.0.0 and I just installed version 4.1.0 and this issue still exists in my case. Here's the entity:

``` vhdl
library ieee;
use ieee.std_logic_1164.all;
use ieee.numeric_std.all;

entity skid_buffer_in is
    port (
        aclk    : in  std_logic;
        aresetn : in  std_logic;
        -- registered
        r_valid : in  std_logic;
        r_ready : out std_logic;
        r_data  : in  std_logic_vector;
        -- combinational
        c_valid : out std_logic;
        c_ready : in  std_logic;
        c_data  : out std_logic_vector
    );
end entity skid_buffer_in;

architecture rtl of skid_buffer_in is
    constant data_width : natural := r_data'length;
    signal b_data  : std_logic_vector(r_data'range);
    signal b_valid : std_logic; 
begin

    skid : process(aclk, aresetn)
    begin
        if (aresetn = '0') then
            r_ready <= '0';
            b_data  <= (others => '0');
            b_valid <= '0';
        elsif rising_edge(aclk) then
            if c_ready then
                b_valid <= '0';
            end if;
            if (r_valid and r_ready) = '1' then
                if c_ready = '0' then
                    b_data <= r_data;
                end if;
                b_valid <= not c_ready;
                r_ready <= c_ready;
            else
                r_ready <= c_ready or not b_valid;
            end if;
        end if;
    end process skid;

    with to_bit(b_valid) select c_data <= 
        r_data when '0',
        b_data when '1';
    with to_bit(b_valid) select c_valid <= 
        r_valid and r_ready when '0',
        '1' when '1';
    
end architecture rtl;
```
and an example instantiating entity (please excuse syntax errors as this is an excerpt from larger file):
``` vhdl
library ieee;
use ieee.std_logic_1164.all;
use ieee.numeric_std.all;

entity inst is
    port (
        aclk    : in  std_logic;
        aresetn : in  std_logic;
        wvalid  : in  std_logic;
        wready  : out std_logic;
        wdata   : in  std_logic_vector(31 downto 0);
        wstrb   : in  std_logic_vector( 3 downto 0)
    );
end entity inst;

architecture struct of inst is
    constant data_bytes : natural := 4;
    signal bi_wvalid, bi_wready   : std_logic;
    signal bi_wdata : std_logic_vector(data_bytes*8-1 downto 0);
    signal bi_wstrb : std_logic_vector(data_bytes-1   downto 0);
begin

    w_skid : entity work.skid_buffer_in
    port map(
        aclk    => aclk,
        aresetn => aresetn,
        r_valid => wvalid,
        r_ready => wready,
        r_data  => wstrb & wdata,
        c_valid => bi_wvalid,
        c_ready => bi_wready,
        c_data(wdata'length+wstrb'high downto wdata'length) => bi_wstrb,
        c_data(wdata'high downto 0) => bi_wdata
    );

end architecture struct;
```
Surprisingly the following subsitution does not exhibit the internal error, but instead the simulation error: "slice direction doesn't match index direction"
``` vhdl
r_data(wdata'length+wstrb'high downto 0)  => wstrb & wdata,
```
I tested with mcode and llvm backends. The mcode backend throws the error when starting simulation while llvm throws the error at compile time.

_Originally posted by @n8tlarsen in https://github.com/ghdl/ghdl/issues/2190#issuecomment-2215571585_

dalex78:
The index range of an unconstrained formal that a port map constrains through individual associations took its direction from the slice only when there was a single association element, and from the index subtype of the interface object otherwise -- 'natural', hence 'to'.  So a formal sliced in several parts was built as a 'to' object, and the 'downto' slices that constrained it were then rejected:

  d : entity work.sub port map (o(3 downto 2) => ha, o(1 downto 0) => la);
  error: slice direction doesn't match index direction

As designs are written 'downto' almost everywhere, this made slicing an unconstrained formal in more than one part unusable.  Take the direction from the slices when they all agree, which is what commit a93226f23 already did for a single slice.  Indexed names have no direction and don't take part; slices that disagree keep the direction of the index subtype and are still reported.

That error left a null type behind, and three places used it without checking, raising CONSTRAINT_ERROR right after the message.  Stop on the erroneous association in each of them instead.

Closes #2688.

### DIFF
--- a/src/simul/simul-vhdl_elab.adb
+++ b/src/simul/simul-vhdl_elab.adb
@@ -888,12 +888,22 @@ package body Simul.Vhdl_Elab is
          Formal := Inter;
       end if;
       Synth_Object_Name (Port_Inst, Formal, Formal_Base, Typ, Off);
+      if Formal_Base = No_Valtyp then
+         --  The formal is in error (eg: a slice whose direction doesn't match
+         --  the one of the port).
+         return;
+      end if;
       Typ := Unshare (Typ, Global_Pool'Access);
       Formal_Sig := Formal_Base.Val.S;
       Formal_Ep := (Formal_Sig, Off, Typ);
 
       Synth_Object_Name
         (Assoc_Inst, Get_Actual (Assoc), Actual_Base, Typ, Off);
+      if Actual_Base = No_Valtyp then
+         --  The actual is in error (eg: a slice whose direction doesn't match
+         --  the one of the port).
+         return;
+      end if;
       Typ := Unshare (Typ, Global_Pool'Access);
       Actual_Sig := Actual_Base.Val.S;
       Actual_Ep := (Actual_Sig, Off, Typ);
@@ -966,6 +976,11 @@ package body Simul.Vhdl_Elab is
          Formal := Inter;
       end if;
       Formal_Ep := Compute_Sub_Signal (Port_Inst, Formal);
+      if Formal_Ep.Base = No_Signal_Index then
+         --  The formal is in error (eg: a slice whose direction doesn't match
+         --  the one of the port).
+         return;
+      end if;
 
       Actual_Ep := (No_Signal_Index, No_Value_Offsets, null);
 
--- a/src/synth/synth-vhdl_insts.adb
+++ b/src/synth/synth-vhdl_insts.adb
@@ -1075,6 +1075,13 @@ package body Synth.Vhdl_Insts is
          Synth_Individual_Formal
            (Syn_Inst, Inter_Typ, Get_Formal (Iassoc), Typ, Offs);
 
+         if Typ = null then
+            --  The formal is in error (eg: a slice whose direction doesn't
+            --  match the one of the port).
+            Value_Offset_Tables.Free (Els);
+            return No_Net;
+         end if;
+
          --   2. synth expression
          V := Synth_Single_Input_Assoc
            (Syn_Inst, Typ, Syn_Inst, Get_Actual (Iassoc), Iassoc);
@@ -1175,6 +1182,13 @@ package body Synth.Vhdl_Insts is
          --   1. compute type and offset
          Synth_Individual_Formal
            (Syn_Inst, Inter_Typ, Get_Formal (Iassoc), Typ, Offs);
+
+         if Typ = null then
+            --  The formal is in error (eg: a slice whose direction doesn't
+            --  match the one of the port).
+            Release_Expr_Pool (Marker);
+            exit;
+         end if;
 
          --   2. Extract the value.
          O := Build_Extract (Get_Build (Syn_Inst), Port, Offs.Net_Off, Typ.W);
--- a/src/vhdl/vhdl-sem_assocs.adb
+++ b/src/vhdl/vhdl-sem_assocs.adb
@@ -1159,7 +1159,6 @@ package body Vhdl.Sem_Assocs is
          declare
             Index_Constraint : Iir;
             Index_Subtype_Constraint : Iir;
-            Indiv_Assoc : Iir;
             Dir : Direction_Type;
          begin
             --  Create an index subtype.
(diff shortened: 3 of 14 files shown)


tgingold:
I need to check with the LRM if this is compliant.

tgingold:
I think this is not conformant with LRM08 5.3.2.2 Index constraints and discrete ranges  e) 2):
```
If the subtype index range is undefined, and the interface object or subelement is associated by
more than one association element or is associated by a single association element in which the
formal designator is a slice name, then the direction of the index range of the object is that of
the corresponding index subtype of the base type of the interface object,
```

dalex78:
### The first attempt, and why it was wrong

The obvious reading is that the error itself is the bug: the user wrote two `downto` slices, so surely the formal should be built `downto`. `Finish_Individual_Assoc_Array` in `src/vhdl/vhdl-sem_assocs.adb` gives the constructed index range the direction of the *slice* when there is exactly one association element, and the direction of the index subtype (`natural`, hence `to`) otherwise — so a formal sliced in two parts is built `to` and its own `downto` slices are then rejected. The first version of the fix took the direction from the slices whenever they all agreed, which made the reporter's design analyze, elaborate, simulate and synthesize.

Tristan Gingold rejected that on the PR, quoting LRM08 5.3.2.2 e) 2):

> If the subtype index range is undefined, and the interface object or subelement is associated by more than one association element or is associated by a single association element in which the formal designator is a slice name, then the direction of the index range of the object is that of the corresponding index subtype of the base type of the interface object,

He is right, and the clause is not ambiguous. It names two cases — associated by *more than one* association element, and associated by a *single* association element whose formal designator is a slice name — and gives both the same answer: the direction comes from the index subtype of the base type. For `std_logic_vector` that base type is indexed by `natural`, which is ascending, so the object's index range is ascending and a `downto` slice of it is an error under the slice rule of LRM08 8.5. GHDL's message is a correct diagnostic, not a bug, and the first version of the fix was making illegal code legal.

The measured behaviour of that first version, on a formal sliced in two parts (`o(3 downto 2) => ha, o(1 downto 0) => la`) reported from inside the instantiated entity:

```
sub: o'left=3 o'right=0 ascending=false
```

whereas the rule requires `o'left=0 o'right=3 ascending=true`.

### The open question: the single-slice case (#2765)

The same clause covers "a single association element in which the formal designator is a slice name", and GHDL does *not* follow it there. Commit a93226f23 (Tristan Gingold, Sept 2024, "Finish_Individual_Assoc_Array: handle single individual assoc with slice", Fix #2765) made a single slice give its own direction to the formal, and `testsuite/gna/issue2765` locks that in with `data_i(5 downto 0) => count_i` on an unconstrained `data_i : in std_ulogic_vector`, which must analyze and simulate cleanly. Measured on current master:

| port map | index range built | conformant per e) 2) |
|---|---|---|
| `o(3 downto 0) => s` | `3 downto 0` | no — should be `0 to 3` |
| `o(3 downto 2) => ha, o(1 downto 0) => la` | `0 to 3`, slices rejected | yes |
| `o(0 to 1) => ha, o(2 to 3) => la` | `0 to 3` | yes |

So master is already non-conformant for one slice and conformant for several, and this fix leaves that asymmetry exactly as it is. Making GHDL conformant means reverting a93226f23 as well and changing `testsuite/gna/issue2765` to expect an error, which re-breaks the design that issue was filed about. That is a maintainer policy call, not something to decide inside a crash fix, so it is raised on the PR rather than done here.

dalex78:
I'll see if a fix for all of that can be made.

dalex78:
Ticket also have its message updated: https://github.com/ghdl/ghdl/issues/2688#issuecomment-5417631353

PR #3389, opened 2026-08-25
vhdl: keep the declared purity of a function apart from the analysed one

dalex78:
# Issue #1574 follow-up — separate the purity a function was declared with from the purity it was found to have

Analysis by an AI agent.

## What this is about

PR #3378 was merged with a reservation, and the maintainer left it in the code as a FIXME (commit 7c3088265):

> I am not fully convinced about the fix: it basically makes the function impure. There is a consequence: if you reprint the AST, it won't match the sources. I will add a comment about that.

```ada
      --  The function is not pure, whatever it was declared.  Make it
      --  impure.
      --  FIXME: this violates reprint.
      Set_Pure_Flag (Subprg, False);
```

He is right, and the consequence is wider than reprint.

## Why overwriting Pure_Flag is the wrong carrier

`Pure_Flag` is *what the source declared*. Twelve places read it, and most of them implement an LRM rule about the declared purity, not about what the body turned out to do. Overwriting it mid-analysis therefore changes four separate behaviours. All four were reproduced on master (`d7d5d55e8`) against the same build with the one line reverted:

**1. Reprint no longer matches the source, and `ghdl fmt` aborts** — the reported consequence, plus a sharper form of it.

```vhdl
pure function f return std_logic is begin return s; end f;
```
`ghdl --reprint --std=08 -frelaxed` prints `impure function f` on master and `pure function f` without the line. Only when the source spelled the keyword, since `Vhdl.Prints` gates on `Get_Has_Pure`.

`ghdl fmt` on the same file does not merely print the wrong thing, it stops:

```
raised TYPES.INTERNAL_ERROR : vhdl-formatters.adb:69
```
("need to print: impure, but read pure from file"). The formatter is what an editor integration runs, so the merged fix turns a warning-level design into a bug box in the editor.

**2. Every violation after the first is silently dropped.** `Sem_Check_Pure` returns early when `Get_Pure_Flag` is false, so once the flag is cleared the function stops being checked. A function reading two signals reports two warnings without the line and one with it.

**3. An unrelated rule stops being enforced, depending on declaration order.**

```vhdl
function f return bit is
  constant c : bit := s1;   --  clears Pure_Flag here
  file fp : text;           --  "cannot declare a file in a pure function": LOST
begin return c; end f;
```
Swap the two declarations and the message comes back. `Vhdl.Sem_Decls` checks that rule against `Get_Pure_Flag`.

**4. Strict mode grows a misleading cascade.** With a resolution function that breaks the rule, master adds two errors that were not there before:

```
t7.vhdl:8:21:error: resolution function "resolv" must be pure
t7.vhdl:8:21:error: no matching resolution function for "resolv"
```

The flag is also copied into generic instantiations (`Vhdl.Sem_Inst`) and shown by `--disp-tree`, both of which then report a purity the source never wrote.

## And it only covers one of the three shapes

The crash is that synthesis keeps the instance const for a function whose `Pure_Flag` is set, takes `Synth_Static_Subprogram_Call`, and asserts the result is static. A function can fail to be pure in three ways, and `Sem_Check_Pure` is only the first:

| shape | on master |
|---|---|
| reads a signal of the enclosing declarative part | fixed by #3378 |
| declared pure, calls an impure **function** | still crashes, `synth-vhdl_stmts.adb:3050` |
| declared pure, calls a procedure that reads a signal | still crashes, `synth-vhdl_stmts.adb:719` |
| declared pure, calls a *pure-declared* function that reads a signal | still crashes, `synth-vhdl_stmts.adb:3050` |

And one case that the merged fix *does* handle and that any replacement must keep handling: a **procedure** that calls such a function, called in turn from a clocked process. Clearing `Pure_Flag` is what makes `Sem_Call_Purity_Check` mark that procedure impure; simply reverting the line brings the crash back there.

The last three are reported by `Sem_Call_Purity_Check`, which — like `Sem_Check_Pure` before #3378 — reported and recorded nothing. The fourth also needs the impurity to *propagate*: `Sem_Call_Purity_Check` gives up on a callee whose `Pure_Flag` is set, so the fact has to travel with the analysed purity, not with the declared one.

## What this change does

Give a function the same thing a procedure already has: `Purity_State`. `Pure_Flag` goes back to meaning only *"what the source declared"*, and `Purity_State` records *"the analysis found a violation"*.

- `src/vhdl/vhdl-nodes.ads`: `Purity_State (State2)` is no longer "Only for `Iir_Kind_Procedure_Declaration`". `State2` was already physically present and unused on a function node, so this costs nothing but the `make -C src/vhdl` regeneration of `vhdl-nodes_meta.adb`, and `Init_Node` zeroes it, i.e. `Unknown`, for the function nodes that never go through `Sem_Subprogram_Specification` (predefined operators, implicit operations).
- `Sem_Check_Pure` sets `Purity_State` to `Impure` instead of clearing `Pure_Flag`, in both of its function branches.
- `Sem_Call_Purity_Check` and `Update_And_Check_Pure_Wait` do the same where they already report a violation for a function.
- `Sem_Call_Purity_Check` and `Sem_Call_All_Sensitized_Check` also ask `Is_Pure_Function` rather than `Get_Pure_Flag` about the *callee*. The first is what makes the impurity propagate — up a chain of functions that are all declared pure, and out to a procedure that calls one, which is the one case the merged fix does handle and a plain revert would lose. The second matters just as much: without it a `process (all)` that calls such a function is not sensitive to the signal it reads, which is a wrong simulation result rather than a missing diagnostic.
- `Vhdl.Utils.Is_Pure_Function` answers the question synthesis actually has — declared pure *and* never found otherwise — and `Synth_Subprogram_Call` uses it in place of `Get_Pure_Flag`. It is guarded on the node kind: `Is_Function_Declaration` is also true for an interface function and for a function instantiation, and neither carries the new state.

Nothing else changes: no other consumer of `Pure_Flag` sees a different value than it did before #3378.

## Testing

- `testsuite/synth/issue1574/` now synthesizes six designs instead of two: the report's impure function, the thread's implicitly-pure one, the three call shapes above, and the procedure caller. It fails on master.
- `testsuite/gna/issue1574/` is new and covers the analysis side: all three violations of one function are reported, `ghdl --reprint` still prints `pure function`, `ghdl fmt` still accepts the file, and a `process (all)` calling the function is sensitive to the signal it reads. It fails on master, and the last check also fails on a plain revert of #3378.

Both pass on mcode, llvm and gcc, and the full `sanity`, `gna` and `synth` suites pass on all three backends.

## Note on the trigger

A pure rule violation is a hard error by default under `--std=08`; it is a warning under `--std=93c` (GHDL's default) or with `-frelaxed`. So this whole area is only reachable when the rules are relaxed — which is exactly why the AST must stay honest about what the source said while the compiler carries on.


dalex78:
@tgingold I am not convinced that this is going in the right direction, I don't know. I'll let you judge the fix. I don't think I'll iterate on that one anymore.

tgingold:
I think this is correct.

dalex78:
I'll resolve the branch conflict, and I think you'll be able to merge it then.

PR #3390, opened 2026-08-25
Bug/issue2955 follow up

dalex78:
# Issue #2955 follow-up — check that the individual associations agree on the element subtype

Analysis by an AI agent.

## What this is about

PR #3377 was merged with a reservation:

> I think this is OK but there might be missing checks for other elements to have matching bounds.

The merged code takes the element subtype of the synthesized actual from the *first* individual association, on the strength of a comment:

```ada
   --  Take it from the first association; they all have the same subtype.
```

Nothing checks that, and they need not.

## What the LRM requires

LRM08 5.3.2.1: *"An array object is a composite object consisting of elements that have the same subtype."* The formal is one object with one element subtype, so the several associations do not each contribute one — together they must agree.

LRM08 5.3.2.2 e) 3) says where that subtype comes from when the interface does not constrain it (*"the index range is obtained from the object or value denoted by the actual designator"*), and then closes the rule:

> For a given array interface object, or for a given array subelement of an interface object, it is an error if application of the preceding rules yields different index ranges for any corresponding array subelements of the given interface object or given subelement.

That is exactly the missing check, and note *"any corresponding array subelements"*: the rule is recursive.

## What goes wrong without it

Two channels whose actuals are 16 and 8 bits wide, on current master:

| shape | mcode | gcc, llvm |
|---|---|---|
| port, `i_data(0) => a16, i_data(1) => a8` | `TYPES.INTERNAL_ERROR : simul-vhdl_simul.adb:3613` with the bug box | `bound check failure` at elaboration |
| **generic**, `g_data(0) => to_signed(11,8), g_data(1) => to_signed(22,16)` | **no diagnostic at all**: `g_data(1)` reads back as `len=8 v=0`, exit status 0 | `bound check failure` at elaboration |

The generic case is the bad one: the value 22 is silently lost and the run succeeds.

## What this change does

**1. Check the associations against each other** (`Vhdl.Sem_Assocs`). The element subtype is now settled once for the whole actual, by `Finish_Individual_Assoc_Array_El_Subtype`, called from `Finish_Individual_Association1` right after the array walk. It takes the element subtype from the first association that is fully constrained and reports the ones that do not match it:

```
mismatch.vhd:25:55:error: element subtype of individual association does not match the one at line 25:39
```

Settling it once, over the whole choice tree, is what makes a multi-dimensional array work: `Finish_Individual_Assoc_Array` recurses per dimension and reaches the innermost chain once per outer index, so a check placed there compares only the elements of one row against each other. `i_data(0,0)`…`i_data(1,1)` are now all compared.

Three details:

- **A slice formal part gives its own element subtype.** With `i_data(0 to 1) => aa`, the actual is a slice of the array, so `Get_Type (Get_Actual (…))` is the *array* subtype; the merged code installed `signed_vector` as the element subtype of a `signed_vector`. That design is legal and it crashed with an internal error on mcode — `elab-vhdl_objtypes.adb:588`, and `elab-vhdl_insts.adb:651` before #3377 — which it no longer does: mcode now elaborates and simulates it correctly. gcc and llvm still end in a bound check failure there, exactly as they did before, so the shape is not fully supported yet; the remaining half is in the translator, not here.
- **A conversion on the actual defines the constraint**, per LRM08 5.3.2.2 e) 3), so the subtype comes from the conversion rather than from `Get_Actual`, which returns the pre-conversion expression. A conversion on the *formal* is different: the LRM names the conversion's *parameter* subtype there, which is not what the conversion node carries, so such an association is left out of the comparison rather than compared against the wrong type — otherwise `w8(o(1)) => x8`, which works today, would be rejected.
- **`Get_Actual` is not valid on an `Iir_Kind_Association_Element_Open`**, which has an `Open_Actual` instead; the merged code read it unguarded.

Two deliberate limits, both consistent with what GHDL already does elsewhere:

- **Lengths, not index ranges.** The LRM sentence above is about index ranges, so `signed(15 downto 0)` against `signed(0 to 15)` is formally an error. GHDL accepts it today and produces the right result, and the existing check for the constrained case — `Eval_Is_In_Bound` in `Sem_Association_By_Expression`, *"actual constraints don't match formal ones"* — also compares lengths only. Rejecting a direction difference would be a new error on code that works; it is a separate decision for the maintainer, so this check compares lengths, recursively through nested array elements.
- **Locally static bounds only.** Bounds that are not locally static cannot be compared during analysis, exactly as for the neighbouring *"indexes of individual association mismatch"* check. They are left to the run-time check — see (2).

**2. Report the run-time mismatch instead of crashing** (`Simul.Vhdl_Simul`). When the bounds are not locally static the mismatch only shows up while connecting the ports, and `Connect` answered it with `raise Internal_Error`, i.e. GHDL's bug box — where gcc and llvm report a bound check failure. It now reports, with the location of the association:

```
dyn_mismatch.vhd:28:62: length of actual doesn't match length of formal
ghdl:error: error during elaboration
```

## What is still not checked

An interface whose *index* bounds are locally static but whose *element* bounds are not — `sv(0 to 1)(w-1 downto 0)` with `w` a generic — goes down the other branch of `Finish_Individual_Association1` and never reaches this code, so a mismatch there is still only caught at elaboration (cleanly now, see (2)). Widening the check to that branch means giving it an actual subtype to compare against, which is a larger change than this one.

## Testing

`testsuite/gna/issue2955/` keeps the report's design and the data-crossing check, and adds:

- `mismatch.vhd` — two ports of different width: rejected during analysis;
- `gen_mismatch.vhd` — the same on a generic, the silent-data-loss case: rejected during analysis;
- `dyn_mismatch.vhd` — the same mismatch behind a function call, so the bounds are not locally static: a clean elaboration error and no bug box;
- `slice.vhd` — a slice as the formal part: no longer an internal error (it still fails on gcc and llvm, as before, so the test only asserts the absence of the bug box);
- `conv.vhd` — a conversion function on the formal part of an out port: still accepted, still correct;
- `dir.vhd` — same length, opposite direction: still accepted, still correct.

The test fails on master. It passes on mcode, llvm and gcc, and the full `sanity`, `gna` and `synth` suites pass on all three backends.

### DIFF
--- a/src/simul/simul-vhdl_simul.adb
+++ b/src/simul/simul-vhdl_simul.adb
@@ -3598,7 +3598,8 @@ package body Simul.Vhdl_Simul is
    -- PORT is the formal, while SIG is the actual.
    procedure Connect (Dst : Memtyp;
                       Src : Memtyp;
-                      Mode : Connect_Mode) is
+                      Mode : Connect_Mode;
+                      Loc : Node) is
    begin
       pragma Assert (Dst.Typ.Kind = Src.Typ.Kind);
 
@@ -3610,13 +3611,20 @@ package body Simul.Vhdl_Simul is
                Etyp : constant Type_Acc := Dst.Typ.Arr_El;
             begin
                if Len /= Src.Typ.Abound.Len then
-                  raise Internal_Error;
+                  --  The formal and the actual do not have the same length.
+                  --  That is a bound check failure of the association, not
+                  --  an internal error: it happens for instance when the
+                  --  elements of an individual association do not all have
+                  --  the same subtype and the bounds are not locally static,
+                  --  so that Sem cannot report it.
+                  Error_Msg_Exec
+                    (Loc, "length of actual doesn't match length of formal");
                end if;
                for I in 1 .. Len loop
                   Connect ((Etyp, Sig_Index (Dst.Mem, (I - 1) * Etyp.W)),
                            (Src.Typ.Arr_El,
                             Sig_Index (Src.Mem, (I - 1) * Etyp.W)),
-                           Mode);
+                           Mode, Loc);
                end loop;
             end;
             return;
@@ -3628,7 +3636,7 @@ package body Simul.Vhdl_Simul is
                   Connect ((E.Typ, Sig_Index (Dst.Mem, E.Offs.Net_Off)),
                            (Src.Typ.Rec.E (I).Typ,
                             Sig_Index (Src.Mem, E.Offs.Net_Off)),
-                           Mode);
+                           Mode, Loc);
                end;
             end loop;
          when Type_Logic
@@ -3862,7 +3870,7 @@ package body Simul.Vhdl_Simul is
             --  LRM93 12.6.2
             --  A signal is said to be active [...] if one of its source
             --  is active.
-            Connect (To_Memtyp (C.Actual), Form2, Connect_Source);
+            Connect (To_Memtyp (C.Actual), Form2, Connect_Source, C.Assoc);
          end;
       end if;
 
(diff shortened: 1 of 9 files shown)


dalex78:
Again @tgingold, a large change, I can't really follow. I'll probably stop iterating on that issue.

Issue #3088, opened 2025-11-06
GHDL Bug occurred


HuskyNator:
**Description**
Compilation error, requesting I file a bug report.

**Expected behaviour**
Either successful compilation or a note about incorrect code.

**How to reproduce?**
Analyze these 2 files.
Minimal working example:

```vhd :file: ripple_carry_adder.vhd
library ieee;
context ieee.ieee_std_context;

library exercises;

entity ripple_carry_adder is
    generic (bit_count: positive);
	port (
		sum : out std_logic_vector(bit_count - 1 downto 0);
		c_out : out std_logic
	);
end entity;

architecture structural of ripple_carry_adder is
begin
    -- ...
end architecture;
```

```vhd :file: ripple_carry_adder_tb.vhd
library ieee;
context ieee.ieee_std_context;

entity ripple_carry_adder_tb is
	generic (width : positive := 4);
end entity;

architecture tb of ripple_carry_adder_tb is
	signal sum : unsigned(width downto 0) := (others => '0');
begin
	UUT : entity work.ripple_carry_adder
        generic map(bit_count => width)
		port map (
            unsigned(sum) => sum(width - 1 downto 0),
            c_out => sum(width)
        );
	process
	begin
        -- ...
		wait for 10 ns;
		std.env.finish;
	end process;
end architecture;
```

```sh :image: ghdl/ghdl:bullseye-mcode
ghdl -a --std=08 ripple_carry_adder.vhd
ghdl -a --std=08 ripple_carry_adder_tb.vhd
```
_(I hope I'm not supposed to change the name of this image ^)_

**Context**
- OS: Windows 10
- Origin:
  - [-] Package manager: `version`
```
GHDL 5.1.1 (tarball) [Dunoon edition]
 Compiled with GNAT Version: 15.1.0
 llvm 20.1.7 code generator
```

If a `GHDL Bug occurred` block is shown in the log, please paste it here:

```
******************** GHDL Bug occurred ***************************
Please report this bug on https://github.com/ghdl/ghdl/issues
GHDL release: 5.1.1 (tarball) [Dunoon edition]
Compiled with GNAT Version: 15.1.0
Target: x86_64-w64-mingw32
C:\Users\pault\Documents\Code\VHDL\exercises\
Command line:
[...]

PR #3394, opened 2026-08-27
ghdldrv: give each design file its own object file

### LINKED ISSUE #539: Packages with same file name in different subdirectories results in compilation error

### LINKED ISSUE #1622: Multiple definitions linkage errors on elaboration after analyzing two different files with the same name

dalex78:
Closes #539 
Closes #1622 

# Issues #539 and #1622 — object files collide when two sources share a base name

Analysis by an AI agent.

## What was left open

Both issues report the same thing: with the gcc or llvm backend, `ghdl -a` writes an object file named after the source file's *base* name into the library directory, so two design files with the same base name overwrite each other's object file, and the link fails with a wall of "multiple definition" messages about symbols the user never wrote.

The previous change ([#3383](https://github.com/ghdl/ghdl/pull/3383)) only implemented the sanity check tgingold had considered ("Maybe I should add code to detect this situation"), and left the naming open, because the thread has three proposals for it:

- a hash of the full file path, Mercotui on #539 — tgingold: *"Adding a hash would prevent from moving files"*;
- a unique number per object file, tgingold on #1622: *"A fix could be to append a unique number to each object file"*;
- mirroring the source directory under the object directory, jpf91 on #1622: *"`src/test.vhd` => `obj/mylib/src/test.o`"*.

## What the fix does

An object file has to be told apart from another by everything that tells the two source files apart, and `Ghdllocal.Get_Object_Prefix` now names it after the source file itself: its name, its **extension**, and a **number** identifying its directory.

```
src/test.vhd   ->  test.vhd-23236.o
tests/test.vhd ->  test.vhd-50195.o
ext/pkg.vhd    ->  pkg.vhd-66865.o
ext/pkg.vhdl   ->  pkg.vhdl-59879.o
top.vhd        ->  top.vhd.o
```

All of them stay directly in the library directory, as they always have.

- The **extension** is what closes the case Mercotui reported on #539: `example.vhd` and `example.vht` are two different files of the *same* directory, so nothing about the directory can separate them.
- The **number** is tgingold's proposal from #1622. It is a hash (fnv-1a, five decimal digits) of the file name as it was written on the command line — its relative path, its name and its extension, which together are unique within a library. A source file given without any directory has nothing to disambiguate and keeps a plain object file name.

Three places name an object file and all three go through that one function: `Ghdllocal.Append_Suffix`, which names the object (and asm, and post) file of an analysis and is also what `--clean` and `--remove` delete; `Ghdldrv.Get_Object_Filename`, used by `--gen-makefile`, by `-f`, and by `Missing_Object_File`, which is how `-m` decides to recompile; and `Ghdldrv.Add_File_List`, which turns the file list written at elaboration into the object files handed to the linker.

Because the library file records the file name as it was given on the command line (`file "/home/u/proj/" "src/test.vhd"`), analysis and link derive the same name from the same string, and nothing new has to be stored in the library.

`Check_Object_Collision` from #3383 stays, now as a backstop for a collision of that number rather than as the general case.

## Why this hash is not the hash tgingold rejected

Mercotui proposed hashing the **full file path**, and tgingold's objection was *"Adding a hash would prevent from moving files"* — an absolute path changes when the project is moved, so every object file would be renamed and recompiled.

This number is computed on the path **as it was written on the command line**, which is relative in the case that matters. `../../src/ieee/numeric_std.vhdl` hashes the same wherever the project sits, and so does `src/test.vhd`. Only a source given by an absolute path gets a number that moves with it, and it is then no worse off than the design file itself, which the library records by that same absolute path.

## Why not mirror the source directory under the object directory

That was implemented first, and then dropped, for two reasons found while testing it:

- The library directory is the current directory by default. Mirroring then puts the object files **inside the user's source subdirectories** — `ghdl -m top` in a project without `--workdir` scatters `.o` files across `src/`, `tests/` and so on. Only a build directory makes the mirror look like jpf91's example.
- It does not close #539 anyway. `example.vhd` and `example.vht` are in the same directory, so they still need the extension, and a source whose directory cannot be mirrored — an absolute path, or one with a `..` component, which is how the vhdl libraries of ghdl itself are analyzed (`../../src/ieee/numeric_std.vhdl`) — still needs a number. The mirror was a third mechanism on top of the two that are actually needed.

`install.vhdllib` also copies `$(LIBDST_DIR)/$$d/*`, one flat directory at a time, so mirroring the object files of the standard libraries into subdirectories would have meant not installing them at all.

The one thing the mirror gave that this does not is jpf91's other point: with mirrored names, a makefile wildcard rule can derive the source of an object. `--gen-makefile` still emits a correct explicit rule per file, but `test.vhd-23236.o` cannot be turned back into `src/test.vhd` by a pattern.

## Things to know before merging

- **Every object file is renamed**, once: `test.o` becomes `test.vhd-23236.o`, and `std_standard.o` becomes `std_standard.vhdl.o`. The two places that spell the latter out — `Ghdldrv.Link` and the `libs.vhdl.standard` targets in `libraries/Makefile.inc` — were updated.
- **Upgrading is handled but leaves litter.** `-m` recompiles the files whose object is not at the new name (`Missing_Object_File` looks it up through the same function), and the link list only contains the new names, so a stale `pkg.o` from an older GHDL is never linked in — but it is not deleted either, and `--clean` will not remove it.
- The number is five decimal digits. Two files that collide on it are reported by `Check_Object_Collision`, not silently mislinked.
- `Ghdllocal.Get_Base_Name` no longer has any caller: object naming was its only use.

## Testing

- `testsuite/gna/issue539` — the reporter's `pkg1/pkg.vhd` + `pkg2/pkg.vhd`, through `ghdl -i` and `ghdl -m`, run, asserting there is one object file per source and that `--clean` removes them; then Mercotui's case, `ext/pkg.vhd` + `ext/pkg.vhdl`, analyzed, linked and run, asserting the extension is what separates them.
- `testsuite/gna/issue1622` — the reporter's `src/test.vhd` + `tests/test.vhd`, analyzed, elaborated and run, asserting one object file per source; then a `..` path and an absolute path, asserting each gets its own; then a source analyzed without any directory, asserting it keeps a plain object file name (that is how `std_standard` is compiled).

Both skip on the JIT backends, which write no object files.

Full `gna` (1139 tests), `sanity` and `synth` suites pass on all four backend variants: llvm, gcc, mcode and llvm-jit.

### DIFF
--- a/libraries/Makefile.inc
+++ b/libraries/Makefile.inc
@@ -369,19 +369,19 @@ $(IEEE19_DIR)/ieee-obj19.cf: $(ANALYZE_DEP) $(IEEE19_SRCS) $(SYNOPSYS19_SRCS) $(
 
 ANALYZE_STD=$(GHDL) --bootstrap-standard $(GHDL_FLAGS)
 
-$(STD87_DIR)/std_standard.o: $(GHDL) std.v87
+$(STD87_DIR)/std_standard.vhdl.o: $(GHDL) std.v87
 	cd $(STD87_DIR); $(ANALYZE_STD) --std=87
 
-$(STD93_DIR)/std_standard.o: $(GHDL) std.v93
+$(STD93_DIR)/std_standard.vhdl.o: $(GHDL) std.v93
 	cd $(STD93_DIR); $(ANALYZE_STD) --std=93
 
-$(STD08_DIR)/std_standard.o: $(GHDL) std.v08
+$(STD08_DIR)/std_standard.vhdl.o: $(GHDL) std.v08
 	cd $(STD08_DIR); $(ANALYZE_STD) --std=08
 
-$(STD19_DIR)/std_standard.o: $(GHDL) std.v19
+$(STD19_DIR)/std_standard.vhdl.o: $(GHDL) std.v19
 	cd $(STD19_DIR); $(ANALYZE_STD) --std=19
 
-libs.vhdl.standard: $(STD87_DIR)/std_standard.o \
- $(STD93_DIR)/std_standard.o \
- $(STD08_DIR)/std_standard.o \
- $(STD19_DIR)/std_standard.o
+libs.vhdl.standard: $(STD87_DIR)/std_standard.vhdl.o \
+ $(STD93_DIR)/std_standard.vhdl.o \
+ $(STD08_DIR)/std_standard.vhdl.o \
+ $(STD19_DIR)/std_standard.vhdl.o
--- a/src/ghdldrv/ghdldrv.adb
+++ b/src/ghdldrv/ghdldrv.adb
@@ -351,15 +351,13 @@ package body Ghdldrv is
       Table_Low_Bound => 1,
       Table_Initial => 16);
 
-   --  The object file of a design file is named after the *base* name of its
-   --  source file, in the library directory.  So two design files with the
-   --  same base name in different directories (dir1/pkg.vhdl and
-   --  dir2/pkg.vhdl) are both compiled to the same pkg.o: the second
-   --  analysis overwrites the first object file, and the link then either
-   --  fails with a pile of "multiple definition" messages or silently uses
-   --  the wrong object.  Detect it here, where the link list is built, and
-   --  say what is wrong instead of leaving the user with linker errors
-   --  about symbols they never wrote.  See #539 and #1622.
+   --  Two design files of a library are not supposed to share an object
+   --  file anymore, as Ghdllocal.Get_Object_Prefix identifies each one by
+   --  the name of its source file and a number.  Only that number may
+   --  collide.  Keep checking it here, where the link list is built:
+   --  otherwise one object file overwrites the other and the link either
+   --  fails with a pile of "multiple definition" messages or silently
+   --  uses the wrong object.  See #539 and #1622.
    procedure Check_Object_Collision (Obj : String; Src : String) is
    begin
       for I in Objlist.First .. Objlist.Last loop
@@ -368,7 +366,7 @@ package body Ghdldrv is
          then
             Error ("'" & Src & "' and '" & Objsrclist.Table (I).all
                      & "' are both compiled to '" & Obj & "'");
-            Error ("(object files are named after the source file base name;"
+            Error ("(the object file is named after the source file;"
                      & " rename one of these files)");
             raise Compile_Error;
          end if;
@@ -449,9 +447,9 @@ package body Ghdldrv is
                Filelist.Append (File);
             else
                if To_Obj then
-                  File := new String'(Dir (1 .. Dir_Len)
-                                      & Get_Base_Name (Line (1 .. L))
-                                      & Obj_Suffix);
+                  File := new String'
+                    (Get_Object_Prefix (Dir (1 .. Dir_Len), Line (1 .. L))
+                       & Obj_Suffix);
                   Check_Object_Collision (File.all, Line (1 .. L));
                else
                   File := new String'(Substitute (Line (1 .. L)));
@@ -475,7 +473,7 @@ package body Ghdldrv is
    begin
       Dir := Get_Library_Directory (Get_Library (File));
       Name := Get_Design_File_Filename (File);
-      return Image (Dir) & Get_Base_Name (Image (Name)) & Obj_Suffix;
+      return Get_Object_Prefix (Image (Dir), Image (Name)) & Obj_Suffix;
    end Get_Object_Filename;
 
    procedure Add_Argument (Inst : in out Instance; Arg : String_Acc) is
@@ -1184,7 +1182,7 @@ package body Ghdldrv is
               String'(Get_Machine_Path_Prefix & Get_Directory_Separator
                       & "std" & Get_Directory_Separator
                       & Get_Version_Path & Get_Directory_Separator
-                      & "std_standard" & Obj_Suffix);
+                      & "std_standard.vhdl" & Obj_Suffix);
             P := P + 1;
             Args (P) := Std_File;
          else
--- a/src/ghdldrv/ghdllocal.adb
+++ b/src/ghdldrv/ghdllocal.adb
@@ -593,6 +593,47 @@ package body Ghdllocal is
       return Filename (First .. Last);
    end Get_Base_Name;
 
+   --  A short number identifying the source file among the files of a
+   --  library.  This is a hash (fnv-1a) of the file name as it was written
+   --  on the command line: its relative path, its name and its extension,
+   --  which together are unique in a library.  It is computed on the path
+   --  as written, so it doesn't change when relative sources are moved.
+   function Get_File_Number (File : String) return String
+   is
+      Hash : Uns32 := 16#811c_9dc5#;
+      Img : String (1 .. 5);
+      Val : Uns32;
+   begin
+      for I in File'Range loop
+         Hash := Hash xor Uns32 (Character'Pos (File (I)));
+         Hash := Hash * 16#0100_0193#;
+      end loop;
+
+      --  Five decimal digits are enough to tell apart the files of a
+      --  design; Ghdldrv.Check_Object_Collision reports the rest.
+      Val := Hash mod 100_000;
+      for I in reverse Img'Range loop
+         Img (I) := Character'Val (Character'Pos ('0') + Val mod 10);
+         Val := Val / 10;
+      end loop;
+      return Img;
+   end Get_File_Number;
+
+   function Get_Object_Prefix (Dir : String; File : String) return String
+   is
+      Pos : constant Natural := Get_Basename_Pos (File);
+      --  The name of the source file, extension included.
+      Name : constant String := File (Pos + 1 .. File'Last);
+   begin
+      if Pos < File'First then
+         --  No directory: the name is enough to identify the file within
+         --  the library.  This is also the case of std_standard.
+         return Dir & Name;
+      else
+         return Dir & Name & '-' & Get_File_Number (File);
+      end if;
+   end Get_Object_Prefix;
+
    function Append_Suffix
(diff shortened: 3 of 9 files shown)


dalex78:
I am in the process of checking how other compilers deal with this (probably should have done that first). I don't see why that would be a specific GHDL issue. GCC and others must have a way to deal with this. Might as well align on what they do.

dalex78:
# How gcc and the build systems around it name object files

Background: while working on the object file naming for issues #539 / #1622 / #3088, the question came up of whether gcc faces the same problem and how it deals with it.  The answer splits in two, because gcc has the problem in one place and dodges it everywhere else.

## Plain gcc doesn't solve it — it clobbers

```
$ gcc -c src/a.c lib/a.c ; echo $?
0
$ nm a.o
0000000000000000 T la
```

Exit 0, no warning, `src/a.c`'s object silently gone.  `gcc -c` derives the name from the *basename* into the cwd, exactly like GHDL's original scheme.

GCC gets away with it because of a structural difference: it has `-o`, and no library.  **The build system owns the object name, and gcc never has to recover it** — the makefile already records which object came from which source.  GHDL can't do that: it owns a persistent library directory and must recompute the object name at link time from the `.cf`, with no makefile to consult.  That is the whole reason GHDL has this problem and gcc doesn't.

## So the real comparison is the build systems on top of gcc — and they hit the exact same problem

**Automake** does what `ghdllocal.adb` currently does — it detects the collision and refuses (`/usr/bin/automake:1961`, automake 1.16):

```perl
err_am "object '$object' created by '$full' and '$object_map{$object}'"
```

Its fix is the `subdir-objects` option, i.e. mirroring.  And the comment right above its naming code is almost word for word the trade-off GHDL hits: per-target flags force a `DERIVED_OBJECT` prefix which *"will result in a name that is too long for losing systems"*, so it strips directory components under `subdir-objects` and offers `_SHORTNAME` as a manual last resort.  Note what they did **not** reach for: a hash.

**CMake** (measured with cmake 4.4.2, Unix Makefiles) mirrors, and keeps the full source name:

```
main.c.o
src/a.c.o
lib/a.c.o
deep/a/b/c/a.c.o
tmp/claude-1000/.../scratchpad/outside.c.o     <- ../outside.c
```

Two things worth noting.  `a.c.o` — extension kept, exactly the `test.vhd.o` scheme adopted here.  And its answer to a source *outside* the tree is to splice the **absolute path** under the object dir — precisely the `..` case GHDL could not mirror, solved by making paths longer still.  CMake carried a `CMAKE_OBJECT_PATH_MAX` knob for years over that pain (4.4 no longer defines it).  Also: the object dir is per-target, so one source used by two targets is two objects.

**Meson** takes the third route — it flattens the relative path into the name with `_` (`libfoo.a.p/src_foo.c.o`).  (From memory; meson is not installed on this machine, so this one is unverified, unlike the automake and cmake results above.)

**And gcc itself, where it *does* face the recovery problem: gcov.**  `.gcda` / `.gcno` files have to be found again later with no makefile in hand — the same bind as GHDL's link step.  GCC's answer with `-fprofile-dir=` is to mangle the whole path into a single flat name, `/` becoming `#`: `#home#user#src#a.gcda`.  So when gcc has to recover a name from a flat directory, it does reversible flat mangling.

## What that says for GHDL

- **Nobody hashes.**  Every one of them preserves derivability, either by mirroring (cmake, automake) or by reversible flattening (gcov, meson).  A hash is the one thing the ecosystem consistently avoids, and jpf91's complaint on #1622 is the reason why.
- **Everyone keeps the source extension** in the object name.  The scheme on this branch already matches.
- The system whose constraints are closest to GHDL's — gcov, which must recover names from a flat directory — is the one that chose flat mangling, not mirroring.
- But GHDL has one constraint none of them has: `install.vhdllib` copies the library **flat**, and GHDL's own libraries are analyzed through `../../src/ieee/...`.  That is what makes cmake's answer untransplantable and what makes the NAME_MAX ceiling of a flat reversible encoding a real problem rather than a theoretical one.

This precedent nudges away from the plain hash: derivability is worth paying for, and the shape that fits GHDL's constraints is **mirroring when the workdir differs from the analysis directory, with the hash demoted to a fallback** — cmake/automake behaviour, which gives jpf91 his `build/src/test.vhd.o`, and hashing only where mirroring cannot apply.  Automake's `_SHORTNAME` escape hatch is the precedent saying that a fallback for the awkward cases is acceptable rather than a design failure.

One caveat that argues for leaning on `--gen-makefile` / `--gen-depends` regardless: **none of these systems actually relies on a pattern rule.**  CMake, meson and automake all emit an explicit rule per object; the derivable name is a convenience for humans reading the build tree, not the mechanism.  Which is exactly what `--gen-makefile` already provides, including the object-to-object dependencies that a `%.o: %.vhd` pattern rule could never express.

